PR #304, opened 2015-02-12
Optimization sine/cosine functions.

gamerforEA:
Using sin/cos lookup table.

Benchmark result:
FastMath: 832249-843543 scores
QuickMath: 882158-885743 scores

Middle SPS on rendering:
FastMath: ~807000 SPS
QuickMath: ~850000 SPS

### DIFF
--- a/chunky/src/java/se/llbit/chunky/main/Chunky.java
+++ b/chunky/src/java/se/llbit/chunky/main/Chunky.java
@@ -65,6 +65,7 @@ import se.llbit.chunky.world.RegionQueue;
 import se.llbit.chunky.world.World;
 import se.llbit.chunky.world.listeners.ChunkTopographyListener;
 import se.llbit.log.Log;
+import se.llbit.math.QuickMath;
 import se.llbit.math.Vector3d;
 import se.llbit.util.OSDetector;
 import se.llbit.util.OSDetector.OS;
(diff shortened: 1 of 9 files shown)


gamerforEA:
Hmmm, older comments saved... Strangely.

llbit:
I have still not benchmarked this to see if it works well in practice. The large lookup tables could blow up the CPU cache. Sometimes it's better to recompute instead of waiting for memory loads.

The precision of the sine/cosine functions is not super important, so it does not matter if they compute the most precise value, and it would be nice to replace the Apache math library. On the other hand it is generally not a good idea to maintain your own math functions - better to leave that to a well tested library.

llbit:
I tried to manually merge this pull request, but can't download the patch from GitHub:

```
$ curl -L https://github.com/llbit/chunky/pull/304.patch
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   138    0   138    0     0    215      0 --:--:-- --:--:-- --:--:--   215
100    32    0    32    0     0     24      0 --:--:--  0:00:01 --:--:--   205
Sorry, this diff is unavailable.
```

The FastMath class is used in many more places in Chunky now than when this pull request was created, and not only the sine and cosine functions are used - `FastMath.pow()` is also used for example. I don't really want to go through and edit all the sine/cosine uses to benchmark this pull request now. As I said in my previous comment I have some doubts that this would actually improve performance. However, that is moot since I don't seem to be able to merge this pull request at all, and I would have to rewrite the whole thing manually.

Please feel free to make a new pull request with the changes based on the current version of Chunky. I will make sure to benchmark it quickly if you create a new pull request, so that it doesn't take years to resolve the pull request as in this case.

PR #306, opened 2015-02-16
Fix typo in README.md

matthiasvegh:
Slight typo, thought I'd fix it.


llbit:
That line should actually be removed since log4j is no longer used.

Issue #364, opened 2016-05-13
Request - implement 360 VR projection


jason-green-io:
According to this: https://developers.google.com/cardboard/jump/rendering-ods-content.pdf

leMaik:
This idea is awesome! I started to implement this some hours ago and it's alreay working. Just need to add some GUI elements and then I'll send a PR. :+1:

jason-green-io:
Ohhhhh, I can't wait to play with this!

The way you implemented it you have to generate left and right and then
stitch them together outside Chunky, correct?

leMaik:
@jason-green-io Exactly. You have to render one image with the _ODS (left eye)_ projection mode and one with _ODS (right eye)_ and then join them.
Example image: https://lemaik.github.io/rs-sphere/

Self advertising: You can use [rs-sphere](https://github.com/leMaik/rs-sphere/] to display ODS images in your browser, with gyro support and all.

PR #368, opened 2016-06-05
Added a projector for omni-directional stereo

leMaik:
This PR adds a new projector for [Omni-directional Stereo][ods-google] pictures and two new projection modes (left and right eye). The resulting two images of a scene can be used to create stereographic 360° VR spheres. :globe_with_meridians:

Requested in #364. I also already signed the CLA. :+1: 

[ods-google]: https://developers.google.com/vr/jump/rendering-ods-content.pdf

Edit: I uploaded a [demo sphere](https://lemaik.github.io/rs-sphere/). Just click on the VR icon on the bottom right of the page. Works great with Google Cardboard.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
@@ -38,10 +38,17 @@ public class OmniDirectionalStereoProjector implements Projector {
   /**
    * The interpupillary distance of the viewer, in meters.
    */
-  private double interpupillaryDistance = 0.069;
+  private static final double interpupillaryDistance = 0.069;
+
+  private final double scale;
 
   public OmniDirectionalStereoProjector(Eye eye) {
     this.eye = eye;
+    if (eye == Eye.LEFT) {
+      scale = -interpupillaryDistance / 2;
+    } else {
+      scale = interpupillaryDistance / 2;
+    }
   }
 
   @Override
@@ -54,12 +61,6 @@ public class OmniDirectionalStereoProjector implements Projector {
     double theta = (x + 0.5) * FastMath.PI - FastMath.PI;
     double phi = FastMath.PI / 2 - (y + 0.5) * FastMath.PI;
 
-    double scale;
-    if (eye == Eye.LEFT) {
-      scale = -interpupillaryDistance / 2;
-    } else {
-      scale = interpupillaryDistance / 2;
-    }
     pos.set(FastMath.cos(theta) * scale, 0, FastMath.sin(theta) * scale);
     direction.set(FastMath.sin(theta) * FastMath.cos(phi), FastMath.sin(phi), -FastMath.cos(theta) * FastMath.cos(phi));
   }
--- a/chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
@@ -31,11 +31,6 @@ import java.util.Random;
  */
 public class OmniDirectionalStereoProjector implements Projector {
   /**
-  * The eye that this projector produces rays for.
-  */
-  private final Eye eye;
-
-  /**
    * The interpupillary distance of the viewer, in meters.
    */
   private static final double interpupillaryDistance = 0.069;
@@ -43,7 +38,6 @@ public class OmniDirectionalStereoProjector implements Projector {
   private final double scale;
 
   public OmniDirectionalStereoProjector(Eye eye) {
-    this.eye = eye;
     if (eye == Eye.LEFT) {
       scale = -interpupillaryDistance / 2;
     } else {
--- a/chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
@@ -1,4 +1,4 @@
-/* Copyright (c) 2016 Jesper Öqvist <jesper@llbit.se>
+/* Copyright (c) 2016 Chunky contributors
  *
  * This file is part of Chunky.
  *
--- a/chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
@@ -23,9 +23,10 @@ import java.util.Random;
 
 /**
  * A projector for Omni-Directional Stereo (ODS) images. x is mapped to yaw, y is mapped to
- * pitch. This projector is like the {@link PanoramicProjector} but can create distinct
- * images for the left and the right eye to create panoramic stereo images that are perfect
- * for viewing on VR devices.
+ * pitch. This projector, unlike the {@link PanoramicProjector}, can create distinct
+ * images for the left and the right eye by slightly displacing the view ray origins based on the
+ * viewing angle to account for the inter-pupillary distance. This allows to create panoramic
+ * stereo images that are perfect for viewing on VR devices.
  *
  * @see <a href="https://developers.google.com/vr/jump/rendering-ods-content.pdf">Rendering Omni‐directional Stereo Content</a>
  */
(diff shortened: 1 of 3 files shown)


leMaik:
Now that the JavaFX rewrite seems to be done: Should I rebase (and update) my PR or should this be a plugin. Personally, I like the idea of adding new projection modes via plugins, but that would require an API to do so.

llbit:
@leMaik I think this can be added to the core Chunky module. I'd like to make it possible to add custom projection modes through plugins too, but it would require more code changes.

leMaik:
@llbit Allright, I'll update the PR. Seems that a few people find this useful. :)

chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
```diff
@@ -0,0 +1,84 @@
...
+			Vector3d d) {
+		apply(x, y, o, d);
+	}
+
+	@Override
+	public void apply(double x, double y, Vector3d o, Vector3d d) {
+		double theta = (x + 0.5) * FastMath.PI - FastMath.PI;
+		double phi = FastMath.PI / 2 - (y + 0.5) * FastMath.PI;
+				
+		double scale;
```
llbit: The scale parameter can be set when an ODS projector is created, thus avoiding this if-statement in the projection code. The ODS projector is only created once, but projects very many rays.

chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+public class OmniDirectionalStereoProjector implements Projector {
+	/**
+	 * The eye that this projector produces rays.
+	 */
+	private final Eye eye;
+	/**
+	 * The interpupillary distance of the viewer, in meters.
+	 */
+	private double interpupillaryDistance = 0.069;
```
llbit: This should be a static final since it is never changed.

chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.util.Random;
+
+import org.apache.commons.math3.util.FastMath;
+
+import se.llbit.math.Vector3d;
+
+/**
+ * A projector for Omni-Directional Stereo (ODS) images. x is mapped to yaw, y is mapped to
+ * pitch.
```
llbit: There should be a comment explaining that this projector is different from the equirectangular panorama projector, and why it is different.

chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
```diff
@@ -0,0 +1,84 @@
+/* Copyright (c) 2014 Jesper Öqvist <jesper@llbit.se>
```
llbit: Change this to Copyright (c) 2016 Chunky contributors.

llbit: Please change this to Copyright (c) 2016 Chunky contributors.

chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
leMaik: It's intentionally neither static nor final since I planned to add a setting for the IPD later, as you suggested above.

chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
leMaik: Done.

chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
leMaik: That's why I added a link to `@see`, but I will extend the description.

chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
leMaik: You're absolutely right. Thanks!

leMaik:
@llbit Allright, I fixed everything you commented on the lines.

Regarding the upside-down images: Set the orientation to -180, -90, 180 (the fields should really have labels...) to rotate them correctly.

chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
```diff
@@ -23,7 +23,9 @@
 
 /**
  * A projector for Omni-Directional Stereo (ODS) images. x is mapped to yaw, y is mapped to
- * pitch.
+ * pitch. This projector is like the {@link PanoramicProjector} but can create distinct
+ * images for the left and the right eye to create panoramic stereo images that are perfect
+ * for viewing on VR devices.
```
llbit: The interesting difference between ODS and equirectangular projection is that they don't produce identical images because the ODS projector slightly displaces the view ray based on the viewing angle to account for the inter-pupillary distance. Equirectangular projection, on the other hand, works more like a pinhole projector - it uses a single point of origin for all rays. So the important difference is that they are not identical and the ray origin is altered in ODS projection.

leMaik: How about
```
/**
 * A projector for Omni-Directional Stereo (ODS) images. x is mapped to yaw, y is mapped to
 * pitch. This projector, unlike the {@link PanoramicProjector}, can create distinct
 * images for the left and the right eye by slightly displacing the view rays based on the
 * viewing angle to account for the inter-pupillary distance. This allows to create panoramic
 * stereo images that are perfect for viewing on VR devices.
 *
 * @see <a href="https://developers.google.com/vr/jump/rendering-ods-content.pdf">Rendering Omni‐directional Stereo Content</a>
 */
```

llbit: Looks good except I'd change "displacing the view rays" to "displacing the view ray origins"

leMaik: Allright, there you go!

llbit:
Thanks for making the changes. I added two small changes: I flipped the y-axis and changed capitalization in `interPupillaryDistance` so that "pupillary" is formatted like a separate word.

chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
llbit: Excellent!

chunky/src/java/se/llbit/chunky/renderer/projection/OmniDirectionalStereoProjector.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+package se.llbit.chunky.renderer.projection;
+
+import org.apache.commons.math3.util.FastMath;
+import se.llbit.math.Vector3;
+
+import java.util.Random;
+
+/**
+ * A projector for Omni-Directional Stereo (ODS) images. x is mapped to yaw, y is mapped to
```
llbit: The "x is mapped to yaw, y is mapped to pitch" part is fairly non-descriptive. The comment is from the equirectangular projector which maps linearly between image coordinates and yaw/pitch. In the ODS projector the mapping is a bit more complicated so the comment should just be removed.

I didn't think about this before I merged your commits so I will update the comment myself. Just wanted to add this comment here for completeness sake.

leMaik:
Sounds good. But by flipping _one_ axis, you flipped the image horizontally.

Equirectangular projection:
image

ODS projection (look at the tree and the sun's position over the mountain):
image

llbit:
@leMaik Oops! Thanks for notifying me. I have mirrored it horizontally now, should make it un-mirrored!

leMaik:
@llbit Thanks for the quick fix! Another small thing (that I don't know how to fix) is that the camera's yaw needs to be lowered by exactly 90° in order to point to the same location in ODS as in equirectangular projection.

The following orientations make the camera point to the same location:
equirectangular projection; orientation 90, -90 ,0
ods projection; orientation 0, -90, 0

llbit:
@leMaik It should be fixed now: 34a26c6569c193181105d72d62dc996b97ce9c90

PR #418, opened 2017-02-19
Add Hable (Uncharted 2) and ACES filmic tone mapping

leMaik:
(I hit the enter key before writing a description, sorry!)

I know that there are plugins, but these two new tone mappings seem to be a useful addition to the core, as there already is the _tonemapping operator 1_.

Sometimes, the scene looks pretty grayish with gamma correction and even with the existing tone mapping operator

Gamma correction:
markus-100-gamma

Tone mapping op. 1:
markus-100-tonemap1

Hable's tonemapping looks really great, it preserves some colors better than the above modes:
markus-100-hable

And while we're at it, the ACES tone mapping curve aims to give the scenes a more filmic look, although it requires careful adjusting of the sunlight and exposure (my scene looks a little too blueish, but the contrast and color intensity is much better):
markus-100-aces
By the way, ACES [is the default tonemap in Unreal Engine 4](https://docs.unrealengine.com/latest/INT/Support/Builds/ReleaseNotes/2015/4_8/).

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/Postprocess.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/Postprocess.java
@@ -38,10 +38,14 @@ public enum Postprocess {
     }
   },
   TONEMAP2 {
-    @Override public String toString() { return "ACES filmic tone mapping"; }
+    @Override public String toString() {
+      return "ACES filmic tone mapping";
+    }
   },
   TONEMAP3 {
-    @Override public String toString() { return "Hable tonemapping"; }
+    @Override public String toString() {
+      return "Hable tonemapping";
+    }
   };
 
   public static final Postprocess DEFAULT = GAMMA;
--- a/chunky/src/java/se/llbit/chunky/renderer/Postprocess.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/Postprocess.java
@@ -36,6 +36,12 @@ public enum Postprocess {
     @Override public String toString() {
       return "Tonemap operator 1";
     }
+  },
+  TONEMAP2 {
+    @Override public String toString() { return "ACES filmic tone mapping"; }
+  },
+  TONEMAP3 {
+    @Override public String toString() { return "Hable tonemapping"; }
   };
 
   public static final Postprocess DEFAULT = GAMMA;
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/chunky/renderer/Postprocess.java
```diff
@@ -36,6 +36,12 @@
     @Override public String toString() {
       return "Tonemap operator 1";
     }
+  },
+  TONEMAP2 {
+    @Override public String toString() { return "ACES filmic tone mapping"; }
```
llbit: I prefer to not use this single-line style of method declaration. It's a personal preference, but it's better to keep everything consistent. See for example the other `toString()` methods in this file.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1927,6 +1927,37 @@ public void postProcessPixel(int x, int y, double[] result) {
...
+          break;
+        case TONEMAP3:
+          // http://filmicgames.com/archives/75
+          float hA = 0.15f;
+          float hB = 0.50f;
+          float hC = 0.10f;
+          float hD = 0.20f;
+          float hE = 0.02f;
+          float hF = 0.30f;
+          r *= 16; // manually adjust exposure to approx. match other post-processing modes
```
llbit: I looked at the linked blog, and they seem to use the same exposure adjustment constant. Did you try adjusting it?

Generally I prefer proper sentences in comments nowadays, i.e. start with capital letter and end with a period.

chunky/src/java/se/llbit/chunky/renderer/Postprocess.java
leMaik: Whoops... My IDE collapsed the other blocks so they looked like one-liners.
I (and my IDE) would prefer to have the `@Override` in a separate line, and I've never really seen it on the same line as the method declaration, what do you think?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
leMaik: Indeed, they use the same exposure adjustment. I should call it "hardcoded" instead of "manual" exposure adjustment. It's a little weird because they also use it for the _tonemap1_ operator, but you don't and it still looks fine, while Hable's tonemap looks far too dark without adjusting the exposure.

chunky/src/java/se/llbit/chunky/renderer/Postprocess.java
llbit: I have no preference about annotation placements - I usually have them on the same line if it fits.

Issue #416, opened 2017-02-12
Let the plugins specify their main class


leMaik:
At the moment, installing plugins for non-dev users is a little more complicated than it probably should be. I like the fact the an md5 checksum is required, because this enforces at least a little security.

Specifying the main class, however, isn't straight-forward for most users (that might not even know what a class is). I'd suggest to add a manifest file inside the plugins (i.e. something like the `plugin.yml` inside Bukkit plugins). Such a file could specify the main class and more meta data that the PlugMan could display (e.g. author, version, website).

If you want, I'd love to start working on this by adding support for a plugin definition file inside the plugin.jar (I'd use JSON, as pretty much everything in Chunky is configured using JSON already) and updating PlugMan. :+1:

llbit:
Yes, having some sort of manifest for plugins to specify their main class would be excellent! 
It would be great if you want to add this feature. I'm not working on it at the moment. Using JSON for the plugin manifest files is a good idea. I suggest using a format similar to the format used in the `.chunky/plugin` directory. If plugins have manifests, then the plugin Jars could be dropped into that folder directly.

I had planned to eventually integrate the plugin manager into the launcher. You could work on that too, if you want to.

The MD5 checksum was not indented for security purposes, only to verify that the Jar file is not corrupt. Right now the MD5 checksum is generated by the plugin manager, but I think that should also be part of the plugin manifest and provided by plugin authors.

leMaik:
@llbit I'll start working on this asap, thank you for the feedback!

Regarding the MD5 sum: My inception knowledge tells me that it might be hard to put the md5 sum of the jar into a file into the jar whose checksum should be put into a file into the jar whose checksum should be put… :smile: Should we just drop it then?

llbit:
> Regarding the MD5 sum: My inception knowledge tells me that it might be hard to put the md5 sum of the jar into a file into the jar whose checksum should be put into a file into the jar whose checksum should be put… :smile: Should we just drop it then?

Haha, you are right! I didn't think that through completely. Jar files have an check sum built-in via the Zip format anyway, so it is probably redundant to use extra MD5 sums. Let's forget about checksums for the moment, it's always possible to add that later.

leMaik:
Allright, I'll skip the checksums then. :+1: 

[...]

PR #436, opened 2017-07-06
Add a manifest for plugins

leMaik:
This PR adds manifest parsing for plugins, resolving #416.

The manifest looks like this:
```
{
  "name": "Demo Plugin",
  "version": "1.0.0",
  "author": "leMaik",
  "main": "de.leMaik.demoplugin.Plugin",
  "targetVersion": "1.4.3"
}
```

The name, version, author and targetVersion are optional but can be used for the plugin list later. If the actual Chunky version doesn't match the `targetVersion`, a warning is printed on the console (for now, this could be improved later).


chunky/src/java/se/llbit/chunky/main/Chunky.java
```diff
@@ -51,20 +34,21 @@
...
  * Chunky is a Minecraft mapping and rendering tool created by
  * Jesper Öqvist (jesper@llbit.se).
- *
+ * <p>
  * <p>Read more about Chunky at http://chunky.llbit.se .
  */
 public class Chunky {
 
-  /** A log receiver suitable for headless rendering. */
+  /**
```
llbit: Please try to avoid formatting changes (like this and the extra "<p>") above in code not directly related to your changes in the future. It's just a minor issue, but it makes it easier to review the important changes. Don't worry about changing it now though, that would just create unnecessary extra work.

leMaik: Sorry, my IDE automatically did this during auto-formatting.

lib/src/se/llbit/chunky/resources/SettingsDirectory.java
```diff
@@ -106,6 +106,13 @@ public static File getWorkingDirectory() {
   }
 
   /**
+   * @return the plugin directory
+   */
+  public static File getPluginsDirectory() {
+    return new File(getWorkingDirectory(), "plugins");
```
llbit: Is there a particular reason you used `getWorkingDirectory()` rather than `getSettingsDirectory()` here? It seems like the settings directory is what you intended to use here.

chunky/src/java/se/llbit/chunky/plugin/ChunkyPlugin.java
```diff
@@ -0,0 +1,112 @@
...
+   *
+   * @return the implementation of this plugin
+   */
+  public Plugin getImplementation() {
+    return implementation;
+  }
+
+  public static ChunkyPlugin load(File pluginJar) throws LoadPluginException {
+    Map<String, String> env = new HashMap<>();
+    env.put("create", "true");
```
llbit: What is the purpose of this? According to [this page](http://docs.oracle.com/javase/7/docs/technotes/guides/io/fsp/zipfilesystemprovider.html) it seems like that would try to create a new zip file if the `pluginJar` file does not actually point to a valid file. That case is guarded against by the caller, but it still seems unnecessary to set up this environment variable.

Would it not be better to use `getFileSystem(String)` instead? Or do you need to use `newFileSystem`?

lib/src/se/llbit/chunky/resources/SettingsDirectory.java
leMaik: You're right, using the settings directory would be better. I'll fix that.

chunky/src/java/se/llbit/chunky/plugin/ChunkyPlugin.java
leMaik: Indeed, the file will always exist here, so [no env is needed](http://docs.oracle.com/javase/7/docs/technotes/guides/io/fsp/zipfilesystemproviderprops.html).

leMaik:
@llbit I only tested this on Linux. I'll apply your fix and check if it still works on Linux.

lib/src/se/llbit/chunky/resources/SettingsDirectory.java
llbit: I fixed it in a follow-up commit, no need to update the pull request.

leMaik:
@llbit Allright, thank you for merging this! Still works on Linux by the way. :tada:

llbit:
@leMaik I tested it on Mac and it works there as well. Thank you for the pull request!

PR #457, opened 2017-09-07
Fixed red sandstone and purpur stairs rendering

electron93:
Before:
stairs-before

After:
stairs-after


llbit:
Thank you! I tested it and the fix works perfectly!

PR #515, opened 2019-02-18
Fix for TableView in the "Load Scene" dialog not stretching when resizing window

booleanbyte:
Fix for issue 510, making the TableView in the "Load Scene" dialog grow vertically when the window is resized.


llbit:
Thanks for the fix!

PR #541, opened 2019-08-22
Adds block coordinates to map tooltip. Fixes #489

### LINKED ISSUE #489: [Feature] Coordinates in the main window
Another very, very minor feature-request.

It would be nice if the main window would display the "block-coordinates" in addition to the chunk-coordinates. Right now the pop-up displays something like:

> Chunk: [10, -3] plains

it would be nice if you could add the block-coordinates, too. Something like:

> Chunk: [10, -3] plains
> Coords: 163, -51

animalcharade:


leMaik:
Better late than never, I guess :sweat_smile: Thank you, @animalcharade

PR #570, opened 2020-04-24
Fix nether portal not emitting light.

leMaik:


chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
```diff
@@ -158,7 +158,7 @@ public Block get(int id) {
     materialProperties.put("minecraft:glowstone", block -> {
       block.emittance = 1.0f;
     });
-    materialProperties.put("minecraft:portal", block -> {
+    materialProperties.put("minecraft:nether_portal", block -> {
```
leMaik, line 161: Just saw that this was changed in 1.13, so the old should stay there.

PR #583, opened 2020-05-10
Add a button to copy the current frame to the clipboard.

leMaik:


chunky/src/java/se/llbit/png/PngFileWriter.java
```diff
@@ -41,9 +40,9 @@
   /**
    * @throws IOException
    */
-  public PngFileWriter(File file) throws IOException {
```
leMaik, line 44: This broke the denoiser plugin (and any other plugin that uses the PngFileWriter to save images).

chunky/src/java/se/llbit/tiff/TiffFileWriter.java
```diff
@@ -40,8 +39,8 @@
 
   private final DataOutputStream out;
 
-  public TiffFileWriter(File file) throws IOException {
```
leMaik, line 43: This broke any plugin that uses the TiffFileWriter to save images.

PR #598, opened 2020-06-01
Make furnace, blast furnace and smoker emitters

### LINKED ISSUE #345: Feature Request: Lit furnaces act as light source
As the title says, is it possible for lit furnaces to act as a light source.

073TheBoss:
Resolves #345. Also fixes typo in the Furnace class.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/block/Furnace.java
+++ b/chunky/src/java/se/llbit/chunky/block/Furnace.java
@@ -30,7 +30,7 @@ public class Furnace extends MinecraftBlock {
               Texture.furnaceTop, Texture.furnaceTop,
           },
       },
-      // Unlit.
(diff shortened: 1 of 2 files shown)


leMaik:
The material properties for all other blocks are configured in [the BlockPalette](https://github.com/llbit/chunky/blob/3f4e4527c449c0ea08413e7744d208cfcb24aff2/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java#L82) class, which decouples block loading from Material properties and can be used for the materials tab later.

Would you mind moving this logic to the `BlockPalette`?

073TheBoss:
Ok, will do. I looked for like an hour to try to find where it was set up and couldn’t find it lol.

PR #604, opened 2020-06-05
New octree implementation

aTom3333:
This PR adds a new implementation of the octree but also leaves the old one. The implementation is abstracted a bit, allowing the use of a different octree implementation with the same octree.

This new implementation doesn't use Nodes as its internal representation, it uses a big int array to be more memory efficient. Nonetheless it offers the same interface that the old implementation, limiting the need to change code using the octree. (the depth is no longer available but came be replaced by a call to getDepth()).

The idea behind this implementation is to reduce greatly the memory usage of the octree and, hopefully, improving the performance by being more cache friendly (the algorithms are the exact same). I suggest reading the code if you are curious how it works, there are comments explaining.
From my limited testing, it seems that both of this goal have been reached. The memory usage of this implementation is almost a fifth of the old one's, this almost halves the memory usage at the process level for a scene with 4k chunks. As for the performance, i've seen an improvement of 30% in render time for a scene with 1k chunks and of 40% for a scene with 4k chunks.

One caveat with this implementation is that, as everything is stored in a single array, sometime the octree could be too big to fit inside. When this is the case the octree falls back to the old, node-based, implementation

### DIFF
--- a/chunky/src/java/se/llbit/chunky/resources/OctreeFileFormat.java
+++ b/chunky/src/java/se/llbit/chunky/resources/OctreeFileFormat.java
@@ -32,8 +32,9 @@ public class OctreeFileFormat {
    * Load octrees and grass/foliage textures from a file.
    *
    * @param in input stream for the file to load the scene from.
+   * @param forceNodeBased Forces the octrees to be loaded as NodeBasedOctree
    */
-  public static OctreeData load(DataInputStream in) throws IOException {
+  public static OctreeData load(DataInputStream in, boolean forceNodeBased) throws IOException {
     int version = in.readInt();
     if (version != OCTREE_VERSION) {
       throw new IOException(String.format(
@@ -42,11 +43,20 @@ public class OctreeFileFormat {
     }
     OctreeData data = new OctreeData();
     data.palette = BlockPalette.read(in);
-    data.worldTree = Octree.load(in);
-    data.waterTree = Octree.load(in);
+    data.worldTree = Octree.load(in, forceNodeBased);
+    data.waterTree = Octree.load(in, forceNodeBased);
     data.grassColors = WorldTexture.load(in);
     data.foliageColors = WorldTexture.load(in);
     return data;
+  }
+
+  /**
+   * Load octrees and grass/foliage textures from a file.
+   *
+   * @param in input stream for the file to load the scene from.
+   */
+  public static OctreeData load(DataInputStream in) throws IOException {
+    return load(in, false);
   }
 
   /**
--- a/chunky/src/java/se/llbit/math/Octree.java
+++ b/chunky/src/java/se/llbit/math/Octree.java
@@ -40,7 +40,19 @@ import se.llbit.log.Log;
  */
 public class Octree {
 
-  static final int BRANCH_NODE = -1;
+  public interface OctreeImplementation {
+    void set(int type, int x, int y, int z);
+    void set(Node data, int x, int y, int z);
+    Node get(int x, int y, int z);
+    Material getMaterial(int x, int y, int z, BlockPalette palette);
+    void store(DataOutputStream output) throws IOException;
+    boolean isInside(Vector3 pos);
+    boolean enterBlock(Scene scene, Ray ray, BlockPalette palette);
+    boolean exitWater(Scene scene, Ray ray, BlockPalette palette);
+    int getDepth();
+  }
+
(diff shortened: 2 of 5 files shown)


chunky/src/java/se/llbit/chunky/resources/OctreeFileFormat.java
```diff
@@ -32,8 +32,9 @@
    * Load octrees and grass/foliage textures from a file.
    *
    * @param in input stream for the file to load the scene from.
+   * @param forceNodeBased Forces the octrees to be loaded as NodeBasedOctree
    */
-  public static OctreeData load(DataInputStream in) throws IOException {
+  public static OctreeData load(DataInputStream in, boolean forceNodeBased) throws IOException {
     int version = in.readInt();
     if (version != OCTREE_VERSION) {
```
leMaik, line 39: Did the file format stay the same? If not, the `OCTREE_VERSION` needs to be increased.

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -195,27 +207,14 @@ public DataNode(int type, int data) {
...
 
-  private final Node[] parents;
-  private final Node[] cache;
-  private int cx = 0;
-  private int cy = 0;
-  private int cz = 0;
-  private int cacheLevel;
+  private OctreeImplementation implementation;
+
+  static private boolean usePacked = true;
```
leMaik, line 217: Maybe we should make this configurable via a Java system variable (`usePacked = !System.getProperty("chunky.useLegacyOctree").equals("true")`) so that users can change this if they encounter issues by launching Chunky with `-Dchunky.useLegacyOctree=true`

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -40,13 +40,25 @@
  */
 public class Octree {
```
leMaik, line 41: Maybe change `Octree` to be the interface that `OctreeImplementation` is at the moment, let `NodeBasedOctree` and `PackedOctree` implement `Octree` and create a `OctreeFactory` class that creates instances of the octree implementation the user wants?

Not sure about this, but seems to be a more Java-like way. Also, there wouldn't be that `implementation` field and manual "redirects" to the actual implementation in the method calls needed. :thinking:

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -242,7 +243,14 @@ public Octree(int octreeDepth, Node node) {
    * @param type The new voxel type to be set
    */
   public synchronized void set(int type, int x, int y, int z) {
-    set(new Node(type), x, y, z);
+    try {
+      implementation.set(type, x, y, z);
+    } catch(PackedOctree.OctreeTooBigException e) {
+      // Octree is too big, switch implementation and retry
+      Log.warn("Octree is too big, falling back to old (slower and bigger) implementation.");
+      switchToNodeBased();
```
leMaik, line 251: Oh... this wouldn't be possible with the changes I suggested above (regarding the interfaces).

chunky/src/java/se/llbit/chunky/resources/OctreeFileFormat.java
aTom3333: The format stayed the same, no need to update the OCTREE_VERSION

chunky/src/java/se/llbit/math/Octree.java
aTom3333: I'm not familiar with that but that seems like a good idea

chunky/src/java/se/llbit/math/Octree.java
aTom3333: Well, as you said next. Keeping it like that, albeit not really idiomatic, let us swap the implementation without having to update any reference to the tree

chunky/src/java/se/llbit/math/Octree.java
aTom3333: Implemented and works fine

aTom3333:
I need to a comment about the switching back to the old implementation in case the octree is too big. I haven't be able to test it as my computer just has not enough RAM to make an octree that big.
It should work but i'd like to be notified if someone has an issue when dealing with very big octrees

jackjt8:
I'll be happy to test this once it ends up in a nightly build. I have got 32GB of RAM which should provide us with a good sized octree.

PR #609, opened 2020-06-09
Reimplement the Materials tab

073TheBoss:
Please let me know if anything needs changed before this can be merged.


chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
```diff
@@ -68,6 +76,16 @@ public int put(BlockSpec spec) {
...
   }
 
+  public void updateProperties(String name, Consumer<Block> properties) {
+    //TODO definitely needs some optimizing
+    materialProperties.put(name, properties);
+    blockMap.forEach((spec, id) -> {
+      Block block = spec.toBlock();
+      applyMaterial(block);
+      palette.set(id, block);
+    });
```
aTom3333, line 85: Hmmm, this doesn't seem very efficient but i don't think there is a way to get a BlockSpec from the name (to get the id via the blockMap), let's see what leMaik thinks

leMaik: It's only done on scene load and it's ~800 blocks, shouldn't be too significant.

chunky/src/java/se/llbit/chunky/world/MaterialStore.java
```diff
@@ -17,16 +17,18 @@
...
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
 // TODO: introduce block tags
 public class MaterialStore {
   public static final Map<String, Collection<Block>> collections = new LinkedHashMap<>();
-  public static final Map<String, Material> idMap = new HashMap<>();
+  public static final Collection<String> idMap = new ArrayList<String>();
```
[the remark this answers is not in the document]
073TheBoss: I figured `MaterialStore` should be kept because it can be used in the future for material groups such as the `blocks:all` that was present in 1.4.5.

PR #613, opened 2020-06-13
Fix water bug in packed octree

aTom3333:
This PR fixes the bug causing water not to look as expected when using the packed octree

expected result:
water-node-octree

result with buggy packed octree:
water-packed-octree

This being said, i believe this PR needs a bit a discussion before being merged.
As i've explain earlier on discord, this issue does not occur with the node based octree because, when testing if children must be merged, they all are compared against the old node that has just been replaced that prevents the merging from happening because the children, even if they all compare equals don't compare equals to the old node.
I believe this behavior is a bug but i think its only bad consequence is that sometimes nodes are not merged when they could have been, wasting a bit of memory.

The bug that i think is really the cause of the water problem is the fact that when merging nodes, the data is not propagated to the parent. This is the bug fixed by this PR (only for the packed octree).
This bug is present in the node based octree but its effect wasn't noticeable because of the other bug that prevents merging.

To recap, from my understanding, there are 3 bugs, 2 being the same bug in two octree implementation. The bug in packed octree was noticeable, but the 2 bugs in nodes based octree weren't cause they are "cancelling" each other.

I think we should fix the 2 bugs in the node based octree as well but i'd like to wait for feedback to see if my interpretation is not wrong.

### DIFF
--- a/chunky/src/java/se/llbit/math/NodeBasedOctree.java
+++ b/chunky/src/java/se/llbit/math/NodeBasedOctree.java
@@ -29,9 +29,10 @@ public class NodeBasedOctree implements Octree.OctreeImplementation {
   /**
    * Root node
    */
-  public final Octree.Node root;
+  public Octree.Node root;
 
   private final Octree.Node[] parents;
+  private final int[] positions;
   private final Octree.Node[] cache;
   private int cx = 0;
   private int cy = 0;
@@ -42,6 +43,7 @@ public class NodeBasedOctree implements Octree.OctreeImplementation {
     depth = octreeDepth;
     root = node;
     parents = new Octree.Node[depth];
+    positions = new int[depth];
     cache = new Octree.Node[depth + 1];
     cache[depth] = root;
     cacheLevel = depth;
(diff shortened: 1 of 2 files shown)


leMaik:
> The bug that i think is really the cause of the water problem is the fact that when merging nodes, the data is not propagated to the parent.

@aTom3333 I agree that this is what should be fixed, for both octree implementations. Also, the `data` needs to be checked before merging the nodes so that only nodes with identical type and data get merged.

aTom3333:
The data is checked for both implementation prior to the merge

leMaik:
Sounds good to me then :100:

aTom3333:
So i just fixed the 2 bugs i talked about in the node based octree, unfortunately, in order to propagate data to the parent node, the parent node needs to be replaced by a DataNode, so we need to keep track of the grandparent node and the index of the parent in the grand parent. In theory it could even bubble up to the root that might have to be replaced and is no longer final

leMaik:
Nice! :tada:

Issue #612, opened 2020-06-12
[discussion] Refactoring suggestion


Stenodyon:
During the past few days I have been going through the code as I'd like to contribute (more than just one-line-change PRs :smile: ), especially to the renderer. I have noticed a few things and I'd like to open a discussion about them.

My main concern is that the `Ray` class does too much, it contains input (origin and direction), output (distance, materials, …) and temporary (t, tNext, …) values. This leads to functions that operate through side-effects modifying the `Ray` passed in input even though these could be pure (for instance, the [`TexturedBlockModel.getIntersectionColor(Ray ray)`](https://github.com/llbit/chunky/blob/573424f349117937bb5112ddb1418395286b00a5/chunky/src/java/se/llbit/chunky/model/TexturedBlockModel.java#L132) function doesn't actually return a color but has the side-effect of setting the ray's color, and uv fields to the result). This makes the code harder to read as what part of the `Ray` is modified by a function call must be kept in mind, an operation might overwrite a previous result, etc. making the computation harder to follow and bugs harder to solve.

My suggestion would be to extract all the output members to other classes and make `Ray` only contain its origin, direction and the medium it traverses. Intersecting the scene would give a `Hit` or `HitRecord` result that contains information specific to a hit that can be used for path construction and shading (distance, normal, materials, …). The temporary values such as `t` and `tNext` should be local to the functions that use them.

I would like to work on this but it is a big change so I wanted to get some opinions, and some feedback in case I made an oversight as the codebase is still fairly new to me.

aTom3333:
Hey, i'm pretty much in the same situation, new to the codebase but want to make big changes.

I share the sentiment that some parts would gain from refactoring them. When reading the code, it stroke me too the amount of things that were happening through the `Ray` class. This being said, i don't really have a lot of experience with path tracing (i've been interested in path tracing for quite a while, contributing to chunky is my way to get experience with it) so i don't know what the best way to do things is but from what i've seen this `HitRecord` approach is quite common.

So yeah, i would say i agree with you but i'm not someone experienced with the matter.

[...]

PR #615, opened 2020-06-14
Refactor octree intersection

Stenodyon:
In the octree intersection functions, the initialization (moving the ray to the boundary of the octree if it is outside) is done in the first iterations of the traversal loop.

* This refactor changes the initialization to directly intersect the octree's bounding box instead of marching to it in several iterations. It also moves it outside of the loop and into its own function `enterOctree(Ray ray)`. On top of making the code more readable it removes some iterations and some branching in a hot loop (though I saw no noticeable performance improvements).
* Variable declarations were also moved to more appropriate locations instead of having them all at the beginning of the function.

The initialization part could be reduced to a simple `AABB.intersect` ideally, but the `AABB` class has been implemented such that the bounding box must be contained within a block and wraps the ray origin to said block, certainly for use with models (despite the doc comments saying it could be used for Bounding Volume Hierarchies).

Related to #612

### DIFF
--- a/chunky/src/java/se/llbit/math/NodeBasedOctree.java
+++ b/chunky/src/java/se/llbit/math/NodeBasedOctree.java
@@ -165,7 +165,7 @@ public class NodeBasedOctree implements Octree.OctreeImplementation {
   }
 
   /**
-   * Moves the ray to the of the octree.
+   * Moves the ray to the boundary of the octree.
    * @param ray Ray that will be moved to the boundary of the octree. The origin, distance and normals will be modified.
    * @return {@code false} if the ray doesn't intersect the octree.
    */
--- a/chunky/src/java/se/llbit/math/NodeBasedOctree.java
+++ b/chunky/src/java/se/llbit/math/NodeBasedOctree.java
@@ -239,11 +239,17 @@ public class NodeBasedOctree implements Octree.OctreeImplementation {
     return true;
   }
 
+  /**
+   *  {@inheritDoc}
+   */
   @Override
   public boolean enterBlock(Scene scene, Ray ray, BlockPalette palette) {
     if (!isInside(ray.o) && !enterOctree(ray))
       return false;
 
+    // Marching is done in a top-down fashion: at each step, the octree is descended from the root to find the leaf
+    // node the ray is in. Terminating the march is then decided based on the block type in that leaf node. Finally the
+    // ray is advanced to the boundary of the current leaf node and the next, ready for the next iteration.
     while (true) {
       // Add small offset past the intersection to avoid
       // recursion to the same octree node!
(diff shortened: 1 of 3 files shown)


chunky/src/java/se/llbit/math/NodeBasedOctree.java
```diff
@@ -164,93 +164,98 @@ public boolean isInside(Vector3 o) {
     return lx == 0 && ly == 0 && lz == 0;
   }
 
+  // Moves the ray to the boundary of the octree. Returns false when the ray doesn't intersect the octree.
```
leMaik, line 167: This could be proper javadoc, maybe you could add that for all the octree classes while you're at it?

chunky/src/java/se/llbit/math/NodeBasedOctree.java
```diff
@@ -284,138 +289,76 @@ public boolean enterBlock(Scene scene, Ray ray, BlockPalette palette) {
...
-            return false;// outside of octree!
-          }
-        } else {
-          return false;// outside of octree!
-        }
-      }
-
-      first = false;
+      if (lx != 0 || ly != 0 || lz != 0)
+        return false;
```
leMaik, line 358: A comment that notes that the ray is outside of the octree here would be good. Explaining how the octree intersection works by a few comments would improve this code.

leMaik:
The PR is great and I'd merge it as-is. The readability improvements are good, but the maintainability could be further improved by documenting how the octree works in the algorithms and by adding some javadoc to the methods.

chunky/src/java/se/llbit/math/NodeBasedOctree.java
Stenodyon: I added a few comments in the intersection functions to try and give an idea of how they work, but I don't fully understand them (especially the part that decides when to terminate the loop).

chunky/src/java/se/llbit/math/NodeBasedOctree.java
```diff
@@ -164,93 +164,109 @@ public boolean isInside(Vector3 o) {
     return lx == 0 && ly == 0 && lz == 0;
   }
 
+  /**
+   * Moves the ray to the of the octree.
```
leMaik, line 168: *to the _boundary_ is the octree

leMaik:
@aTom3333 Would you mind having a look over this? You sure have a better understanding of the octree than I do. :)

chunky/src/java/se/llbit/math/NodeBasedOctree.java
Stenodyon: whoops thanks

aTom3333:
Seems good to me :)

Issue #150, opened 2013-06-17
Selective sampling


abonander:
Let's say I have a large render that's mostly clean, like the following one:

![Noise concentrated in small areas of the image](http://i.imgur.com/25HxzE5.jpg)

Where most of the noise is concentrated in a small area of the image because of the high density of emitters compared to the rest of the scene. It will probably take another 10-40k SPP to clean that noise up. At 1920x1080, that could take a while, considering the given render is currently at 13k SPP after about five hours.

Would it be possible to select only that area of the image to continue rendering? 

If I understand path tracing correctly, the sampling is based on pixels, so it should be as simple as bounding further sampling to the selected area. 

With a reduced sampling area, it should be as clean as the rest of the image in a lot less time than it would have taken sampling the image as a whole, with the render time shortened roughly proportionally by how much smaller the selected area is compared to the canvas.

sn0wyfall:
Yeah I was also thinking of being able to select it myself. I understand that rendering algorithms are tricky. I was also considering literally changing the camera view and selecting rectangular areas with greater noise to render better, but that seems quite complicated haha

MadDog443:
Honestly it might be easier if we had a ![Sky light checkbox](https://user-images.githubusercontent.com/52639083/130314472-a3b381fb-b7ed-410c-a905-b9a1051b2608.png) it would solve a lot of issues cause then we could sample only the emitters which would help a lot, cause we could start out by sampling everything but the emitters then switch to emitters only and set another SPP target.

Easiest way would be to do it the same way the denoiser works, have your skylight renderer, sunlight renderer, and emitter renderer; then you can set your target for each layer to save time rendering.

But if you wanted to go the extra mile, being able to select individual emitters and turn them off would be better, because then we could actually fix problem spots a little easier.

Peregrine05:
I believe that lens shift would be a somewhat complicated way to achieve this result, by narrowing the canvas to be one specific area of the larger canvas. Then only the pixels in the new smaller canvas are being sampled.

PR #618, opened 2020-06-21
Add adaptive sampling

Stenodyon:
I'm opening this draft PR because adding adaptive sampling is a lot of (re)work and I want to be as transparent as possible about what I change.

The main issue with adding this feature is that most of the state of the renderer is actually in the `Scene` class (which in my opinion should only describe the scene and not contain any state) and it is tightly coupled to a lot of things so I have to extract and decouple all that state and logic to the `RenderManager` and other classes.

Once that's done, implementing adaptive sampling should be pretty straightforward.

Related to #150

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/FrameSampler.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/FrameSampler.java
@@ -19,7 +19,7 @@ public abstract class FrameSampler {
    * Queries the FrameSampler for the next job in the queue. Blocks until one is available.
    * @return the next job for a RenderWorker
    */
-  public abstract RenderTile getNextJob() throws InterruptedException;
+  public abstract RenderTask getNextJob() throws InterruptedException;
 
   /**
    * Add a sample for pixel (x, y).
--- a/chunky/src/java/se/llbit/chunky/renderer/FrameSampler.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/FrameSampler.java
@@ -19,7 +19,7 @@ public abstract class FrameSampler {
    * Queries the FrameSampler for the next job in the queue. Blocks until one is available.
    * @return the next job for a RenderWorker
    */
-  public abstract RenderTile getNextJob();
+  public abstract RenderTile getNextJob() throws InterruptedException;
 
   /**
    * Add a sample for pixel (x, y).
--- a/chunky/src/java/se/llbit/chunky/renderer/FrameSampler.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/FrameSampler.java
@@ -16,8 +16,52 @@ import java.util.concurrent.BlockingQueue;
  * A FrameSampler decides how samples are taken and reconstructs the image from them.
  */
 public abstract class FrameSampler {
-  public static final byte SAMPLER_TYPE_UNIFORM = 1;
-  public static final byte SAMPLER_TYPE_ADAPTIVE = 2;
+  public enum Type {
+    /**
+     * {@see UniformFrameSampler}
+     */
+    UNIFORM(1),
+
+    /**
+     * {@see AdaptiveFrameSampler}
+     */
+    ADAPTIVE(2);
+
+    /**
+     * This tag is used to determine the type of the sampler when (de)serializing from and to
+     * data streams.
+     */
+    public final byte tag;
+
+    private Type(int tag) {
+      this.tag = (byte)tag;
+    }
+
+    /**
+     * Returns the enum variant associated with the given tag.
+     * @param tag The tag representing the type
+     * @return The enum variant associated with the given tag
+     * @throws IllegalArgumentException if an invalid tag was given
+     */
+    public static Type from(byte tag) throws IllegalArgumentException {
+      switch (tag) {
+        case 1:
+          return UNIFORM;
+        case 2:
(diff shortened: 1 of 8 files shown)


chunky/src/java/se/llbit/chunky/renderer/FrameSampler.java
```diff
@@ -0,0 +1,173 @@
...
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+import java.util.concurrent.BlockingQueue;
+
+/**
+ * A FrameSampler decides how samples are taken and reconstructs the image from them.
+ */
+public abstract class FrameSampler {
+  public static final byte SAMPLER_TYPE_UNIFORM = 1;
```
leMaik, line 19: This should be an enum instead.

chunky/src/java/se/llbit/chunky/renderer/RenderManager.java
```diff
@@ -21,8 +21,10 @@
 import se.llbit.log.Log;
 import se.llbit.util.TaskTracker;
 
+import java.awt.*;
```
leMaik, line 24: Is AWT actually used in this class?

chunky/src/java/se/llbit/chunky/renderer/Renderer.java
```diff
@@ -75,6 +76,28 @@
...
+   * @return {@code true} if loading was successful
+   */
+  boolean loadDump(File inputFile, TaskTracker taskTracker);
+
+  /**
+   * Saves a snapshot of the current frame to the given directory.
+   * @param directory the directory in which the snapshot will be saved
+   * @param taskTracker progress tracking for this save operation
+   */
+  void saveSnapshot(File directory, TaskTracker taskTracker);
```
leMaik, line 99: While you're at it, maybe this should be changed to `File outputFile` instead of `directory`.

leMaik:
@Stenodyon This looks great so far! :+1: The refactored classes make the concurrent rendering and thread locking/waiting far easier to understand.

chunky/src/java/se/llbit/chunky/renderer/RenderManager.java
Stenodyon: Whoops, I thought my IDE automatically cleaned imports.

chunky/src/java/se/llbit/chunky/renderer/FrameSampler.java
Stenodyon: Good call, initially I just needed the numeric values for loading and saving render dumps but the enum will be useful in the `Scene` too, to specify the sampling strategy.

chunky/src/java/se/llbit/chunky/renderer/FrameSampler.java
```diff
@@ -145,15 +189,17 @@ protected final void postprocess(double[] pixel, double exposure, Postprocess po
...
+      case UNIFORM:
         return new UniformFrameSampler(inStream);
-      case SAMPLER_TYPE_ADAPTIVE:
+      case ADAPTIVE:
         throw new NotImplementedException();
       default:
-        throw new IOException("Invalid frame sampler type: " + (int)samplerType);
+        // if the tag is not a valid type, Type.from() will throw an exception, otherwise we're
+        // guaranteed a valid Type, which means this default statement should never be reached.
+        throw new RuntimeException("Reached unreachable statement");
```
leMaik, line 202: How (other than adding new enum values) could this be reached?

leMaik, line 202: I guess this can be removed then.

chunky/src/java/se/llbit/chunky/renderer/Renderer.java
```diff
@@ -78,25 +78,25 @@
...
 
   /**
-   * Saves a snapshot of the current frame to the given directory.
-   * @param directory the directory in which the snapshot will be saved
+   * Saves a snapshot of the current frame.
+   * @param outputFile the file in which the snapshot will be saved
    * @param taskTracker progress tracking for this save operation
    */
-  void saveSnapshot(File directory, TaskTracker taskTracker);
+  void saveSnapshot(File outputFile, TaskTracker taskTracker);
```
leMaik, line 99: Reminder: This will require updating all callers of `saveSnapshot`

chunky/src/java/se/llbit/chunky/renderer/FrameSampler.java
Stenodyon: I didn't have this at first but without it Java complains that all code paths don't return a value :confused:

073TheBoss:
Related to #150

NJDaeger:
How's this coming along?

Stenodyon:
I'm in the middle of reworking the `Scene` class, I basically moved all the render/renderer state to appropriate classes but there are so many things going through the `Scene` (like there's an ad hoc event system to control the `RenderManager` in there) that I have to move/adapt a lot of things at once that won't fit in nice atomic commits.

leMaik:
I'm closing this in favor of #786 which will add the foundations for adaptive sampling.

Thanks a lot for your work anyway, it inspired a lot of the recent refactoring efforts! :tada:

PR #622, opened 2020-06-28
Align selection rectangle with chunks

### LINKED ISSUE #614: Box selecting the chunk is not positioned correctly
Hi, when selecting chunks, the box surrounding the chunk you're trying to select doesn't align with the chunk you are selecting. See images

image

image

aTom3333:
I was working a bit on the map drawing today so aligned the selection.
Fixes #614 
The fix is a bit strange, especially the part where I had  to offset by one block for x but not for z but it seems to work fine


leMaik:
Thank you @aTom3333! :raised_hands: I really couldn't unsee it.

Issue #624, opened 2020-07-06
ArrayIndexOutOfBoundsException: -3 when rendering a large scene (around 600k chunks)


stormboomer:
When trying to render a large scene (around 600k Chunks) from a scene file only (no pregenerated octree) I am getting an ArrayIndexOutOfBoundsException.

This seems to have to do with the new PackedOctree generation. When using NodeBasedOctree the issue does not seem to occur.

I think that this might have to do with the switch between implementation of NodeBasedOctree and PackedOctree, the switch happens around 400k Chunks

```
Loading chunks: 76.4% (460,741 of 602,866)
Octree is too big, falling back to old (slower and bigger) implementation.
Loading chunks: 100.0% (602,866 of 602,866)
Loading chunks took 53m 9s1)
Finalizing octree took 0m 0s          
Loading scene took 53m 26s

Unchecked exception caused Chunky to close.
java.lang.NullPointerException
	at se.llbit.math.NodeBasedOctree.get(NodeBasedOctree.java:133)
	at se.llbit.math.NodeBasedOctree.getMaterial(NodeBasedOctree.java:141)
	at se.llbit.math.Octree.getMaterial(Octree.java:292)
	at se.llbit.chunky.renderer.scene.OctreeFinalizer.processBlock(OctreeFinalizer.java:58)
	at se.llbit.chunky.renderer.scene.OctreeFinalizer.finalizeChunk(OctreeFinalizer.java:50)
	at se.llbit.chunky.renderer.scene.Scene.loadChunks(Scene.java:969)
	at se.llbit.chunky.renderer.scene.Scene.loadScene(Scene.java:508)
	at se.llbit.chunky.renderer.scene.SynchronousSceneManager.loadScene(SynchronousSceneManager.java:146)
	at se.llbit.chunky.main.Chunky.doHeadlessRender(Chunky.java:164)
	at se.llbit.chunky.main.Chunky.main(Chunky.java:213)

```

leMaik:
This issue was probably caused by a bug when converting the octree from packed to node-based. Since the conversion now uses the store and load methods, this should be fixed.

PR #627, opened 2020-07-11
Bigpacked octree

aTom3333:
Here comes another octree implementation.
Work in the same spirit as the packed octree but using several arrays as to not be limited by the maximum size of an array in java as we discussed earlier this week.
I also added a field in the advanced tab to select the implementation to be used (as well as saving the choice in the scene and as a persistent setting) to make it easier to change the implementation (and as i may have ideas for one more implementation with different trade off)

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -38,7 +38,6 @@ import se.llbit.chunky.renderer.WorkerState;
 import se.llbit.chunky.renderer.projection.ProjectionMode;
 import se.llbit.chunky.resources.BitmapImage;
 import se.llbit.chunky.resources.OctreeFileFormat;
-import se.llbit.chunky.ui.render.MaterialsTab;
 import se.llbit.chunky.world.Biomes;
 import se.llbit.chunky.world.Chunk;
 import se.llbit.chunky.world.ChunkPosition;
@@ -60,7 +59,6 @@ import se.llbit.math.*;
 import se.llbit.math.primitive.Primitive;
 import se.llbit.nbt.CompoundTag;
 import se.llbit.nbt.ListTag;
-import se.llbit.nbt.StringTag;
 import se.llbit.nbt.Tag;
 import se.llbit.png.ITXT;
 import se.llbit.png.PngFileWriter;
@@ -300,6 +298,11 @@ public class Scene implements JsonSerializable, Refreshable {
   private boolean forceReset = false;
 
   /**
+   * The octree implementation to use
+   */
+  private Octree.ImplementationEnum octreeImplementation = Octree.ImplementationEnum.PACKED;
+
+  /**
    * Creates a scene with all default settings.
    *
    * <p>Note: this does not initialize the render buffers for the scene!
@@ -312,8 +315,8 @@ public class Scene implements JsonSerializable, Refreshable {
     sppTarget = PersistentSettings.getSppTargetDefault();
 
     palette = new BlockPalette();
-    worldOctree = new Octree(1);
-    waterOctree = new Octree(1);
+    worldOctree = new Octree(octreeImplementation, 1);
+    waterOctree = new Octree(octreeImplementation, 1);
   }
 
   /**
@@ -428,6 +431,8 @@ public class Scene implements JsonSerializable, Refreshable {
       alphaChannel = other.alphaChannel;
       samples = other.samples;
     }
+
+    octreeImplementation = other.octreeImplementation;
   }
 
   /**
@@ -731,8 +736,8 @@ public class Scene implements JsonSerializable, Refreshable {
 
       // Create new octree to fit all chunks.
       palette = new BlockPalette();
-      worldOctree = new Octree(requiredDepth);
-      waterOctree = new Octree(requiredDepth);
+      worldOctree = new Octree(octreeImplementation, requiredDepth);
+      waterOctree = new Octree(octreeImplementation, requiredDepth);
 
       // Parse the regions first - force chunk lists to be populated!
       Set<ChunkPosition> regions = new HashSet<>();
@@ -1741,12 +1746,12 @@ public class Scene implements JsonSerializable, Refreshable {
         long fileTimestamp = context.fileTimestamp(fileName);
         OctreeFileFormat.OctreeData data;
         try (DataInputStream in = new DataInputStream(new GZIPInputStream(context.getSceneFileInputStream(fileName)))) {
-          data = OctreeFileFormat.load(in);
+          data = OctreeFileFormat.load(in, octreeImplementation);
         } catch(PackedOctree.OctreeTooBigException e) {
           // Octree too big, reload file and force loading as NodeBasedOctree
           Log.warn("Octree was too big when loading dump, reloading with old (slower and bigger) implementation.");
           DataInputStream inRetry = new DataInputStream(new GZIPInputStream(context.getSceneFileInputStream(fileName)));
-          data = OctreeFileFormat.load(inRetry, true);
+          data = OctreeFileFormat.load(inRetry, Octree.ImplementationEnum.NODE);
         }
         worldOctree = data.worldTree;
         worldOctree.setTimestamp(fileTimestamp);
@@ -2275,6 +2280,7 @@ public class Scene implements JsonSerializable, Refreshable {
     if (!actorArray.isEmpty()) {
       json.add("actors", actorArray);
     }
+    json.add("octreeImplementation", octreeImplementation.ordinal());
 
     return json;
   }
@@ -2559,6 +2565,10 @@ public class Scene implements JsonSerializable, Refreshable {
         actors.add(entity);
       }
     }
+
+    octreeImplementation = Octree.ImplementationEnum.values()[
+      json.get("octreeImplementation").asInt(Octree.ImplementationEnum.PACKED.ordinal())
+    ];
   }
 
   /**
@@ -2798,6 +2808,14 @@ public class Scene implements JsonSerializable, Refreshable {
 
   public void setYClipMax(int yClipMax) {
     this.yClipMax = yClipMax;
+  }
+
+  public Octree.ImplementationEnum getOctreeImplementation() {
+    return octreeImplementation;
+  }
+
+  public void setOctreeImplementation(Octree.ImplementationEnum octreeImplementation) {
+    this.octreeImplementation = octreeImplementation;
   }
 
 }
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -300,7 +300,8 @@ public class Scene implements JsonSerializable, Refreshable {
   /**
    * The octree implementation to use
    */
-  private Octree.ImplementationEnum octreeImplementation = Octree.ImplementationEnum.PACKED;
+  private Octree.ImplementationEnum octreeImplementation
+          = Octree.ImplementationEnum.values()[PersistentSettings.getOctreeImplementation()];
 
   /**
    * Creates a scene with all default settings.
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -300,8 +300,7 @@ public class Scene implements JsonSerializable, Refreshable {
   /**
    * The octree implementation to use
    */
-  private Octree.ImplementationEnum octreeImplementation
-          = Octree.ImplementationEnum.values()[PersistentSettings.getOctreeImplementation()];
+  private String octreeImplementation = PersistentSettings.getOctreeImplementation();
 
   /**
    * Creates a scene with all default settings.
@@ -1752,7 +1751,7 @@ public class Scene implements JsonSerializable, Refreshable {
           // Octree too big, reload file and force loading as NodeBasedOctree
           Log.warn("Octree was too big when loading dump, reloading with old (slower and bigger) implementation.");
           DataInputStream inRetry = new DataInputStream(new GZIPInputStream(context.getSceneFileInputStream(fileName)));
-          data = OctreeFileFormat.load(inRetry, Octree.ImplementationEnum.NODE);
+          data = OctreeFileFormat.load(inRetry, "NODE");
         }
         worldOctree = data.worldTree;
         worldOctree.setTimestamp(fileTimestamp);
@@ -2281,7 +2280,7 @@ public class Scene implements JsonSerializable, Refreshable {
     if (!actorArray.isEmpty()) {
       json.add("actors", actorArray);
     }
-    json.add("octreeImplementation", octreeImplementation.ordinal());
+    json.add("octreeImplementation", octreeImplementation);
 
     return json;
   }
@@ -2567,9 +2566,7 @@ public class Scene implements JsonSerializable, Refreshable {
       }
     }
 
-    octreeImplementation = Octree.ImplementationEnum.values()[
-      json.get("octreeImplementation").asInt(Octree.ImplementationEnum.PACKED.ordinal())
-    ];
+    octreeImplementation = json.get("octreeImplementation").asString(Octree.DEFAULT_IMPLEMENTATION);
   }
 
   /**
@@ -2811,11 +2808,11 @@ public class Scene implements JsonSerializable, Refreshable {
     this.yClipMax = yClipMax;
   }
 
-  public Octree.ImplementationEnum getOctreeImplementation() {
+  public String getOctreeImplementation() {
     return octreeImplementation;
   }
 
-  public void setOctreeImplementation(Octree.ImplementationEnum octreeImplementation) {
+  public void setOctreeImplementation(String octreeImplementation) {
     this.octreeImplementation = octreeImplementation;
   }
 
--- a/chunky/src/java/se/llbit/chunky/resources/OctreeFileFormat.java
+++ b/chunky/src/java/se/llbit/chunky/resources/OctreeFileFormat.java
@@ -32,9 +32,9 @@ public class OctreeFileFormat {
    * Load octrees and grass/foliage textures from a file.
    *
    * @param in input stream for the file to load the scene from.
-   * @param forceNodeBased Forces the octrees to be loaded as NodeBasedOctree
+   * @param impl The octree implementation to use
    */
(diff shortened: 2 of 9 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -299,6 +297,12 @@
 
   private boolean forceReset = false;
 
+  /**
+   * The octree implementation to use
+   */
+  private Octree.ImplementationEnum octreeImplementation
```
leMaik, line 303: The enum can just be named `Octree.Implementation`, having the enum end with enum sounds redundant. Or maybe `ImplementationType`?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2275,6 +2281,7 @@ public void setUseCustomWaterColor(boolean value) {
     if (!actorArray.isEmpty()) {
       json.add("actors", actorArray);
     }
+    json.add("octreeImplementation", octreeImplementation.ordinal());
```
leMaik, line 2284: A safer (and more human-readable) way to save the octree implementation would be to save the `octreeImplementation.name()` and then load it with `Octree.ImplementationEnum.valueOf(json.get("octreeImplementation")`. This is what's used to save all other enums in the scene config.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2559,6 +2566,10 @@ public synchronized void importFromJson(JsonObject json) {
         actors.add(entity);
       }
     }
+
+    octreeImplementation = Octree.ImplementationEnum.values()[
+      json.get("octreeImplementation").asInt(Octree.ImplementationEnum.PACKED.ordinal())
```
leMaik, line 2571: See above. Also remember to log a warning if the octreeImplementation from the scene file is unknown and fall back to one known implementation. As it is, it would crash if the user puts in invalid values or opens a newer scene file with a future octree format with an older chunky version.

chunky/src/java/se/llbit/chunky/resources/OctreeFileFormat.java
```diff
@@ -32,9 +32,9 @@
    * Load octrees and grass/foliage textures from a file.
    *
    * @param in input stream for the file to load the scene from.
-   * @param forceNodeBased Forces the octrees to be loaded as NodeBasedOctree
+   * @param impl The octree implementation to use
    */
-  public static OctreeData load(DataInputStream in, boolean forceNodeBased) throws IOException {
+  public static OctreeData load(DataInputStream in, Octree.ImplementationEnum impl) throws IOException {
```
leMaik, line 37: Now that I think about it… instead of having an enum with possible octree implementations and passing that around, what about an `OctreeFactory` that can be configured to return different octree implementations or switches altogether (e.g. by a plugin)?

Might be too over-engineered though.

chunky/src/java/se/llbit/chunky/ui/render/AdvancedTab.java
```diff
@@ -105,6 +106,19 @@ public AdvancedTab() throws IOException {
       PersistentSettings.setNumRenderThreads(value);
       renderControls.showPopup("This change takes effect after restarting Chunky.", renderThreads);
     });
+
+    octreeImplementation.getItems().addAll(Octree.ImplementationEnum.values());
+    octreeImplementation.getSelectionModel().selectedItemProperty()
+      .addListener((observable, oldvalue, newvalue) -> {
+        scene.setOctreeImplementation(newvalue);
+        PersistentSettings.setOctreeImplementation(newvalue.ordinal());
```
leMaik, line 114: Shouldn't the octree also be changed here? Or is it sufficient to do that on next chunk load? If so, maybe add a note that the new octree is used when reloading the chunks.

chunky/src/java/se/llbit/math/BigPackedOctree.java
```diff
@@ -0,0 +1,773 @@
...
+   *    and the lower 32 bits are the node data (that way, for a leaf node the sign bit will be set
+   *    and the long can simply be compared with 0 to determine if it is a branch or a leaf)
+   *
+   */
+  private ArrayList<long[]> treeData = new ArrayList<>();
+
+  /**
+   * The max size of an array we allow is a bit less than the max value an integer can have
+   */
+  private static final int MAX_ARRAY_SIZE = 1 << 30; // MAX_INT is 2^31 - 1 but it is useful to use a power of 2 so we use the biggest power of 2 that is small that MAX_INT
```
leMaik, line 43: *smaller than

chunky/src/java/se/llbit/math/BigPackedOctree.java
```diff
@@ -0,0 +1,773 @@
...
+  private final long SUB_ARRAY_MASK = MAX_ARRAY_SIZE - 1;
+  private final long FULL_ARRAY_MASK = ~SUB_ARRAY_MASK;
```
leMaik, line 45: ^-- these two could also be static final.

chunky/src/java/se/llbit/math/BigPackedOctree.java
```diff
@@ -0,0 +1,773 @@
...
+//        System.out.printf("%d at level %d\n", getAt(nodeIndex), level);
+        level -= 1;
+        lx = x >>> level;
+        ly = y >>> level;
+        lz = z >>> level;
+        nodeIndex = getAt(nodeIndex) + (((lx & 1) << 2) | ((ly & 1) << 1) | (lz & 1));
+      }
+//      System.out.printf("%d at level %d\n", getAt(nodeIndex), level);
+
+      //System.exit(0);
```
leMaik, line 520: The comments and old debug output can probably be removed.

chunky/src/java/se/llbit/math/BigPackedOctree.java
```diff
@@ -0,0 +1,773 @@
...
+    ray.n.set(nx, ny, nz);
+    ray.distance += tMin;
+    return true;
+  }
+
+  /**
+   *  {@inheritDoc}
+   */
+  @Override
+  public boolean enterBlock(Scene scene, Ray ray, BlockPalette palette) {
```
leMaik, line 486: déjà vu… is this implementation always the same regardless of the octree implementation? If so, we could move it up to the interface as a default implementation.

chunky/src/java/se/llbit/math/BigPackedOctree.java
```diff
@@ -0,0 +1,773 @@
...
+      ray.n.set(nx, ny, nz);
+      ray.distance += tNear;
+    }
+  }
+
+  /**
+   *  {@inheritDoc}
+   */
+  @Override
+  public boolean exitWater(Scene scene, Ray ray, BlockPalette palette) {
```
leMaik, line 606: same here

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -379,7 +369,7 @@ public void pack() {
     if(usePacked) {
       if(implementation instanceof NodeBasedOctree) {
         try {
-          implementation = new PackedOctree(implementation.getDepth(), ((NodeBasedOctree) implementation).root);
+          implementation = new BigPackedOctree(implementation.getDepth(), ((NodeBasedOctree) implementation).root);
```
leMaik, line 372: Should the selected packed octree implementation depend on the octree size (= node count)?

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -393,4 +383,42 @@ private void switchToNodeBased() {
     }
   }
 
+
+  static {
+    factories.put(ImplementationEnum.NODE, new ImplementationFactory() {
```
leMaik, line 388: Allright, so we _do_ have the factory pattern here? Awesome! :tada: 

So we could actually allow plugins to register their own octree factories… but the enum prevents us from selecting them. Maybe use a string instead of the enum? :thinking:

lib/src/se/llbit/chunky/PersistentSettings.java
```diff
@@ -413,5 +413,14 @@ public static void changeSettingsDirectory(File directory) {
...
     settings.load(settingsFile);
   }
+
+  public static void setOctreeImplementation(int ordinal) {
+    settings.setInt("octreeImplementation", ordinal);
+    save();
+  }
+
+  public static int getOctreeImplementation() {
+    return settings.getInt("octreeImplementation", 1); // Packed Octree hardcoded as an int, not nice
```
leMaik, line 423: That's why we should save it as a string in the settings. :upside_down_face:

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
aTom3333: Yeah, the name is not the best, i just didn't want to cause confusion with the interface `Octree.OctreeImplementation`

chunky/src/java/se/llbit/math/BigPackedOctree.java
aTom3333: oops ^

chunky/src/java/se/llbit/math/BigPackedOctree.java
aTom3333: Most of the code is the same and should even be put into `Octree` itself, the only things that changes is the internal representation of a node and how to access children.
I think the best way to do that would be to add some `NodeId` interface that the `Octree` could manipulate and let its implementation do the specific stuff (guess i was too lazy to implement that)

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -364,66 +374,53 @@ public int getDepth() {
...
+    AtomicReference<OctreeImplementation> newOctree = new AtomicReference<>();
+    PipedOutputStream out = new PipedOutputStream();
+    PipedInputStream in = new PipedInputStream(out);
+    Thread readerThread = new Thread(() -> {
+      try {
+        OctreeImplementation impl = factory.loadWithNodeCount(nodeCount, new DataInputStream(in));
+        newOctree.set(impl);
+      } catch(IOException e) {
+        e.printStackTrace();
       }
```
aTom3333, line 402: I have a lot of try/catch around here for exceptions that i don't want to leak (well i don't want to have to annotate every method with a `throws` clause but i would be ok rethrowing them as runtime exception). This is not very nice, what should i do?

leMaik: Why does this method spawn a new thread and then join it? It could just do the work in the current thread instead and the caller could decide which thread to run this in. That way, you could just throw the IOException (I think it's fine to allow throwing IOExceptions if octree conversion fails).

aTom3333: The thread is here so i can use PipedStream, this kind of stream block when the buffer is full/empty and so the consumer and the producer must be on different threads.
If instead i go with the save in temporary file and reload, i wouldn't need a thread

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
aTom3333: This type does not exist anymore

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
aTom3333: Now using a string directly

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
aTom3333: Done. As i have implemented, the scene will hold the unknown/invalid value and, at construction, the octree will warn and fallback

chunky/src/java/se/llbit/chunky/resources/OctreeFileFormat.java
aTom3333: Changed, a string is passed around which refers to a factory that must be registered before (possibly by a plug-in)

chunky/src/java/se/llbit/chunky/ui/render/AdvancedTab.java
aTom3333: Changing the octree could potentially be a heavy operation so i want the user to manually reload chunks (like when changing y min/max). Note added in the tooltip

chunky/src/java/se/llbit/math/BigPackedOctree.java
aTom3333: Fixed

chunky/src/java/se/llbit/math/BigPackedOctree.java
aTom3333: Fixed

chunky/src/java/se/llbit/math/BigPackedOctree.java
aTom3333: Fixed

chunky/src/java/se/llbit/math/BigPackedOctree.java
aTom3333: So, i've implemented what i said in my last commit. I'm marking this as resolved but have a look at the commit in question

chunky/src/java/se/llbit/math/BigPackedOctree.java
aTom3333: Same

chunky/src/java/se/llbit/math/Octree.java
aTom3333: This function no longer exist (even before that it wasn't used beyond my tests)

chunky/src/java/se/llbit/math/Octree.java
aTom3333: I've done that, using a string to select the factory to use and new factories can be registered using the `addImplementationFactory` method of `Octree`

lib/src/se/llbit/chunky/PersistentSettings.java
aTom3333: Changed

chunky/src/java/se/llbit/math/Octree.java
aTom3333: I changed this function to use a temporary file on disk as we discussed

chunky/src/java/se/llbit/chunky/ui/render/AdvancedTab.java
```diff
@@ -105,6 +106,20 @@ public AdvancedTab() throws IOException {
       PersistentSettings.setNumRenderThreads(value);
       renderControls.showPopup("This change takes effect after restarting Chunky.", renderThreads);
     });
+
+    octreeImplementation.getItems().addAll("NODE", "PACKED", "BIGPACKED");
```
leMaik, line 110: These have to be dynamic now, something like `Octree.getImplementations()` which returns the keys of the hashmap?

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -46,22 +45,31 @@
...
+    int getData(NodeId node);
+  }
 
-    /**
-     * Intersects the ray with the geometry inside the octree.
-     *
-     * @return {@code false} if the ray did not hit the geometry
-     */
-    boolean enterBlock(Scene scene, Ray ray, BlockPalette palette);
+  public interface NodeId {}
```
leMaik, line 57: an empty interface?

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -260,7 +259,11 @@ public synchronized void set(int type, int x, int y, int z) {
     } catch(PackedOctree.OctreeTooBigException e) {
       // Octree is too big, switch implementation and retry
       Log.warn("Octree is too big, falling back to old (slower and bigger) implementation.");
-      switchToNodeBased();
+      try {
+        switchImplementation("NODE");
+      } catch(IOException ioException) {
+        ioException.printStackTrace();
```
leMaik, line 265: You should rethrow this as a runtime exception (`throw new RuntimeException(e, "Couldn't switch the octree implementation to NODE")`).

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -276,7 +279,11 @@ public synchronized void set(Node data, int x, int y, int z) {
...
```
leMaik, line 285: See above; let's throw exceptions

chunky/src/java/se/llbit/math/NodeBasedOctree.java
```diff
@@ -151,334 +184,52 @@ public void store(DataOutputStream out) throws IOException {
...
-          }
-          return true;
-        } else if (currentBlock != Air.INSTANCE) {
-          TexturedBlockModel.getIntersectionColor(ray);
-          return true;
-        } else {
-          return true;
-        }
+  static public void initImplementation() {
+    Octree.addImplementationFactory("NODE", new Octree.ImplementationFactory() {
```
leMaik, line 213: Since `"NODE"` is used all over the place, maybe we should move it into a constant (`NodeBasedOctree.NAME`)? Not sure about it, though.

chunky/src/java/se/llbit/math/PackedOctree.java
```diff
@@ -692,6 +360,13 @@ public static PackedOctree load(DataInputStream in) throws IOException {
     return tree;
   }
 
+  public static PackedOctree loadWithNodeCount(long nodeCount, DataInputStream in) throws IOException {
```
leMaik, line 363: Is this an optimization because it can pre-allocate the array?

chunky/src/java/se/llbit/math/Octree.java
aTom3333: It just exist to allow octree to manipulate it, it is given to the implementation when information is needed so the method to handle that are in the implementation

leMaik: Oh, allright! :+1:

chunky/src/java/se/llbit/math/PackedOctree.java
aTom3333: Yes and we could exit early if there is too much nodes.
I can also imagine this being useful for a new implementation that is read only (can only be constructed from an existing and finalized octree)

leMaik:
This should fix #624

chunky/src/java/se/llbit/chunky/ui/render/AdvancedTab.java
aTom3333: Description is created dynamically too

PR #631, opened 2020-07-14
Memory usage reduction

aTom3333:
I made some changes to get some free memory usage reduction by flattening some arrays here and there (and removing a copy of the water normal that was loaded twice)
I measured a 100MB memory usage reduction on the scene i tried but it most likely isn't proportional to the scene size as I mostly changed code about the textures (the change on chunk texture could scale with size of the scene)

### DIFF
--- a/chunky/src/java/se/llbit/chunky/model/WaterModel.java
+++ b/chunky/src/java/se/llbit/chunky/model/WaterModel.java
@@ -73,7 +73,7 @@ public class WaterModel {
   static final double height[] =
       {14 / 16., 12.25 / 16., 10.5 / 16, 8.75 / 16, 7. / 16, 5.25 / 16, 3.5 / 16, 1.75 / 16};
 
-  private static final float[][][] normalMap;
+  private static final float[] normalMap;
   private static final int normalMapW;
 
   /**
@@ -85,7 +85,7 @@ public class WaterModel {
     // precompute normal map
     Texture waterHeight = new Texture("water-height");
     normalMapW = waterHeight.getWidth();
-    normalMap = new float[normalMapW][normalMapW][2];
+    normalMap = new float[normalMapW*normalMapW*2];
     for (int u = 0; u < normalMapW; ++u) {
       for (int v = 0; v < normalMapW; ++v) {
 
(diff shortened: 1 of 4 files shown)


chunky/src/java/se/llbit/chunky/resources/Texture.java
```diff
@@ -1033,7 +1035,10 @@ public void getColor(Ray ray) {
    * @return color
    */
   public final float[] getColor(int x, int y) {
-    return linear[width * y + x];
+    // TODO Not allocate here
```
leMaik, line 1038: I think all other `getColor` methods just assume that the returned array is never altered and don't allocate a new array.

chunky/src/java/se/llbit/chunky/model/WaterModel.java
```diff
@@ -73,7 +73,7 @@
   static final double height[] =
       {14 / 16., 12.25 / 16., 10.5 / 16, 8.75 / 16, 7. / 16, 5.25 / 16, 3.5 / 16, 1.75 / 16};
 
-  private static final float[][][] normalMap;
+  private static final float[] normalMap;
```
leMaik, line 76: The memory savings are by not using nested arrays?

aTom3333: Yes, i explained it on discord but i restate here as it offers more visibility.
Every Java object has a header of a given size that is used by the jvm to do its stuff. The exact size of this header depends on the jvm but for example sake let's say it is 16 bytes (even it the size is different, the idea stays the same).
Arrays are object and have this header. In addition arrays store their length which is an additional 4 bytes that us not needed when the size is a constant. (And for arrays of `long`, `double`, or objects, we can add another 4 bytes so array data is aligned on a multiple of 8 bytes)
For example this variable (when loading a normal map of 512*512) uses 24 + 512 * (24 + 512 * (20 + 2*4)) bytes (7,352,344 bytes).
By using a flat array, the new size is 24 + 512*512*2*4 bytes (2,097,152 bytes), saving a lot of memory.

chunky/src/java/se/llbit/chunky/resources/Texture.java
aTom3333: I'm not sure i get what you are saying, can you elaborate?
All the other `getColor` methods end up calling this one. I made that comment thinking that allocating a new array isn't good on performances, a better way would likely be to pass an array as an output parameter but that would need changing every model class (and more)

leMaik: It didn't allocate a new array before, so it probably doesn't need to now.

aTom3333: Oh ok. The problem is caller expect an array with 4 elements containing r,g,b,a at indices 0,1,2,3 but now the component they want isn't in another index so i can't give them the array. If there was a way to give them as array object that acts like a view of some bigger array without allocating that would do the job but i don't think it exists

leMaik: Even then, allocating that object would probably be more expensive than allocating an array of 4 floats. It's fine, I'll merge it. :+1:

aTom3333: That's true

PR #637, opened 2020-07-17
Fix underwater emitter issue

### LINKED ISSUE #528: 2.0/2.1 Underwater Emitters don't illuminate blocks correctly
Underwater emitters do not illuminate blocks and only illuminate the water surface. I have yet to test other conditions.

**Expected results (1.4.6beta2) -** 
superflat_1 12_uw_emitters-1024

**Results from 2.0beta2 -** 
superflat_1 13_uw_emitters-1024

Credit to _Lohfrum_Faleign_ for discovering this.

aTom3333:
While stepping into the path tracing code, i saw that a ray had its previous material set as air even though it was in water, so i added a test to set its previous material to water. I don't really understand why but it fixes the issue with underwater emitters not lighting the block around.
Fixes #528 

Render with fix:
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -654,6 +654,8 @@ public class Scene implements JsonSerializable, Refreshable {
           ray.setPrevMaterial(r.getPrevMaterial(), r.getPrevData());
           ray.setCurrentMaterial(r.getCurrentMaterial(), r.getCurrentData());
           hit = true;
+        } else if(ray.getPrevMaterial() == Air.INSTANCE) {
+          ray.setPrevMaterial(Water.INSTANCE, 1 << Water.FULL_BLOCK);
         }
       }
     } else {


jackjt8:
Does this resolve the issue for **Test 2** where the camera is above the waters surfacing looking under? This would be a much more common scenario compared to the flawed, but more apparent Test 1. I don't believe I included the scene files for Test 2, but just raising the camera up and looking down into the water should be enough. Emittance/exposure might need to be raised a touch as it's less noticable.

1.4.6beta2
image

2.0beta4/5
image

---

The only main difference I am seeing is the underside of the waters surface does not seem to be directly illuminated by the glowstone and is rather only receiving a reflection. This would be a small trade-off towards resolving this emittance issue.

aTom3333:
Yes it does fix the issue too when the camera is above the surface
Before:
above-surface-underwater-emitter-before-change

After:
above-surface-underwater-emitter-after-change

> The only main difference I am seeing is the underside of the waters surface does not seem to be directly illuminated by the glowstone and is rather only receiving a reflection. This would be a small trade-off towards resolving this emittance issue.

Its seems to be a separate issue that has been introduced since your tests. When rendering Test 1 without my changes i get this:
below-surface-underwater-emitter-before-change
And a 16spp version where we see the pixel of the surface are less bright compared to the ones on your test:
below-surface-underwater-emitter-before-change-16spp

jackjt8:
Hmm.. I'll have to look through my collection of snapshots but I should be able to narrow down the changes that caused that. Otherwise this looks good.

aTom3333:
Hmmm, i had a quick look. I went back to the commit before the separation between world octree and water octree (that caused this bug) and i can't reproduce the same king of render, even with full emitter intensity
old-version-underwater-emitter
I don't know how you got that much light on the water surface

jackjt8:
Much confusion.

1.4.5
superflat_1 12_uw_emitters-512

I'm not sure what I did either. Strange.

jackjt8:
Right! I didn't test 1.4.5 in #528 

1.4.5 does not see undersurface illumination (500SPP).


1.4.6_beta2 does (500SPP).
1 4 6b2

As with 2.0beta6 and 2.2.0-323 (only 16 SPP, so ignore the noise).
image

aTom3333:
I don't have 2.2.0-323, but i have 2.2.0-322 and i still don't have the same illumination
image

jackjt8:
There is no emitter noise... I would still expect some to be found on the walls. Something strange is going on and the glowstone isn't emitting any light. For you.

2.2.0-319
image

Can you send me a copy of those builds so I can test them my end?

aTom3333:
Sure, here is the 2.2.0-322 build with its version file
[2.2.0-322.zip](https://github.com/llbit/chunky/files/4937745/2.2.0-322.zip)

jackjt8:
image

Much confusion.

aTom3333:
I'll redownload your scene to be sure i didn't change something

aTom3333:
Same version, same scene, different result...
image

aTom3333:
Oh, found the problem x)
I was using minecraft 1.12 as texture pack. Switching to minecraft 1.15 give me the same thing you have
image
The water texture must be different and change of light is reflected

jackjt8:
I would assume that's why 1.4.6beta2 has this issue and not 1.4.5. 1.4.6b2 has support for 1.13 resource packs.

aTom3333:
So, just to confirm, the changes of this PR do not change how the water surface is illuminated
image

Issue #633, opened 2020-07-15
Finalizing Octree accounts for 2/3rds of load times


jackjt8:
YouTube video - https://youtu.be/zpafwXZHzTI

Console log - [log2.txt](https://github.com/llbit/chunky/files/4928017/log2.txt)

`Load selected chunks` at 10 seconds

`Finalizing Octree` at 7m10s

`Loaded 70911 chunks` at 19m44s

---

> leMaikToday at 22:11
> TL;DR disk i/o is not the bottleneck, building the octree isn't either but apparently replacing hidden blocks with stone and calculating the shape of lava and water is

> Also, calculating the grass and foliage textures. That could be parallized, though! :thinking:

aTom3333:
@jackjt8 Could you do some new measurement on that issue? I don't think we have measured the impact of #638 on a bigger scale

jackjt8:
Thankfully I have the same map so I can just approximately select the same number of chunks and see where we end up. Tested on 2.2-392; Unfortunately the `Finalizing Octree` message is no longer printed to the console so this makes testing a bit harder.

`Load selected chunks` at 4 seconds

`Finalizing Octree` at 7m39s? --- Just judging it based upon the sudden shift if CPU behaviour from constantly loading a single thread to jumping between multiple threads.

`Loaded 73886 chunks` at 10m4s

Old - 420s, 754s, 1174s
2.2-392 - 455s, 145s, 600s
(loading, finalizing, total)

---

Comparing to my previous results and what you obtained in your own testing on https://github.com/chunky-dev/chunky/pull/638, with a bit of guess work, I am seeing an uptick in Chunk loading times of around +8.3% and a reduction in Finalizing time of -80% (inline with your numbers)*.

`* Assuming that I am guessing the point where Octree Finalization is starting correctly.

jackjt8:
Turns out running Chunky via Headless gives us more information and still has the Octree Finalization completion line written.

To confirm: Pre-https://github.com/chunky-dev/chunky/pull/638: Octree building is 390s, Finalizing 700s.

Post-https://github.com/chunky-dev/chunky/pull/638: Octree building is ~330s, Finalizing is about 130s.

---

There is some variance in my results but there is a clear and marked improvement to this issue.

PR #638, opened 2020-07-18
Octree finalization optimization

aTom3333:
Related to #633 
Finalizing the octree is slow, especially when dealing with very large scene, because for each block we need to take action depending on its neighboring blocks, this means walking down the octree up to 10 times for each block. Walking down the octree is cheap but not free so, given the number of blocks this ends up costing a lot of time.
The idea of this PR is to do some of the computation requiring neighboring blocks when loading a chunk, when accessing blocks is cheaper as they are directly stored in a flat array. Blocks that are on the edge of a chunk don't have all of their neighbors available so those are still processed in the finalization phase as before.
While implementing this i also added the "whatever" type i suggested on discord that helps merging nodes more efficiently and added hook for octree implementation to be notified of when finalization takes place.

I did some measurement to compare the loading time before and after this PR. I loaded a 4k chunks scene three times without closing chunky in-between
Results:
|     | Loading time | Finalizing time | Total time |
|------|---------------|--------------------|---------------|
| Before, run 1 | 35s | 43s | 78s |
| Before, run 2 | 37s | 39s | 76s |
| Before, run 3 | 38s | 39s | 77s |
| **Before, average** | **37s** | **40s** | **77s** |
| After, run 1 | 27s | 7s | 34s |
| After, run 2 | 75s | 9s | 84s |
| After, run 3 | 39s | 9s | 48s |
| **After, average** | **46s** | **8s** | **55s** |

Results to take with a grain of salt, the variance for the after is huge (during the second run my computer froze for a bit, most likely due to swapping).

I would appreciate if someone with more memory could make some bigger scale tests

### DIFF
--- a/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
+++ b/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
@@ -26,7 +26,7 @@ import java.util.function.Consumer;
 public class BlockPalette {
   private static final int BLOCK_PALETTE_VERSION = 4;
   public final int airId, stoneId, waterId;
-  public static final int WHATEVER_ID = Octree.WHATEVER_TYPE;
+  public static final int ANY_ID = Octree.ANY_TYPE;
 
   private final Map<String, Consumer<Block>> materialProperties;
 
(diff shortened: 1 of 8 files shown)


aTom3333:
I need to do some measure to be sure but the introduction of the "whatever" type, although it does lower the node count (related to memory usage of the octree) in the end (after finalization), the peak node count seems to be higher and i'm not very pleased with that (if you didn't get it by now i like low memory usage :p)

PR #642, opened 2020-07-21
Fix Position & Orientation tab not updating when camera moved in map

### LINKED ISSUE #634: Camera "Position&Orientation" dialog does not auto-refresh
When I have the "Position & Orientation" dialog opened and move a camera on the Map canvas, the dialog does not auto-refresh. If I change the height of the camera after moving the camera while the dialog is still open, the camera resets to its old position before moving

073TheBoss:
Fixes #634


leMaik:
Good catch!

PR #644, opened 2020-07-21
Add a new optional method to OctreeImplementation interface

aTom3333:
This PR adds a new method to octree implementation interface (with a default implementation) that is used in `enterBlock` and `exitWater`. Functionally it changes nothing but it allow octree implementations to implement a different version (more efficient for example, i have an use case for that)

### DIFF
--- a/chunky/src/java/se/llbit/math/Octree.java
+++ b/chunky/src/java/se/llbit/math/Octree.java
@@ -587,6 +587,10 @@ public class Octree {
       NodeId node = nodeAndLevel.getFirst();
       int level = nodeAndLevel.getSecond();
 
+      lx = x >>> level;
+      ly = y >>> level;
+      lz = z >>> level;
+
       // Test intersection
       Block currentBlock = palette.get(implementation.getType(node));
       Material prevBlock = ray.getCurrentMaterial();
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -456,17 +469,13 @@ public boolean enterBlock(Scene scene, Ray ray, BlockPalette palette) {
...
-        lx = x >>> level;
-        ly = y >>> level;
-        lz = z >>> level;
-        node = implementation.getChild(node, (((lx & 1) << 2) | ((ly & 1) << 1) | (lz & 1)));
-      }
+      Pair<NodeId, Integer> nodeAndLevel = implementation.getWithLevel(x, y, z);
+      NodeId node = nodeAndLevel.getFirst();
+      int level = nodeAndLevel.getSecond();
+
+      lx = x >>> level;
```
leMaik, line 476: do we still need the shifts? They were dropped below.

aTom3333: Good catch, yes we do, they are used to advance the ray out of the node later. The code below is wrong

PR #650, opened 2020-07-26
Fix water blocks having full height if any solid block is above it.

leMaik:
Until now, Chunky made water blocks full-height if there was either a water-filled block or a solid block above it. #619 already fixed water rendering below vines by marking them as non-solid (which was correct because vines aren't solid), but it turns out that it doesn't matter if the block is solid: water never has full height if there's is no water-filled block above it.

Before:
image

After:
image


leMaik:
I also noticed that lily pads aren't hovering in Chunky as they do ingame. In chunky, they are right above the water surface while in Minecraft, they are at the bottom of the block above.

If this was changed in Chunky, it would look like this:

image

So I guess we should keep them on the water surface for now and maybe add an option for hovering lily pads later.

PR #656, opened 2020-07-29
Change job ids to long to fix deadlocks after long render times.

### LINKED ISSUE #654: Long Running render Crash
When doing long time renders (2 Days or more) Chunky just hangs up and stops rendering. There is no error Message.

Analysing Heap dump pointed out that the render manager is hitting the Int Limit on the Atomic Integer that is used to provide a new Job id
image

This Issue is probably caused by this
https://github.com/chunky-dev/chunky/blob/af0d633f427c44ca41420a21ba8394534ff2174d/chunky/src/java/se/llbit/chunky/renderer/RenderManager.java#L360

I think we need something to "dequeue" not used Job IDs
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicInteger.html

leMaik:
Fixes #654, I'll wait for confirmation from @stormboomer before merging this PR.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/RenderManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/RenderManager.java
@@ -24,7 +24,7 @@ import se.llbit.util.TaskTracker;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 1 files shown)


stormboomer:
looks good, as far as I can tell I have now hit the old limit, but it is still continueing to render
image

PR #657, opened 2020-07-29
Update chain model for 1.16.2-pre1

073TheBoss:
chain-rotations-1 16 2-pre1 denoised
1.16.2-pre1 included a change to chains that allows for X/Z alignment. This PR adds that functionality to Chunky.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/block/Chain.java
+++ b/chunky/src/java/se/llbit/chunky/block/Chain.java
@@ -6,13 +6,28 @@ import se.llbit.chunky.resources.Texture;
 import se.llbit.math.Ray;
 
 public class Chain extends MinecraftBlockTranslucent {
-  public Chain(String name, Texture texture) {
+
+  private final int axis;
+
+  public Chain(String name, Texture texture, String axis) {
     super(name, texture);
     localIntersect = true;
+    switch(axis) {
+      default:
+      case "y":
+        this.axis = 0;
+        break;
+      case "x":
+        this.axis = 1;
+        break;
+      case "z":
+        this.axis = 2;
+        break;
+    }
   }
 
   @Override
   public boolean intersect(Ray ray, Scene scene) {
-    return ChainModel.intersect(ray);
+    return ChainModel.intersect(ray, axis);
   }
 }
--- a/chunky/src/java/se/llbit/chunky/block/Chain.java
+++ b/chunky/src/java/se/llbit/chunky/block/Chain.java
@@ -7,23 +7,12 @@ import se.llbit.math.Ray;
 
 public class Chain extends MinecraftBlockTranslucent {
 
-  private final int axis;
+  private final String axis;
 
   public Chain(String name, Texture texture, String axis) {
     super(name, texture);
     localIntersect = true;
-    switch(axis) {
-      default:
-      case "y":
-        this.axis = 0;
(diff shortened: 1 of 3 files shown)


chunky/src/java/se/llbit/chunky/block/Chain.java
```diff
@@ -6,13 +6,28 @@
 import se.llbit.math.Ray;
 
 public class Chain extends MinecraftBlockTranslucent {
-  public Chain(String name, Texture texture) {
+
+  private final int axis;
```
leMaik, line 10: Instead of having this magic number here, let's keep it a string here and move the conversion to an array index into the `ChainModel`

chunky/src/java/se/llbit/chunky/block/MinecraftBlockProvider.java
```diff
@@ -2653,6 +2653,11 @@ private static Block chest(Tag tag, boolean trapped) {
     return new Chest(name, type, facing, trapped);
   }
 
+  private static Block chain(Tag tag, String name, Texture texture) {
+    String axis = tag.get("Properties").get("axis").stringValue("north");
```
leMaik, line 2657: The default value for `axis` is not `north` but `y`.

chunky/src/java/se/llbit/chunky/model/ChainModel.java
```diff
@@ -7,7 +7,7 @@
 import se.llbit.math.Vector4;
 
 public class ChainModel {
-  private static final Quad[] quads =
+  private static final Quad[] alignY =
```
leMaik, line 10: align? Why not `quadsY`, `quadsX` and `quadsZ`?

chunky/src/java/se/llbit/chunky/model/ChainModel.java
```diff
@@ -33,11 +33,22 @@
...
+  static {
+    Quad[] alignX = Model.rotateZ(alignY);
+    Quad[] alignZ = Model.rotateX(alignY);
+
+    axis[0] = alignY;
+    axis[1] = alignX;
+    axis[2] = alignZ;
+  }
+
+  public static boolean intersect(Ray ray, int data) {
```
leMaik, line 47: Should be `String axis` instead of `int data` so that the array index number thing is contained in the model class.

chunky/src/java/se/llbit/chunky/block/Chain.java
073TheBoss: Ok, sounds good. I adapted this from the End Rod classes, so should this be changed there too?

chunky/src/java/se/llbit/chunky/block/MinecraftBlockProvider.java
073TheBoss: I guess that's what I get for copy/pasting code instead of retyping it. lol

chunky/src/java/se/llbit/chunky/block/Chain.java
leMaik: Yeah... Legacy code from back when `data` actually _was_ a number (pre-flattening, i.e. 1.12). No need to go through all that code but I tried to use the new pattern for new blocks when adding 1.14-1.16 support.

PR #659, opened 2020-07-30
Emitter sampling

aTom3333:
This PR is what i showed yesterday on discord, i'm making a draft PR to get some feedback if some of you have ideas on how to improve that.

So, what's that?
This adds emitter sampling to the renderer to greatly increase convergence of scene with emitters. The reason usually scenes lit up only by sun and sky converge quickly is because at every intersection, the sun is sampled, adding its contribution to the ray without the need of a ray randomly hitting it. On the other hand emitters are not sampled so to contribute to the ray, we need to randomly hit them and, especially with small emitters like torches, the probability is low.
The idea is then to sample emitter the same way we sample the sun so as to need solely relying on this low probability.
But whereas there is only one sun in the scene, they could be a lot of emitters. Additionally, emitters that are far away don't contribute a lot. Given that it wouldn't be practical to sample every emitter and the fact we only really care about the close ones, i added a new data structure, the `Grid` that holds the position of emitters in each of its cell (let's say a cell is 16*16*16 blocks) (and the position of the emitters that are in neighboring cells to prevent having a block close to an emitter not being lit up by it because they are in different cells)

I tried two sampling strategies. The first ones is to, for each intersection, sample a single emitter, and the second one is to sample every emitters.

One emitter sample for every intersection, at 50SPP:
night-test-50-sample-one-same-view

All emitter sample for every intersection, at 50SPP:
night-test-50-sample-all

Obviously sampling every emitters gives a much better render but it has a big cost. Indeed for every emitter we want to sample, we need to cast a ray from the intersection point to emitter (called a "shadow ray") to check if there is nothing in-between hiding the light. This operation is complex so doing it for every emitter increases drastically the time required for one SPP ( = SPS is very low).
I haven't done enough testing to know what is the best solution but i'll surely do is let the user choose between those strategy (and the no emitter sampling strategy).

For now, this is highly experimental, here is a list of what i need to do before this is complete (sorted by importance i think):
 - Allow saving and loading the `Grid`, i haven't implemented  atm so i need to reload chunks everytime
 - Add ui option to choose the sampling strategy
 - Take fog into account. For now the effects of the fog on the distance traveled from the emitter to the intersection are not accounted for. Although the distance is usually short, this is incorrect

Things i should do but are not mandatory:
 - Improve the `Grid` implementation to be more memory efficient (the same way `PackedOctree` is done)
 - Allow the user to choose the grid's cells size. This is hardcoded for now but it could be interesting for some users to be able to change it as it  would offer different trade-offs
 - Allow the user to indicate that they doesn't want this additional data structure when making a render without emitters (or with emitters but no sampling but i really think this is superior) so as to not waste memory

Things that would be ideal to do but i'm not sure i'm able to:
 - Implement this so it work with every rays, not only rays generated from diffuse reflexion

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/EmitterSamplingStrategy.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/EmitterSamplingStrategy.java
@@ -1,0 +1,5 @@
+package se.llbit.chunky.renderer;
+
+public enum EmitterSamplingStrategy {
+  None, One, All
+}
--- a/chunky/src/java/se/llbit/chunky/renderer/EmitterSamplingStrategy.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/EmitterSamplingStrategy.java
@@ -1,5 +1,5 @@
 package se.llbit.chunky.renderer;
 
 public enum EmitterSamplingStrategy {
-  None, One, All
+  NONE, ONE, ALL
 }
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
@@ -22,10 +22,7 @@ import se.llbit.chunky.block.Water;
 import se.llbit.chunky.model.WaterModel;
 import se.llbit.chunky.renderer.WorkerState;
 import se.llbit.chunky.world.Material;
-import se.llbit.math.QuickMath;
-import se.llbit.math.Ray;
-import se.llbit.math.Vector3;
-import se.llbit.math.Vector4;
+import se.llbit.math.*;
 
 import java.util.Random;
 
@@ -156,6 +153,9 @@ public class PathTracer implements RayTracer {
 
             float emittance = 0;
 
+            Vector4 indirectEmitterColor = new Vector4(0, 0, 0, 0);
+            double indirectEmitterCoef = 0;
+
             if (scene.emittersEnabled && currentMat.emittance > Ray.EPSILON) {
 
               emittance = addEmitted;
@@ -166,6 +166,34 @@ public class PathTracer implements RayTracer {
               ray.emittance.z = ray.color.z * ray.color.z *
                   currentMat.emittance * scene.emitterIntensity;
               hit = true;
+            } else if(scene.emittersEnabled) {
+              // Sample emitter
+              Ray emitterRay = new Ray(ray);
+              Grid.EmitterPosition pos = scene.getEmitterGrid().sampleEmitterPosition((int)emitterRay.o.x, (int)emitterRay.o.y, (int)emitterRay.o.z, random);
+              if(pos != null) {
+                // TODO Sampling a random point on the model would be better than using a random point in the middle of the cube
+                Vector3 target = new Vector3(pos.x + 0.5 + (random.nextDouble() - 0.5) / 8, pos.y + 0.5 + (random.nextDouble() - 0.5) / 8, pos.z + 0.5 + (random.nextDouble() - 0.5) / 8);
+                emitterRay.d.set(target);
+                emitterRay.d.sub(emitterRay.o);
+                double distance = emitterRay.d.length();
+                emitterRay.d.normalize();
+                indirectEmitterCoef = emitterRay.d.dot(emitterRay.n);
+                if(indirectEmitterCoef > 0) {
+                  emitterRay.emittance.set(0, 0, 0);
+                  emitterRay.o.scaleAdd(Ray.EPSILON, emitterRay.d);
+                  PreviewRayTracer.nextIntersection(scene, emitterRay);
+                  if(emitterRay.getCurrentMaterial().emittance > Ray.EPSILON) {
+                    indirectEmitterColor.set(emitterRay.color);
+                    indirectEmitterColor.scale(emitterRay.getCurrentMaterial().emittance);
+                    // TODO Take fog into account
+                    indirectEmitterCoef *= scene.emitterIntensity;
+                    // Not realistic but offer better convergence and is better artistically
+                    indirectEmitterCoef /= Math.max(distance * distance, 1);
+                  }
+                } else {
+                  indirectEmitterCoef = 0;
+                }
+              }
             }
 
             if (scene.sunEnabled) {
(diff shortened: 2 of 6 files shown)


aTom3333:
And here is comparison with the old way with 50SPP as well:
night-test-50-no-sample

aTom3333:
Made another comparison that's more fair by letting the three strategies render for a fixed time (30 mins)
No emitter sampling:
night-test-30min-no-sample
658 spp
287 426 sps

One emitter sample:
night-test-30min-one-sample
606 spp
264 683 sps

All emitter sample:
night-test-30min-all-sample
176 spp
76 714 sps

Something to note, performance depends on the amount of emitters and of how big the areas lit up by emitter are. I'd say only 20% of this render is an area where emitter have an effect, so performance are impacted as much as if emitters were lighting up 100% of the render.
In this sense this kinda acts like adaptive sampling in the sense where more work is done on area lit up by emitter, which tends to have a big variance, than on other areas.

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
```diff
@@ -166,6 +165,22 @@ public static boolean pathTrace(Scene scene, Ray ray, WorkerState state, int add
               ray.emittance.z = ray.color.z * ray.color.z *
                   currentMat.emittance * scene.emitterIntensity;
               hit = true;
+            } else if(scene.emittersEnabled) {
+              // Sample emitter
+              boolean sampleOne = ray.depth != 0;
+              if(sampleOne) {
+                Grid.EmitterPosition pos = scene.getEmitterGrid().sampleEmitterPosition((int) ray.o.x, (int) ray.o.y, (int) ray.o.z, random);
+                if(pos != null) {
+                  // TODO Sampling a random point on the model would be better than using a random point in the middle of the cube
```
leMaik, line 174: This could be done once all models are made out of quads, i.e. when the json block loader is finished one day

aTom3333:
I had a look to make fog work with that but i'm not sure how i would it. This means we can't have god ray from emitters...
Maybe i'll try again later.
Otherwise i implemented what i viewed as necessary for merging, the rest could be implemented later in another PR

jackjt8:
> I had a look to make fog work with that but i'm not sure how i would it. This means we can't have god ray from emitters...

The current fog implementation is only illuminated by the suns positions. It does not factor whether the sun is enabled, emitters, or color information (outside of the fogs base color). While it would be nice if it was supported, it would require a more extensive look at the fog system... I was meaning on creating a new issue showcasing all these issues but I haven't got around to it yet. I'll try and do that now.

Still, I think people would be happy with the faster convergence.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2694,6 +2738,8 @@ public synchronized void importFromJson(JsonObject json) {
     }
 
     octreeImplementation = json.get("octreeImplementation").asString(PersistentSettings.getOctreeImplementation());
+
+    emitterSamplingStrategy = EmitterSamplingStrategy.valueOf(json.get("emitterSamplingStrategy").asString("None"));
```
leMaik, line 2742: Should be `NONE`

chunky/src/java/se/llbit/chunky/renderer/EmitterSamplingStrategy.java
```diff
@@ -0,0 +1,5 @@
+package se.llbit.chunky.renderer;
+
+public enum EmitterSamplingStrategy {
+  None, One, All
```
leMaik, line 4: Enum values should be ALL_UPPERCASE.

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
```diff
@@ -389,6 +402,36 @@ public static boolean pathTrace(Scene scene, Ray ray, WorkerState state, int add
     return hit;
   }
 
+  private static Vector4 sampleEmitter(Scene scene, Ray ray, Grid.EmitterPosition pos, Random random) {
```
leMaik, line 405: Some javadoc on top of this method that explains what it does (i.e. a summary of this PR) and maybe a link to this PR as `@see` would be nice.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -503,7 +506,10 @@ public synchronized void loadScene(RenderContext context, String sceneName,
       mode = RenderMode.PAUSED;
     }
 
-    if (!loadOctree(context, taskTracker)) {
+    // Try loading the grid unconditionally for now
+    boolean gridLoaded = loadGrid(context, taskTracker);
+    boolean octreeLoaded = loadOctree(context, taskTracker);
+    if (!gridLoaded || !octreeLoaded) {
       // Could not load stored octree.
```
leMaik, line 513: (or grid)

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -742,6 +749,7 @@ public synchronized void loadChunks(TaskTracker progress, World world,
       palette = new BlockPalette();
       worldOctree = new Octree(octreeImplementation, requiredDepth);
       waterOctree = new Octree(octreeImplementation, requiredDepth);
+      emitterGrid = new Grid(requiredDepth, 10); // TODO Make configurable
```
leMaik, line 752: Is it an advanced feature that we should add a UI field for in the future or is this more advanced (like e.g. map loader thread count)? If so, maybe add a system property for it (like `chunky.emitterGridCellSize`)?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1864,6 +1893,20 @@ public synchronized void saveDump(RenderContext context, TaskTracker progress) {
     }
   }
 
+  private synchronized boolean loadGrid(RenderContext context, TaskTracker progress) {
```
leMaik, line 1896: Emitter grid? Just in case more grids are added in the future? Same for the filename.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
aTom3333: Changing it can allow sampling emitters more far away so it changes the way the render looks (at the cost of performance and maybe memory usage)

PR #660, opened 2020-07-30
Adjust fogDensity slider to allow for light fog.

jackjt8:
First time coding Java, plus doing it blind...

fogDensity 0.01 is considered light fog & 0.1 heavy. Having a logarithmic slider here was an improvement but I do feel that offering fogDensity between [0.01 - 2] was not optimal.

Changed the fogDensity range to be [0.001 - 1]; covering from very light fog (0.001) > light fog (0.01) > heavy fog (0.1) > very heavy fog (1).

Opted to create setRange2() as to not break dependencies with setRange() and the 0.01 lower limit for logarithmic calculations (seems to work fine with the fogDensity mind you). - I would have used an optional arg here but I wasn't sure the best approach.. so new method?

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
+++ b/chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
@@ -102,9 +102,9 @@ public class SkyTab extends ScrollPane implements RenderControlsTab, Initializab
     cloudZ.onValueChange(value -> scene.sky().setCloudZOffset(value));
 
     fogDensity.setTooltip("Fog thickness. Set to 0 to disable volumetric fog effect.");
-    fogDensity.setRange(0, 2);
-    fogDensity.clampMin();
+    fogDensity.setRange2(0, 1);
     fogDensity.makeLogarithmic();
+    fogDensity.clampMin();
     fogDensity.onValueChange(value -> scene.setFogDensity(value));
 
     skyFogDensity.setTooltip(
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
```diff
@@ -102,9 +102,9 @@ public SkyTab() throws IOException {
     cloudZ.onValueChange(value -> scene.sky().setCloudZOffset(value));
 
     fogDensity.setTooltip("Fog thickness. Set to 0 to disable volumetric fog effect.");
-    fogDensity.setRange(0, 2);
-    fogDensity.clampMin();
+    fogDensity.setRange2(0, 1);
```
leMaik, line 105: Why not add `sliderMin` as third argument to `setRange`?

lib/src/se/llbit/chunky/ui/SliderAdjuster.java
```diff
@@ -64,6 +64,12 @@ public void setRange(double min, double max) {
     }
   }
 
+  public void setRange2(double min, double max) {
```
leMaik, line 67: ...so that this could be removed?

chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
jackjt8: From my limited research optional args aren't really supported by Java which would have been my choice. There is the option of using an overload but I wasn't sure which one would be preferred.

leMaik: Either add another method with a third argument all call that one from the method with two arguments. Or just remove the old one and change all places where it is used (probably not that many in Chunky).

jackjt8: While it is only 33 occurrences, given we only need the additional parameter for this one case I do believe using the overload is the best choice. I'll make the necessary changes such that we have a method setRange(min, max, SliderMin) and I'll remove the setRange2().

Issue #653, opened 2020-07-28
Show Guides on Render Preview missing + Additional guides


jackjt8:
An option missing from Chunky 2.X is the `Show Guides` right click option with the Render Preview.

![image](https://user-images.githubusercontent.com/3740917/88657596-14d9a300-d0ca-11ea-88b8-eb50dfe518fe.png)

![image](https://user-images.githubusercontent.com/3740917/88657663-333f9e80-d0ca-11ea-91d1-9e47cfa3c4b1.png)

---

While eyeballing is an option, having guides is a big help when framing up shots. Additional guides could also prove useful.

Additional guides:
- Golden Ratio - **Might add via a plugin**
- ~~Rule of Quarters~~
- ~~Diagonals?~~

leMaik:
:100: This would be so cool!

jackjt8:
Additional guides (may) be added via a plugin.

PR #661, opened 2020-07-31
Readded rule of third guides

jackjt8:
Fixes the missing guides noted in #653 however more work would be required to add additional.

Used the guide code found in the 1.4.5 release as a template. Hopefully it's up to standard.

image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
+++ b/chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
@@ -19,8 +19,10 @@ package se.llbit.chunky.ui;
 import javafx.application.Platform;
 import javafx.geometry.Bounds;
 import javafx.geometry.Point2D;
+import javafx.scene.Group;
 import javafx.scene.canvas.Canvas;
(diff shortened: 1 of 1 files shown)


leMaik:
Rule of thirds seems to be the common thing. I can't find much on rule of quarters, golden ratio would also mean that we'd need to allow mirroring the guides and change the image resolution to the golden ratio (or move the guides). I find a lot about diagonals, but we can't just overlay a few of them; it's more about putting diagonals _somewhere_ in the image.

I'd say merge this and don't implement the rest of #653

jackjt8:
> Rule of thirds seems to be the common thing. I can't find much on rule of quarters, golden ratio would also mean that we'd need to allow mirroring the guides and change the image resolution to the golden ratio (or move the guides). I find a lot about diagonals, but we can't just overlay a few of them; it's more about putting diagonals _somewhere_ in the image.
> 
> I'd say merge this and don't implement the rest of #653

Fixed the scaling by piggybacking the Guide scaling in updateCanvasScale().
image

Admittedly I was just throwing around ideas and Rule of thirds is the most common. I may add the Golden ratio via a plugin if I can figure it out.

PR #662, opened 2020-07-31
Fix for tall flowers in materials tab

073TheBoss:
Allows tall flowers (Rose bush, Lilac, & Peony) to be changed by the Materials Tab. This was caused by the `MinecraftBlockProviders.largeFlower()` function changing the name of the block before passing it on.


aTom3333:
You may want to look at sunflowers and tall grass too i think

073TheBoss:
Both of those work correctly

leMaik:
Can we change the `MinecraftBlockProviders.largeFlower()` to not append the "half" string to the block name instead? Because that's not actually it's block name (= block ID).

073TheBoss:
Changes have been made.

PR #663, opened 2020-07-31
Fix corrupted chunk loading

### LINKED ISSUE #651: over 50% of chunks do not load in when creating a scene from selected chunks
My prev. issue was closed because someone said it was fixed.  It was not fixed in the latest snapshot and the issue seems to have gotten worse (even less chunks show up) I will now paste in what I said last time and link to the prev. issue allong with post a few more updated photos
  

https://github.com/llbit/chunky/issues/648

The attached screenshots show the issue.

When I first installed everything worked fine, then the next day I decide to make a even bigger render, that's when I first expirenced this bug. Now it does not matter how much ram is allocated it does this and I can only make small renders (much smaller than what I could do before) lastly when I try to make a new scene after it does this I get a error window that says, " warning, can't load chunks right now" I get this box whenever I make a new scene until I restart the software. I tried deleting and reinstalling chunky and even tried different versions (latest normal and latest snapshot) they both do this.

Screenshot (409)
Screenshot (408)
Screenshot (407)
Screenshot (406)
Screenshot (405)

leMaik:
Fixes #651 (for real this time, I tested it with that world) and adds a system property to customize the number of map loader threads (useful for large worlds, i.e. `-Dchunky.mapThreads=32`).

### DIFF
--- a/chunky/src/java/se/llbit/chunky/world/Chunk.java
+++ b/chunky/src/java/se/llbit/chunky/world/Chunk.java
@@ -153,7 +153,10 @@ public class Chunk {
     request.add(Chunk.LEVEL_BIOMES);
     request.add(Chunk.LEVEL_HEIGHTMAP);
     Map<String, Tag> data = getChunkData(request);
-
+    // TODO: improve error handling here.
+    if (data == null) {
+      return;
+    }
     surfaceTimestamp = dataTimestamp;
     version = chunkVersion(data);
     loadSurface(data);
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/chunky/world/Chunk.java
```diff
@@ -153,7 +153,10 @@ public synchronized void loadChunk() {
     request.add(Chunk.LEVEL_BIOMES);
     request.add(Chunk.LEVEL_HEIGHTMAP);
     Map<String, Tag> data = getChunkData(request);
-
+    // TODO: improve error handling here.
```
leMaik, line 156: just like at the other place in this file :grimacing:

PR #666, opened 2020-08-03
Temp fix for not being able to disable fog

jackjt8:
Need to figure out why using double `sliderMin` here leads to a 4.9E-324 in DoubleAdjuster.java but hardcoding doubles doesn't.


jackjt8:
bump

PR #670, opened 2020-08-16
Remove static registries

leMaik:
For [ChunkyMap](https://github.com/leMaik/ChunkyMap) and the upcoming Render Service, I'm using Chunky as a library to render scenes. Having lots of shared (= static) but dynamic data between Chunky instances brings concurrency issues when rendering more than one scene at once.

This PR makes the block registry non-static. The goal is to have the `Chunky` class contain everything that is needed for rendering so that one `Chunky` class can be used per concurrent rendering without problems. Having less global static fields also helps to make the code easier to understand by making dependencies explicit by having to pass them through to other classes or methods.

---

Update: This branch somehow has turned into everything that I changed for use with ChunkyMap and the render service.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/main/Chunky.java
+++ b/chunky/src/java/se/llbit/chunky/main/Chunky.java
@@ -17,9 +17,19 @@
  */
 package se.llbit.chunky.main;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
 import se.llbit.chunky.PersistentSettings;
 import se.llbit.chunky.block.BlockProvider;
-import se.llbit.chunky.block.BlockSpec;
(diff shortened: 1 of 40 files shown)


chunky/src/java/se/llbit/chunky/main/Chunky.java
```diff
@@ -52,21 +62,12 @@
...
-
 /**
- * Chunky is a Minecraft mapping and rendering tool created by
- * Jesper Öqvist (jesper@llbit.se).
+ * Chunky is a Minecraft mapping and rendering tool created by Jesper Öqvist (jesper@llbit.se).
  *
  * <p>Read more about Chunky at <a href="https://chunky.llbit.se">https://chunky.llbit.se</a>.
  */
-public class Chunky {
+public class Chunky implements BlockProviderRegistry {
```
aTom3333, line 70: It feels a bit strange that `Chunky` itself is the `Registry` but it shouldn't be a problem

leMaik: That's why I made it an interface. This allows us to let anything else implement the registry later without much changes

leMaik:
Unless someone tries to render multiple scenes with different block providers in the same process, this isn't needed anymore. Since not even ChunkyCloud does that, I'll close it.

PR #679, opened 2020-08-23
Lectern book

leMaik:
This PR adds rendering of books placed on top of a lectern. The book is not yet customizable but provides everything that is needed to render the enchantment table book.

0000_lectern-100 denoised

It also fixes triangles always being double-sided, even if they belong to single-sided quads, which causes rendering issues if two quads overlap but have their textures on different sides (e.g. the book cover, the book pages and the campfire flame).

|before|after|
|---|---|
|image|image|

### DIFF
--- a/chunky/src/java/se/llbit/chunky/world/model/CubeModel.java
+++ b/chunky/src/java/se/llbit/chunky/world/model/CubeModel.java
@@ -23,7 +23,9 @@ import se.llbit.chunky.resources.TexturePackLoader;
 import se.llbit.chunky.resources.texturepack.SimpleTexture;
 import se.llbit.chunky.resources.texturepack.TextureLoader;
 import se.llbit.log.Log;
+import se.llbit.math.DoubleSidedQuad;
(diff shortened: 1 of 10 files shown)


chunky/src/java/se/llbit/chunky/world/model/CubeModel.java
```diff
@@ -82,47 +84,53 @@ public CubeModel(Collection<Cube> cubes, double uvScale, Map<String, Texture> te
                 new int[] { 1, 1 },
                 new int[][] { { 0, 0 }, { 0, 1 } },
                 new int[][] { { 2, 1 }, { 2, 0 } },
-                uv);
+                uv,
+                Math.abs(cube.start.y - cube.end.y) > Ray.EPSILON);
```
leMaik, line 88: We want double-sided faces if the height is > 0 to e.g. see a helmet from inside the helmet. Same for the other directions. Using single-sided faces for height = 0 allows using the `CubeModel` for entities that have such elements, e.g. the book or the campfire.

PR #701, opened 2020-10-07
Added support for exporting PFM files.

StanleyMines:
Export to PFM directly from Chunky, without needing to complete a render with the Denoise plugin.

PfmFileWriter class code is modified from PFM saving code in @leMaik's Denoise Plugin.

### DIFF
--- a/chunky/src/java/se/llbit/pfm/PfmFileWriter.java
+++ b/chunky/src/java/se/llbit/pfm/PfmFileWriter.java
@@ -24,8 +24,7 @@ public class PfmFileWriter implements AutoCloseable {
     this.out = new DataOutputStream(out);
   }
 
-  public PfmFileWriter(File file) throws IOException
-  {
+  public PfmFileWriter(File file) throws IOException {
     this(new FileOutputStream(file));
   }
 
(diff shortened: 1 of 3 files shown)


chunky/src/java/se/llbit/pfm/PfmFileWriter.java
```diff
@@ -0,0 +1,88 @@
...
+        out.write(ByteBuffer.allocate(4)
+            .order(byteOrder)
+            .putFloat((float)pixel[1])
+            .array());
+        out.write(ByteBuffer.allocate(4)
+            .order(byteOrder)
+            .putFloat((float)pixel[2])
+            .array());
+      }
+    }
```
aTom3333, line 86: This feels inefficient :/ (i don't know for sure without measuring), i'd like to experiment a bit with it

aTom3333:
Seems correct to me

chunky/src/java/se/llbit/pfm/PfmFileWriter.java
StanleyMines: It does feel a bit inefficient, but it is how leMaik's Denoiser plugin implemented it.

aTom3333:
So I had a quick look about the inefficient writing, it can easily be improved by writing to the file row by row.
I join the changes as a patch (i had to rename it to .txt because of github) if you want to apply it before your PR is merged (otherwise, i'd make a PR later)
[Write_to_pfm_row_by_row.txt](https://github.com/chunky-dev/chunky/files/5339978/Write_to_pfm_row_by_row.txt)
That's a simple change that makes the writing almost 10 times faster (from my single data point, a 1080*1920 render), not that it matters much in the grand scheme of things but that's always nice

leMaik:
@StanleyMines Thanks for the PR, looks good! :tada: 

@aTom3333 Since you are a contributor to this repository, you should be able to commit directly to @StanleyMines' branch. If the scene has post-processing disabled, the call to `postProcessRow` could be skipped entirely.

StanleyMines:
Suggestions from @aTom3333 and @leMaik implemented (not leMaik's 2nd one though. Is quite a bit faster than before though (far faster than TIFF)).

leMaik (approved):

leMaik (approved):

leMaik:
@StanleyMines Nice first PR! :rocket:

PR #704, opened 2020-10-16
Remove data nodes from the octree

leMaik:
In Chunky 1.x, data nodes were used to store block data. As of _the flattening_ in Chunky 2, new block palette entries are created for every block variant (= data configuration) so that data nodes are only used for water and lava to store the corner configuration (different water levels at the four corners; only the base level of the water block is stored in the chunk data, not the corner levels).

This RP remove data nodes from the octree and creates new block palette entries for every data/level configuration of water and lava. This reduces the size of the packed octree by 50% and should not have any negative performance impacts.

Also, this fixes the lava level if there isn't a solid block above lava (used to be full block height, but it's only 14/16th ingame).

### DIFF
--- a/chunky/src/java/se/llbit/math/Octree.java
+++ b/chunky/src/java/se/llbit/math/Octree.java
@@ -42,6 +42,7 @@ public class Octree {
 
   public interface OctreeImplementation {
     void set(int type, int x, int y, int z);
+    @Deprecated
     void set(Node data, int x, int y, int z);
     Node get(int x, int y, int z);
     Material getMaterial(int x, int y, int z, BlockPalette palette);
@@ -52,6 +53,7 @@ public class Octree {
     boolean isBranch(NodeId node);
     NodeId getChild(NodeId parent, int childNo);
     int getType(NodeId node);
+    @Deprecated
     int getData(NodeId node);
     default void startFinalization() {}
     default void endFinalization() {}
@@ -202,6 +204,7 @@ public class Octree {
       }
     }
 
+    @Deprecated
     public int getData() {
       return 0;
     }
@@ -217,7 +220,9 @@ public class Octree {
 
   /**
    * An octree node with extra data.
+   * @deprecated Not used anymore, only kept for compatibility with plugins
    */
+  @Deprecated
   static public final class DataNode extends Node {
     final int data;
 
@@ -482,7 +487,7 @@ public class Octree {
       Material prevBlock = ray.getCurrentMaterial();
 
       ray.setPrevMaterial(prevBlock, ray.getCurrentData());
-      ray.setCurrentMaterial(currentBlock, implementation.getData(node));
+      ray.setCurrentMaterial(currentBlock, 0);
 
       if (currentBlock.localIntersect) {
         if (currentBlock.intersect(ray, scene)) {
@@ -596,7 +601,7 @@ public class Octree {
       Material prevBlock = ray.getCurrentMaterial();
 
       ray.setPrevMaterial(prevBlock, ray.getCurrentData());
-      ray.setCurrentMaterial(currentBlock, implementation.getData(node));
+      ray.setCurrentMaterial(currentBlock, 0);
 
       if (!currentBlock.isWater()) {
         if (currentBlock.localIntersect) {
@@ -612,7 +617,7 @@ public class Octree {
         }
       }
 
-      if ((implementation.getData(node) & (1 << Water.FULL_BLOCK)) == 0) {
+      if (!(currentBlock instanceof Water && ((Water) currentBlock).isFullBlock())) {
         if (WaterModel.intersectTop(ray)) {
           ray.setCurrentMaterial(Air.INSTANCE, 0);
           return true;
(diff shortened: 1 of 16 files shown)


chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -482,7 +487,7 @@ public boolean enterBlock(Scene scene, Ray ray, BlockPalette palette) {
       Material prevBlock = ray.getCurrentMaterial();
 
       ray.setPrevMaterial(prevBlock, ray.getCurrentData());
-      ray.setCurrentMaterial(currentBlock, implementation.getData(node));
+      ray.setCurrentMaterial(currentBlock, 0);
```
aTom3333, line 490: setCurrentMaterial should probably be changed to only take one parameter

chunky/src/java/se/llbit/math/PackedOctree.java
```diff
@@ -235,8 +230,7 @@ private void mergeNode(int nodeIndex, int typeNegation, int data) {
   private boolean nodeEquals(int firstNodeIndex, int secondNodeIndex) {
     boolean firstIsBranch = treeData[firstNodeIndex] > 0;
     boolean secondIsBranch = treeData[secondNodeIndex] > 0;
-    return ((firstIsBranch && secondIsBranch) || treeData[firstNodeIndex] == treeData[secondNodeIndex]) // compare types
-            && treeData[firstNodeIndex + 1] == treeData[secondNodeIndex + 1]; // compare data
+    return ((firstIsBranch && secondIsBranch) || treeData[firstNodeIndex] == treeData[secondNodeIndex]); // compare types
     // FIXME possible bug here as we always compare the data even when dealing with nodes that don't really have data
     // The data int could potentially contain some junk leftover of a previous node
     // In theory it should be reset to 0 but we need to be careful
```
aTom3333, line 236: I left this comment while initially implementing the packed octree, turns out that was not a problem but now the comment is more irrelevant than ever

chunky/src/java/se/llbit/math/PackedOctree.java
```diff
@@ -222,7 +218,6 @@ private void mergeNode(int nodeIndex, int typeNegation, int data) {
     int childrenIndex = treeData[nodeIndex];
```
aTom3333, line 218: mergeNode should take the data anymore

chunky/src/java/se/llbit/math/PackedOctree.java
```diff
@@ -393,17 +382,11 @@ private void storeNode(DataOutputStream out, int nodeIndex) throws IOException {
...
       }
     } else {
-      boolean isDataNode = (treeData[nodeIndex + 1] != 0);
-      if(isDataNode) {
-        out.writeInt(type | DATA_FLAG);
-        out.writeInt(treeData[nodeIndex + 1]);
-      } else {
-        out.writeInt(type);
-      }
+      out.writeInt(type);
```
aTom3333, line 389: Minor style nitpicking, the `out.writeInt(type)` could be done out side of the `if`

chunky/src/java/se/llbit/math/Ray.java
```diff
@@ -386,7 +387,17 @@ public void setPrevMaterial(Material mat, int data) {
...
+      if (mat instanceof Water) {
+        this.currentData = ((Water) mat).data;
+      } else if (mat instanceof Lava) {
+        this.currentData = ((Lava) mat).data;
+      } else {
+        this.currentData = 0;
+      }
+    } else {
+      this.currentData = data;
+    }
```
aTom3333, line 400: Does the ray really need to store the data anymore?

aTom3333 (commented):
Nice job, can't wait to see memory usage will be impacted :D

chunky/src/java/se/llbit/math/Ray.java
leMaik: The ray's data is used in the water octree traversal code and removing it causes bugs there. This keeps it compatible with the rest of Chunky until we refactor the way water works (i.e. not require the ray to know the current and previous water data).

chunky/src/java/se/llbit/math/PackedOctree.java
leMaik: That would write it in both cases because we don't return in `if (isDataNode)`

aTom3333: Yeah, the same line is in both branches (and the if branch does additional stuff)

chunky/src/java/se/llbit/math/Ray.java
aTom3333: But can't the data be queried from the currentMaterial?

chunky/src/java/se/llbit/math/Octree.java
leMaik: I overloaded it now. It is still used to copy state between rays in `Scene#worldIntersection`.

chunky/src/java/se/llbit/math/Ray.java
leMaik: Maybe, but that's for another PR

leMaik: I don't know, it's a bit weird in the water intersection and world intersect code. With some refactoring, `data` can be dropped in the future, but that's probably for another PR.

aTom3333 (approved):

aTom3333 (approved):

PR #706, opened 2020-10-20
Improvements regarding emitter sampling

aTom3333:
Various improvements regarding emitter sampling that needed to be done:
 - Only create the emitter grid when using sampling strategy One or All (when switching to one of those 2, a warning appears to inform that the chunks need to be reloaded. A problem with this is that when changing sampling strategy before loading any chunks, the warning still appears, it would be better if it didn't)
 - Add some advanced options to control the size of the grid cell and, as suggested by Ryhida, add an option to prevent usual emitter contribution when using emitter sampling. 
 - Change the internal structure used to store the emitter grid to be (hopefully) more memory efficient. This only changes internals, the interface or the format on disk are unchanged so full backward compatibility.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
@@ -156,7 +156,7 @@ public class PathTracer implements RayTracer {
 
             Vector4 indirectEmitterColor = new Vector4(0, 0, 0, 0);
 
-            if (scene.emittersEnabled && currentMat.emittance > Ray.EPSILON) {
+            if (scene.emittersEnabled && (!scene.isPreventNormalEmitterWithSampling() || scene.getEmitterSamplingStrategy() == EmitterSamplingStrategy.NONE) && currentMat.emittance > Ray.EPSILON) {
 
               emittance = addEmitted;
               ray.emittance.x = ray.color.x * ray.color.x *
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
@@ -166,7 +166,7 @@ public class PathTracer implements RayTracer {
               ray.emittance.z = ray.color.z * ray.color.z *
                   currentMat.emittance * scene.emitterIntensity;
               hit = true;
-            } else if(scene.emittersEnabled && scene.emitterSamplingStrategy != EmitterSamplingStrategy.NONE) {
+            } else if(scene.emittersEnabled && scene.emitterSamplingStrategy != EmitterSamplingStrategy.NONE && scene.getEmitterGrid() != null) {
               // Sample emitter
               boolean sampleOne = scene.emitterSamplingStrategy == EmitterSamplingStrategy.ONE;
               if(sampleOne) {
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -298,6 +298,10 @@ public class Scene implements JsonSerializable, Refreshable {
 
   private Grid emitterGrid;
 
+  private int gridSize = PersistentSettings.getGridSizeDefault();
+
+  private boolean preventNormalEmitterWithSampling = PersistentSettings.getPreventNormalEmitterWithSampling();
+
   /**
    * The octree implementation to use
    */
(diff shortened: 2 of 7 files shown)


leMaik:
> A problem with this is that when changing sampling strategy before loading any chunks, the warning still appears, it would be better if it didn't)

I can look into this. :+1:

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
```diff
@@ -156,7 +156,7 @@ public static boolean pathTrace(Scene scene, Ray ray, WorkerState state, int add
 
             Vector4 indirectEmitterColor = new Vector4(0, 0, 0, 0);
 
-            if (scene.emittersEnabled && currentMat.emittance > Ray.EPSILON) {
+            if (scene.emittersEnabled && (!scene.isPreventNormalEmitterWithSampling() || scene.getEmitterSamplingStrategy() == EmitterSamplingStrategy.NONE) && currentMat.emittance > Ray.EPSILON) {
```
leMaik, line 159: should this be `&& !(scene.isPreventNormalEmitterWithSampling() || scene.getEmitterSamplingStrategy() == EmitterSamplingStrategy.NONE)`?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -509,9 +513,11 @@ public synchronized void loadScene(RenderContext context, String sceneName,
     }
 
     // Try loading the grid unconditionally for now
-    boolean emiiterGridLoaded = loadEmitterGrid(context, taskTracker);
+    boolean emitterGridLoaded = true;
```
leMaik, line 516: Why does this default to `true` even if no emitter grid is loaded after this line?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1018,7 +1025,7 @@ public synchronized void loadChunks(TaskTracker progress, World world,
                 }
                 worldOctree.set(octNode, x, cy - origin.y, z);
 
-                if (block.emittance > 1e-4) {
+                if (emitterGrid != null & block.emittance > 1e-4) {
```
leMaik, line 1028: `&&`

chunky/src/java/se/llbit/chunky/ui/render/LightingTab.java
```diff
@@ -109,6 +110,8 @@ public LightingTab() throws IOException {
     emitterSamplingStrategy.getSelectionModel().selectedItemProperty()
             .addListener((observable, oldvalue, newvalue) -> {
               scene.setEmitterSamplingStrategy(newvalue);
+              if(newvalue != EmitterSamplingStrategy.NONE && scene.getEmitterGrid() == null)
+                Log.warn("The world need to be reloaded for emitter sampling to work");
```
leMaik, line 114: *needs (and maybe a `// TODO add nice UI dialog for this` :sweat_smile:)

chunky/src/res/se/llbit/chunky/ui/render/AdvancedTab.fxml
```diff
@@ -36,6 +36,8 @@
             <ChoiceBox fx:id="octreeImplementation" prefWidth="150.0" />
           </children>
         </HBox>
+        <IntegerAdjuster fx:id="gridSize" />
+        <CheckBox fx:id="preventNormalEmitterWithSampling" mnemonicParsing="false" text="Prevent normal emitter when using emitter sampling" />
```
leMaik, line 40: What exactly does this option do?

leMaik (changes requested):
Thank you for fixing this and even improving the grid. :tada:

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
aTom3333: No, I think this is right. The condition for doing normal emitter (aka not sampled) should be true when we do not prevent ir or when there is no strategy

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
aTom3333: To not reload the world if it is not needed (well, it is true the name doesn't really convey the right meaning, i'll rename this variable)

chunky/src/res/se/llbit/chunky/ui/render/AdvancedTab.fxml
aTom3333: It disables the usual contribution of emitters when emitter sampling is used as they produce noise when emitter sampling tries to not require a lot of SPP

leMaik (approved):

PR #708, opened 2020-10-23
Fix the growth of packed octree

aTom3333:
Didn't see that during the review of the changes to packed octree but when adding new nodes, we only need to grow by 8 ints now. Without this small change, we were using as much memory as previously but wasting half of it


leMaik:
Whoops… I only searched for `* 2` when updating the code. This should explain why memory usage didn't decrease @jackjt8 :sweat_smile:

jackjt8:
Yup. Will perform the tests tomorrow with the retained memory from VisualVM as that seems to be more revealing in these cases.

PR #711, opened 2020-10-28
Fix some 1.16 chunks being loaded incorrectly.

leMaik:
Fixes bugs like this:

image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/world/Chunk.java
+++ b/chunky/src/java/se/llbit/chunky/world/Chunk.java
@@ -295,7 +295,7 @@ public class Chunk {
           if (blockStates.isLongArray(dataSize)) {
             // since 20w17a, block states are aligned to 64-bit boundaries, so there are 64 % bpb
             // unused bits per block state; if so, the array is longer than the expected data size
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/chunky/world/Chunk.java
```diff
@@ -292,7 +295,12 @@ private static void loadBlockData(@NotNull Map<String, Tag> data, @NotNull int[]
           if (blockStates.isLongArray(dataSize)) {
             // since 20w17a, block states are aligned to 64-bit boundaries, so there are 64 % bpb
             // unused bits per block state; if so, the array is longer than the expected data size
-            boolean isAligned = blockStates.longArray().length > dataSize;
+            boolean isAligned = data.get(DATAVERSION).intValue() >= DATAVERSION_20w17a;
+            if (isAligned) {
+              // entries are 64-bit-padded, re-calculate the bits per block
+              // this is the dataSize calculation from above reverted, we know the actual data size
+              bpb = blockStates.longArray().length / 64;
```
leMaik, line 302: There are chunks where this re-calculation changes the bpb value. Previously, those chunks were loaded incorrectly (i.e. with a too small bpb value).

PR #714, opened 2020-11-01
Don't store the linear texture and gamma correct on the fly

aTom3333:
New memory optimization we talked about in discord.
Remove the linear texture stored as a float array and only keep the BitmapImage (where every pixel, 4 components, is stored as an int).
Gamma correction is then done on the fly by using a look up table.
I made some tests to measure the performance impact this has, The difference in time I measured is in the order of 4% in favor of the gamma correction on the fly! But my RAM was pretty much saturated when I did those measurement so they may not be representative of usages where more memory is available.
In term of memory reduction, the size of textures has been divided by 5 (from one int and 4 float per pixel to only one int), for 16*16 textures this means a reduction of about 4MB

### DIFF
--- a/chunky/src/java/se/llbit/chunky/resources/Texture.java
+++ b/chunky/src/java/se/llbit/chunky/resources/Texture.java
@@ -931,7 +931,8 @@ public class Texture {
   protected int height;
   protected int avgColor;
   private float[] avgColorLinear;
-  private float[] linear;
+  private boolean useAverageColor = false;
+  private float[] avgColorFlat;
 
   private Image fxImage = null;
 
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/chunky/resources/Texture.java
```diff
@@ -931,7 +931,8 @@
   protected int height;
   protected int avgColor;
   private float[] avgColorLinear;
-  private float[] linear;
+  private boolean useAverageColor = false;
```
leMaik, line 934: Why is this stored per texture?

leMaik (approved):

chunky/src/java/se/llbit/chunky/resources/Texture.java
aTom3333: I didn't want to change the behavior, the value value is stored depending on the value value of the option at the moment `setTexture` is called (previously by storing something different into linear, now by remembering the flag) but it is true I can make it static (in which case i need to initialize `avgColorFlat` everytime but this isn't a problem)

chunky/src/java/se/llbit/math/ColorUtil.java
```diff
@@ -28,6 +28,15 @@
  */
 public final class ColorUtil {
 
+  // Look up table used to speed up gamma correction
+  public static final float[] toLinearLut = new float[256];
```
leMaik, line 32: This should be private, it's pretty much an internal optimisation

jackjt8:
Performance:

_~12.5/24 GiB, 64 Regions, 4 Render threads, Default 16x resource pack, single run._

* 392-g93097632 - 1614 seconds; 328,764 SPS

* 393-g5d75a997 - 1631 seconds; 325,355 SPS

---

Multiple runs & memory usage with HD resource packs will follow tomorrow.

jackjt8:
x512 Resource pack

392-g93097632 - 17465.3 MiB ; 1546 seconds

393-g5d75a997 - 11744.5 MiB ; 1640 seconds

Memory usage is reduced by about 30% by this change bringing it in line with x16 resource packs however performance is reduced by upto 6% (based upon my latest numbers).

PR #715, opened 2020-11-05
Fix wrong render time after loading a scene that was rendered before …

leMaik:
…but has no dump.


jackjt8:
[Headless issue]

leMaik:
@jackjt8 I have now added a label for headless mode.

PR #717, opened 2020-11-08
Check position is in bound of the grid

aTom3333:
When using emitter sampling, check that the position at which we want to retrieve emitter is in the grid. Hopefully should fix the bug @jackjt8 encountered

### DIFF
--- a/chunky/src/java/se/llbit/math/Grid.java
+++ b/chunky/src/java/se/llbit/math/Grid.java
@@ -129,6 +129,10 @@ public class Grid {
     int gridX = x / cellSize;
     int gridY = y / cellSize;
     int gridZ = z / cellSize;
+
+    if(isOOB(gridX) || isOOB(gridY) || isOOB(gridZ))
+      return null;
+
     int index = cellIndex(gridX, gridY, gridZ);
 
     int start = constructedGrid[2*index];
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/math/Grid.java
```diff
@@ -141,17 +145,26 @@ public EmitterPosition sampleEmitterPosition(int x, int y, int z, Random random)
     return emitterPositions.get(emitterIndex);
   }
 
+  private boolean isOOB(int coord)
```
leMaik, line 148: Maybe rename this to `isOutOfBounds`? And maybe `isOutOfBounds(int x, int y, int z)` just in case the grid gets non-cubic in the future?

leMaik (approved):

leMaik:
@jackjt8 Can you verify that this does fix the bug?

jackjt8:
Success! Without this fix we would otherwise instantly have multiple ArrayIndexOutOfBoundsExceptions from each worker thread. I am currently at 4 SPP and counting which does indicate the fix worked. I'll keep the scene rendering but I believe this has resolve the issue.

image

jackjt8 (approved):

chunky/src/java/se/llbit/math/Grid.java
aTom3333: Making the grid non cubic is indeed the next step (but I was too lazy to do that today) so this change makes sense

PR #720, opened 2020-11-14
Change emitter grid to not store useless cells

aTom3333:
Instead of having an emitter grid that spans the whole octree, the emitter grid is changed to only span as much as is necessary for the emitter in the scene. This should further decrease memory usage of the grid.

### DIFF
--- a/chunky/src/java/se/llbit/math/Grid.java
+++ b/chunky/src/java/se/llbit/math/Grid.java
@@ -53,14 +53,14 @@ public class Grid {
     this.cellSize = cellSize;
     long totalSize = (1L << octreeDepth);
     this.gridSize = (int) ((totalSize + (cellSize-1)) / cellSize);
-    minX = maxX = minY = maxY = minZ = maxZ = 1;
+    minX = maxX = minY = maxY = minZ = maxZ = -1;
   }
 
   // Constructor used by the load method
   private Grid(int gridSize, int cellSize, int overloadSelectFlag /*unused*/) {
     this.gridSize = gridSize;
     this.cellSize = cellSize;
-    minX = maxX = minY = maxY = minZ = maxZ = 1;
+    minX = maxX = minY = maxY = minZ = maxZ = -1;
   }
 
   public void addEmitter(int x, int y, int z) {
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/math/Grid.java
```diff
@@ -8,7 +8,7 @@
 import java.util.Random;
 
 public class Grid {
-  private static final int GRID_FORMAT_VERSION = 0;
+  private static final int GRID_FORMAT_VERSION = 1;
```
leMaik, line 11: Is the new format version still compatible with version 0 so that version 0 files can be converted to version 1?

chunky/src/java/se/llbit/math/Grid.java
```diff
@@ -47,32 +46,56 @@ public EmitterPosition(int x, int y, int z) {
...
+
+  private boolean isOutOfBounds(int x, int y, int z)
+  {
+    return x < offsetX || x >= sizeX + offsetX
+            || y < offsetY || y >= sizeY + offsetY
+            || z < offsetZ || z >= sizeZ + offsetZ;
   }
 
   /**
    * Needs to be call when every emitter has been added
```
leMaik, line 87: :thinking: Maybe we should add a `GridBuilder` to enforce this? That would also allow to drop minXYZ and maxXYZ from the grid itself. But that's just minor refactoring, nothing to worry about in this PR.

chunky/src/java/se/llbit/math/Grid.java
```diff
@@ -213,9 +234,14 @@ public static Grid load(DataInputStream in) throws IOException {
       throw new RuntimeException("Unknown grid format version, can't load the grid");
```
leMaik, line 234: I guess it shouldn't be too hard to make this backward-compatible? If version is 1, then offsetXYZ=0 and sizeXYZ=readInt(), right?

leMaik (changes requested):

chunky/src/java/se/llbit/math/Grid.java
aTom3333: Yes, they could be loaded by setting the offsets to 0 and the sizes to the old gridSize. Didn't think of that and thought reloading the chunks would be enough. Should I do it?

leMaik: I always try to keep the files compatible so that the original chunks aren't required anymore once they have been loaded so that new things "just work" with new Chunky versions.
If you don't want to, that's fine too. It's something that could be added later on.

PR #729, opened 2020-11-21
Emitter sampling reload chunks

leMaik:
The UX when enabling emitter sampling was slightly confusing (reload _world_ instead of _chunks_) so here is the proper dialog from the TODO. If the user doesn't reload the chunks, it disables emitter sampling.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/render/LightingTab.java
+++ b/chunky/src/java/se/llbit/chunky/ui/render/LightingTab.java
@@ -21,6 +21,10 @@ import javafx.fxml.FXML;
 import javafx.fxml.FXMLLoader;
 import javafx.fxml.Initializable;
 import javafx.scene.Node;
+import javafx.scene.control.Alert;
+import javafx.scene.control.Alert.AlertType;
+import javafx.scene.control.ButtonBar.ButtonData;
+import javafx.scene.control.ButtonType;
 import javafx.scene.control.CheckBox;
 import javafx.scene.control.ChoiceBox;
 import javafx.scene.control.ScrollPane;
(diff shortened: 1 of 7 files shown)


chunky/src/java/se/llbit/chunky/ui/render/LightingTab.java
```diff
@@ -108,16 +113,30 @@ public LightingTab() throws IOException {
...
+            warning.getButtonTypes().setAll(
+              ButtonType.CANCEL,
+              new ButtonType("Reload chunks", ButtonData.FINISH));
+            warning.setTitle("Chunk reload required");
+            ButtonType result = warning.showAndWait().orElse(ButtonType.CANCEL);
+            if (result.getButtonData() == ButtonData.FINISH) {
+              controller.getRenderController().getSceneManager().reloadChunks();
+            } else {
+              scene.setEmitterSamplingStrategy(EmitterSamplingStrategy.NONE);
+              emitterSamplingStrategy.getSelectionModel().select(EmitterSamplingStrategy.NONE);
```
aTom3333, line 130: Maybe make it clear in the text of the button that emitter sampling will be disabled or don't disable emitter sampling. Maybe the user wants to do several changes that would need the chunks to be reloaded (like y min/max) but doesn't want to reload for every change

aTom3333 (approved):

chunky/src/java/se/llbit/chunky/ui/render/LightingTab.java
leMaik: Good point! Didn't think of that

Issue #315, opened 2015-04-19
Loading scene should switch the current world


llbit:
When loading a scene from a different world it would be nice if Chunky loaded that world for the 2D map overview.

PR #730, opened 2020-11-21
Don't clear the selected chunks when reloading the world.

leMaik:


leMaik:
Every few weeks I wonder why this isn't merged yet. For future me: I wanted to also make it switch the world directory after loading a scene that uses a different world, as proposed in #315

PR #756, opened 2020-12-21
Change compression method for render dump

aTom3333:
Implement the FPC algorithm to compress the render dumps instead of using gzip.
Compression ratio is similar but with higher reading/writing speed (around 10x).
The render dump format is changed. Backward compatibility is kept by trying to read a magic number (to differentiate from an old dump) and if it doesn't match, puting back the bytes so they are present for the gzip stream

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -16,7 +16,8 @@
  */
 package se.llbit.chunky.renderer.scene;
 
-import java.io.FileOutputStream;
+import java.io.*;
+
 import org.apache.commons.math3.util.FastMath;
 import se.llbit.chunky.PersistentSettings;
 import se.llbit.chunky.block.Air;
@@ -33,6 +34,7 @@ import se.llbit.chunky.entity.Poseable;
 import se.llbit.chunky.renderer.*;
 import se.llbit.chunky.renderer.projection.ProjectionMode;
 import se.llbit.chunky.resources.BitmapImage;
+import se.llbit.chunky.resources.FloatingPointCompressor;
 import se.llbit.chunky.resources.OctreeFileFormat;
 import se.llbit.chunky.world.Biomes;
 import se.llbit.chunky.world.Chunk;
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -89,6 +82,8 @@
  */
 public class Scene implements JsonSerializable, Refreshable {
 
+  private static final byte[] DUMP_FORMAT_MAGIC_NUMBER = {0x44, 0x55, 0x4D, 0x50};
```
leMaik, line 85: `DUMP` :smile:

chunky/src/java/se/llbit/chunky/resources/FloatingPointCompressor.java
```diff
@@ -0,0 +1,215 @@
...
+      current = bits;
+
+      return Double.longBitsToDouble(bits);
+    }
+  }
+
+  public static void compress(double[] input, OutputStream output) throws IOException {
+    try(BufferedOutputStream out = new BufferedOutputStream(output)) {
+      if(input.length % 3 != 0)
+        throw new RuntimeException("Dump doesn't have a multiple of 3 values");
```
leMaik, line 138: This should probably be an `IllegalArgumentException`

chunky/src/java/se/llbit/chunky/resources/FloatingPointCompressor.java
```diff
@@ -0,0 +1,215 @@
...
+        gEncoder.encodeSingleWithOddTerminator(input[idx+1], out);
+        bEncoder.encodeSingleWithOddTerminator(input[idx+2], out);
+      }
+    }
+  }
+
+  public static void decompress(double[] output, InputStream input) throws IOException {
+    try(BufferedInputStream in = new BufferedInputStream(input)) {
+      if(output.length % 3 != 0)
+        throw new RuntimeException("Dump doesn't have a multiple of 3 values");
```
leMaik, line 164: See above

chunky/src/java/se/llbit/chunky/resources/FloatingPointCompressor.java
```diff
@@ -0,0 +1,215 @@
...
+      if(pixels % 2 == 1) {
+        int idx = 3*size;
+        rEncoder.encodeSingleWithOddTerminator(input[idx], out);
+        gEncoder.encodeSingleWithOddTerminator(input[idx+1], out);
+        bEncoder.encodeSingleWithOddTerminator(input[idx+2], out);
+      }
+    }
+  }
+
+  public static void decompress(double[] output, InputStream input) throws IOException {
```
leMaik, line 161: (nit) For the `compress` method, it's `input, output`. Should the arguments be switched here?

leMaik (approved):

chunky/src/java/se/llbit/chunky/resources/FloatingPointCompressor.java
aTom3333: Yeah, I let like that to get the types in the same order but I don't really care

PR #767, opened 2020-12-31
Sliders to zero

### LINKED ISSUE #733: Sliders to 0
Any setting sliders will not allow you to go to zero, but the textbox does allow it to.

blueplanet150:
Fixes #733 

The issue says that there are setting sliders, which dont go down to zero, but the textbox allows it. 
I noticed there are 2 casses where this happens:
1. When DoubleAdjuster.setRange(min, max, slider_min) is called to set the range. I suppose this is for the logarithmic scaling, slider_min preventing values too close to zero. I fixed this problem, but it was only in one case (fog density). Now, setRange(min, max) will automatically set the minimum for logarithmic scaling (logScaleStart) to  0.01. For slider positions < logScaleStart, the value is set to 0. Also, for values < logScaleStart, the slider position is set to zero.
2. When DoubleAdjuster(setRange(min, max) is called to set the range, with min > 0. In this case,the slider will go down to min, but the textbox can still be set to ANY value, including zero. I did not fix this. Its a different Problem.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
+++ b/chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
@@ -89,7 +89,7 @@ public class SkyTab extends ScrollPane implements RenderControlsTab, Initializab
     horizonOffset.onValueChange(value -> scene.sky().setHorizonOffset(value));
 
     cloudSize.setName("Cloud size");
-    cloudSize.setRange(0.1, 128);
+    cloudSize.setRange(0, 128);
     cloudSize.clampMin();
     cloudSize.makeLogarithmic();
     cloudSize.onValueChange(value -> scene.sky().setCloudSize(value));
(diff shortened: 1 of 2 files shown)


leMaik:
Nice first PR! :partying_face: 

So can the slider now move to 0? If so, I'd say the bug is fixed.

blueplanet150:
Thank you :)
Yes it can move to zero.

jackjt8:
Thank you for resolving that. I did notice it when I enabled logarithmic scaling however I couldn't quite figure it out.

jackjt8 (approved):
While I can't comment on how the code is laid out or anything like that; It seems to look good.

Important aspect is that not only does the slider function between 0.0, 0.01 -> 1.0 but text based inputs are still respected for values between 0.0 & 0.01. And values are correctly written to the JSON.

leMaik (approved):

leMaik:
Thanks again, @blueplanet150. I just removed the commented console logger import.

PR #771, opened 2021-01-05
fixes & optimisations

NotStirred:
These are just random things I see and optimise as I go through my other PR

There will be a lot of commits as I'm splitting things up. Should probably all be squashed before merge

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/SynchronousSceneManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/SynchronousSceneManager.java
@@ -222,7 +222,9 @@ public class SynchronousSceneManager implements SceneProvider, SceneManager {
   @Override public boolean pollSceneStateChange() {
     if (scene.shouldRefresh() && (scene.getForceReset() || resetHandler.allowSceneRefresh())) {
       return true;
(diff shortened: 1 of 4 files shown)


NotStirred:
Github had a massive panic when I rebased onto master, sorry about that

chunky/src/java/se/llbit/chunky/renderer/scene/SynchronousSceneManager.java
```diff
@@ -222,10 +222,7 @@ public void setOnChunksLoaded(Runnable onChunksLoaded) {
   @Override public boolean pollSceneStateChange() {
     if (scene.shouldRefresh() && (scene.getForceReset() || resetHandler.allowSceneRefresh())) {
       return true;
-    } else if (scene.getMode() != storedScene.getMode()) {
-      return true;
-    }
-    return false;
+    } else return scene.getMode() != storedScene.getMode();
```
leMaik, line 225:
```suggestion
    } else {
      return scene.getMode() != storedScene.getMode();
    }
```

leMaik (approved):
I don't like the single-line `else`, but I can also just fix that myself :+1:

PR #776, opened 2021-01-06
Add rough specular reflections

leMaik:
This adds a new _roughness_ property to `Material`. In the _Materials_ tab, it can be configured with the _Smoothness_ slider, which controls the _perceptual smoothness_, which is then converted to the roughness. Perceptual smoothness feels more linear when working with it and it's also what LabPBR 1.3 uses for roughness maps.

image


jackjt8:
Had a look at the provided reference, and this other one I found where [section 25.2 has Glossy Reflection implementation](http://www.raytracegroundup.com/downloads/Chapter25.pdf), this implementation seems to be correct.

leMaik:
Thanks for looking into this, @jackjt8! Let's see what snapshot users discover…

Specular color isn't implemented yet btw. but it's on my list.

PR #778, opened 2021-01-07
Beacon beam

073TheBoss:
This PR adds the beam to beacons. It doesn't yet detect stained glass or solid blocks above the beacon, but I will continue to work on that for a future PR.

Ideally, I wouldn't have changed the Block class, but it was the best solution I could come up with to keep the entities list from being polluted with beams that weren't actually there.

The beams will show up in the Entities pose tab, where you can change the height, scaling, and y-rotation. I've disabled the x and z rotation because it would cause the beam to split up into segments, and the beam doesn't ever rotate that way in game.

As always, let me know if I need to change anything.

default_2021-01-06_19-45-30-1500

### DIFF
--- a/chunky/src/java/se/llbit/chunky/block/Block.java
+++ b/chunky/src/java/se/llbit/chunky/block/Block.java
@@ -58,7 +58,7 @@ public abstract class Block extends Material {
     return name;
   }
 
-  public boolean isBlockEntity() {
+  public boolean isBlockEntity(CompoundTag entityTag) {
     return false;
   }
 
--- a/chunky/src/java/se/llbit/chunky/block/Block.java
+++ b/chunky/src/java/se/llbit/chunky/block/Block.java
@@ -58,7 +58,7 @@ public abstract class Block extends Material {
     return name;
   }
 
-  public boolean isBlockEntity(CompoundTag entityTag) {
+  public boolean isBlockEntity() {
     return false;
   }
 
--- a/chunky/src/java/se/llbit/chunky/block/Head.java
+++ b/chunky/src/java/se/llbit/chunky/block/Head.java
@@ -55,7 +55,7 @@ public class Head extends MinecraftBlockTranslucent {
   }
 
   @Override
-  public boolean isBlockEntity() {
+  public boolean isBlockEntity(CompoundTag entityTag) {
     return type == Kind.PLAYER;
   }
 
--- a/chunky/src/java/se/llbit/chunky/block/Head.java
+++ b/chunky/src/java/se/llbit/chunky/block/Head.java
@@ -55,15 +55,19 @@ public class Head extends MinecraftBlockTranslucent {
   }
 
   @Override
-  public boolean isBlockEntity(CompoundTag entityTag) {
-    return type == Kind.PLAYER;
+  public boolean isBlockEntity() {
+    return true;
   }
 
   @Override
   public Entity toBlockEntity(Vector3 position, CompoundTag entityTag) {
-    String textureUrl = getTextureUrl(entityTag);
-    return textureUrl != null ? new HeadEntity(position, textureUrl, rotation, 1)
-        : new SkullEntity(position, type, rotation, 1);
+    if (type == Kind.PLAYER) {
+      String textureUrl = getTextureUrl(entityTag);
+      return textureUrl != null ? new HeadEntity(position, textureUrl, rotation, 1)
+          : new SkullEntity(position, type, rotation, 1);
+    } else {
+      return null;
+    }
   }
 
   public static String getTextureUrl(CompoundTag entityTag) {
--- a/chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
+++ b/chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
@@ -1,0 +1,138 @@
+package se.llbit.chunky.entity;
+
+import se.llbit.chunky.resources.Texture;
+import se.llbit.chunky.world.Material;
+import se.llbit.chunky.world.material.TextureMaterial;
+import se.llbit.json.Json;
+import se.llbit.json.JsonObject;
+import se.llbit.json.JsonValue;
+import se.llbit.math.*;
+import se.llbit.math.primitive.Primitive;
+import se.llbit.nbt.CompoundTag;
+import se.llbit.util.JsonUtil;
+
+import java.util.Collection;
+import java.util.LinkedList;
+
+public class BeaconBeam extends Entity implements Poseable {
+
+  public static final Material beaconBeamMaterial = new TextureMaterial(Texture.beaconBeam);
+
+  private static final Quad[] beam = new Quad[]{
+      new Quad(
+          new Vector3(5 / 16.0, 0.0, 5 / 16.0),
+          new Vector3(5 / 16.0, 0.0, 11 / 16.0),
+          new Vector3(5 / 16.0, 1.0, 5 / 16.0),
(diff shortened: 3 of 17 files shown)


chunky/src/java/se/llbit/chunky/block/Block.java
```diff
@@ -58,7 +58,7 @@ public String toString() {
     return name;
   }
 
-  public boolean isBlockEntity() {
+  public boolean isBlockEntity(CompoundTag entityTag) {
```
leMaik, line 61: Hm… :thinking:
Instead of changing this signature, we could also treat this as "maybe a block entity" and allow `getBlockEntity` to return `null` if there is none.

Another (more involved) way would be to remove  `isBlockEntity` completely and move the logic into a `BlockEntityBlock` instead and check via `block instanceof BlockEntityBlock`.

I'm not sure what to do here. What do you think?

chunky/src/java/se/llbit/chunky/block/Head.java
```diff
@@ -55,7 +55,7 @@ public Entity toEntity(Vector3 position) {
   }
 
   @Override
-  public boolean isBlockEntity() {
+  public boolean isBlockEntity(CompoundTag entityTag) {
     return type == Kind.PLAYER;
```
leMaik, line 59: Re previous comment: This logic could be moved to `getBlockEntity` and return null if the type isn't a player. Same for `WallHead`

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
```diff
@@ -0,0 +1,136 @@
...
+  };
+
+  private final JsonObject pose;
+  private double scale = 1;
+  private int height = 256;
+
+  public BeaconBeam(Vector3 position) {
+    super(position);
+    this.pose = new JsonObject();
+    pose.add("all", JsonUtil.vec3ToJson(new Vector3(0, 0, 0)));
```
leMaik, line 55: What I like about the `height` attribute is that you can make the beams _not_ stop at a certain height. But maybe you could calculate the default height here with 255 - position.y (please check that formula for off-by-one errors :smile:) so that all beacons are as high as ingame by default?

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
```diff
@@ -0,0 +1,136 @@
...
+        .rotateY(allPose.y) // Ignore x and z rotation because it breaks the beam into segments
+        .translate(0.5, 0.5, 0.5)
+        .translate(position.x + offset.x, position.y + offset.y, position.z + offset.z));
+  }
+
+  public Collection<Primitive> primitives(Transform transform) {
+    Collection<Primitive> faces = new LinkedList<>();
+    //Have 1 block tall model and repeat it for height * scale.
+    //This addresses the texture stretching problem and
+    //allows for the height to be changed.
```
leMaik, line 80: Also, this will allow tinting individual segments… which is exactly what you'll need to add colored beacons. :+1: 
(Maybe using a special `TintedBeaconMaterial`… I think we already have a material for overlays which is used for leather armor.

chunky/src/java/se/llbit/chunky/ui/render/EntitiesTab.java
```diff
@@ -254,6 +256,19 @@ private void updateEntity(Entity entity) {
           controls.getChildren().add(page2Angle);
         }
       }
+      else if(entity instanceof BeaconBeam) {
+        BeaconBeam beam = (BeaconBeam) entity;
+        IntegerAdjuster height = new IntegerAdjuster();
+        height.setName("Height");
+        height.setTooltip("Modifies the height of the beam. Useful if your scene is taller than the world height.");
+        height.set(256);
```
leMaik, line 264: This should probably be `height.set(beam.getHeight())`

leMaik (changes requested):

chunky/src/java/se/llbit/chunky/block/Block.java
073TheBoss: I think the `BlockEntityBlock` solution would be a more complete solution, but yeah, it is much more involved and probably doesn't have much benefit right now as the Beacon would be the only block using that functionality. Allowing for `getBlockEntity` to return null would be much easier.

leMaik: :+1: Then go for that variant.

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
073TheBoss: I can change that. Beacons in game do go up infinitely, though.

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
```diff
@@ -0,0 +1,136 @@
...
+    return primitives(Transform.NONE
+        .translate(-0.5, -0.5, -0.5)
+        .scale(scale)
+        .rotateY(allPose.y) // Ignore x and z rotation because it breaks the beam into segments
+        .translate(0.5, 0.5, 0.5)
+        .translate(position.x + offset.x, position.y + offset.y, position.z + offset.z));
+  }
+
+  public Collection<Primitive> primitives(Transform transform) {
+    Collection<Primitive> faces = new LinkedList<>();
```
aTom3333, line 77: Just change that for an ArrayList? It probably won't make a noticeable difference but in most cases ArrayList are faster and they are always more memory efficient (not that the collection itself will be in memory tho)

leMaik: If Java had generators, we could just `yield` the faces as they are created without creating that temporary list at all. :disappointed:

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
leMaik: No need to change this then

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  public Collection<Primitive> primitives(Transform transform) {
+    ArrayList<Primitive> faces = new ArrayList<>();
+    //Have 1 block tall model and repeat it for height * scale.
+    //This addresses the texture stretching problem and
+    //allows for the height to be changed.
+    for (int i = 0; i < height; i++) {
+      for (Quad quad : beam) {
+        quad.addTriangles(faces, beaconBeamMaterial, transform.translate(0.0, 1.0 * i * scale, 0.0));
```
jackjt8, line 84:
```
                -Math.cos(allPose.z - Math.PI / 2) * i * scale,
                (Math.sin(allPose.z - Math.PI / 2) * Math.sin(allPose.x - Math.PI / 2)) * i * scale,
                Math.cos(allPose.x - Math.PI / 2) * i * scale
```

I'm so fucking done with this janky rotation... For the most part the code, which Boss provided to me, was correct however there was the issue of a missing allPose.z aspect to calculating the Y offset. After adding this component the rotation along two singular axis is correct. The main issue arises from when we try and rotate along two axis where the segments break apart due to incorrect segment rotation. Additionally the order we enter rotational components into the UI can impact how the (segment) rotation is applied and in some cases segment rotation in an axis gets locked out. If someone could look at fixing the segment rotation order that might just let us rotate beams freely.

Pitch 90
image

Roll 45
image


Roll 90
image

Pitch & Roll @ 90 
image

leMaik (approved):

PR #779, opened 2021-01-09
Precompute coefficient for intersection computation

aTom3333:
I tried to optimize a bit octree traversal, unfortunately some of the things I tried didn't give good result but here are some changes that seem to have a positive impact.
Basically the idea is to optimize the the computation for intersection ray/axis-aligned plane (and a computation of intersection ray-axis-aligned-box is made of 6 such computations). This is done by caching the coefficient of the computation such as the computation consist only in a multiplication and an addition. (division being substantially more expensive than multiplication or addition).
In addition, to avoid having to update those coefficient after each box test, I don't update the origin of the ray until it is needed (before returning from the function or before inking another function that expects the origin of the ray to be updated).
This means that during the intersection test, the origin of the ray isn't always inside the block so the 6 faces must be tested unconditionally (Note that this approach wouldn't work in general, the ray could not intersect the box and the code would say that it does as it tests infinite planes, here this is not an issue because x, y, z are still updated every iteration so the box we test against is one we know the ray intersects, we just need to know where).
In term of performance improvement, it seems to be about 4-7% reduction in render time

### DIFF
--- a/chunky/src/java/se/llbit/math/Octree.java
+++ b/chunky/src/java/se/llbit/math/Octree.java
@@ -457,15 +457,28 @@ public class Octree {
 
     int depth = implementation.getDepth();
 
+    double distance = Ray.EPSILON;
+
+    int x = (int) QuickMath.floor(ray.o.x + ray.d.x * Ray.OFFSET);
+    int y = (int) QuickMath.floor(ray.o.y + ray.d.y * Ray.OFFSET);
+    int z = (int) QuickMath.floor(ray.o.z + ray.d.z * Ray.OFFSET);
+
+    // floating point division are slower than multiplication so we cache them
+    // We also try to limit the number of time the ray origin is updated
+    // as it would require to recompute those values
+    double invDx = 1 / ray.d.x;
+    double invDy = 1 / ray.d.y;
+    double invDz = 1 / ray.d.z;
+    double offsetX = -ray.o.x * invDx;
+    double offsetY = -ray.o.y * invDy;
+    double offsetZ = -ray.o.z * invDz;
+
     // Marching is done in a top-down fashion: at each step, the octree is descended from the root to find the leaf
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -490,19 +502,33 @@ public boolean enterBlock(Scene scene, Ray ray, BlockPalette palette) {
       ray.setCurrentMaterial(currentBlock);
 
       if (currentBlock.localIntersect) {
+        // Other functions expect the ray origin to be in the block they test so here time
+        // to update it
+        ray.o.scaleAdd(distance, ray.d);
```
leMaik, line 507: Should we add a comment on the fact that `offsetX,Y,Z` need to be updated but this is deferred to lines 515 and 523?
It's clear when you know what the code _should_ do, though.

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -490,19 +502,33 @@ public boolean enterBlock(Scene scene, Ray ray, BlockPalette palette) {
...
           ray.exitBlock(x, y, z);
+          offsetX = -ray.o.x * invDx;
+          offsetY = -ray.o.y * invDy;
+          offsetZ = -ray.o.z * invDz;
           continue;
         }
       } else if (!currentBlock.isSameMaterial(prevBlock) && currentBlock != Air.INSTANCE) {
+        // Origin and distance of ray need to be updated
+        ray.o.scaleAdd(distance, ray.d);
+        ray.distance += distance;
```
leMaik, line 531: `ray.o` and `ray.distance` are updated but `offsetX,Y,Z` and `invDx,y,z` are not?

leMaik (approved):

chunky/src/java/se/llbit/math/Octree.java
aTom3333: Yep because we are leaving the function so they are no longer needed

Issue #739, opened 2020-11-27
Error messages can be created behind the chunky window


ProNoob135:
Sometimes error messages are created behind the chunky window, which is problematic because it cannot be moved or interacted with while they are displayed. (especially when there is a large amount of them from loading specific worlds.)
A temporary solution could be an option to disable error pop-ups.

blueplanet150:
Could you please describe a specific case where an error message is created behind the main window? Or could you please supply one of those worlds which produce many error messages? Then we can reproduce the bug.

jackjt8:
Issue was first reported over on leMaik's fork prior to write access being given to leMaik over here.

https://github.com/leMaik/chunky/issues/6 - The issue is typically with `Unknown Item ID` & (upto 2x) `Unknown octree Implementation` Warnings for which a new Warning window is spawned for each Unknown. Given only one window can have active focus all other Warnings tend to be pushed behind the Chunky window, and while Warnings exist you cannot simply minimise the Chunky window, you instead need to switch windows via the taskbar or alt+tab to `Ok` each Warning one after another.

![image](https://user-images.githubusercontent.com/3740917/104610743-48aa8400-567c-11eb-8b91-42b2ecf879e4.png)

Fixing player skin downloads certainly reduced the number of Warnings that can occur however it didn't solve the underlying UI.

Example world (trimmed): [Etho's LP.zip](https://github.com/chunky-dev/chunky/files/5815302/Etho.s.LP.zip)

The Error Window handles this much better as all errors end up in one Window rather than multiple.

PR #782, opened 2021-01-15
Reworked Error and Warning boxes.

StanleyMines:
This should fix chunky-dev/chunky#739.

Changes:
* Error box now puts itself ontop of the chunky window. (if possible, won't let it be put behind any other window).
* Group Warnings together and use ChunkyErrorDialog object.
* Having warnings up no longer prevents the main UI from being interacted with (side effect of using Stage instead of Alert).
* Pressing escape to dismiss ChunkyErrorDialog only triggers on KeyReleased, so you wont accidentally dismiss all errors and warnings when meaning to only dismiss one of the two.
* Log.Level enum now holds its name in a final string.
* Renamed ErrorDialog.fxml to DialogError.fxml and copied most of that into DialogWarning.fxml (Feel free to rename back).
* Other little bits of code cleanup.


jackjt8 (commented):
These changes function as expected. And makes dealing with multiple warnings much easier.

Just need to wait for an actual code review now..

leMaik (approved):
Awesome! :tada: 

Two minor things that I'll just change myself: `ErrorDialog` and `WarningDialog` are pretty good names. I'll also move the level name out of the `Level` enum (that's UI-related; maybe we want to allow translating them or something eventually).

chunky/src/java/se/llbit/chunky/ui/UILogReceiver.java
```diff
@@ -34,31 +34,28 @@
...
       case WARNING:
-        Platform.runLater(() -> {
-          Alert warning = Dialogs.createAlert(Alert.AlertType.WARNING);
-          warning.setContentText(message);
-          warning.show();
-        });
-        break;
+        if (warningDialog == null)
+          try { warningDialog = new ChunkyErrorDialog(level); }
+          catch (IOException e) { throw new Error("Failed to create warning dialog", e); }
```
NotStirred, line 51:
```suggestion
        if (warningDialog == null) {
          try { 
            warningDialog = new ChunkyErrorDialog(level); 
          } catch (IOException e) { 
            throw new Error("Failed to create warning dialog", e); 
          }
        }
```

chunky/src/java/se/llbit/chunky/ui/UILogReceiver.java
```diff
@@ -34,31 +34,28 @@
...
-            }
-          }
-          errorDialog.addErrorMessage(message);
-          errorDialog.show();
-        });
-        break;
+      default:
+        if (errorDialog == null)
+          try { errorDialog = new ChunkyErrorDialog(level); }
+          catch (IOException e) { throw new Error("Failed to create error dialog", e); }
```
NotStirred, line 57:
```suggestion
      default:
        if (errorDialog == null) {
          try { 
            errorDialog = new ChunkyErrorDialog(level); 
          } catch (IOException e) {
            throw new Error("Failed to create error dialog", e); 
          }
        }
```

NotStirred (changes requested):
Misc style changes

chunky/src/java/se/llbit/chunky/ui/UILogReceiver.java
```diff
@@ -34,31 +34,36 @@
...
       case WARNING:
-        Platform.runLater(() -> {
-          Alert warning = Dialogs.createAlert(Alert.AlertType.WARNING);
-          warning.setContentText(message);
-          warning.show();
-        });
-        break;
+        if (warningDialog == null) {
+          try { 
+            warningDialog = new ChunkyErrorDialog(level);
```
leMaik, line 51: :bug: This will result in the warning dialog being called "Info Summary" if the level of the first message is INFO.

PR #785, opened 2021-01-18
Fix module-path if javafxDir contains spaces

jackjt8:
Adds double quotes around javafxDir fixing launch issues with directories like `C:\Program Files\AdoptOpenJDK\jdk-11.0.9.11-hotspot\lib`.

And increment the ChunkyLauncher version to `v1.12.1`.

### DIFF
--- a/launcher/src/se/llbit/chunky/launcher/JavaFxLocator.java
+++ b/launcher/src/se/llbit/chunky/launcher/JavaFxLocator.java
@@ -60,9 +60,10 @@ public class JavaFxLocator {
     // add the options again so the launcher can use them for chunky
     cmd.add("--javaOptions");
     StringBuilder javaOptions = new StringBuilder();
-    javaOptions.append("--module-path \\\"");
-    javaOptions.append(javafxDir.toAbsolutePath().toString());
-    javaOptions.append("\\\" --add-modules ");
+    javaOptions.append("--module-path ");
+    // Escape the path twice to make the second launcher pass the options to Chunky retaining the double speechmarks (fixes paths with spaces)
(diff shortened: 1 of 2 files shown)


launcher/src/se/llbit/chunky/launcher/JavaFxLocator.java
```diff
@@ -60,9 +60,9 @@ private static void runWithJavafx(Path javafxDir, String[] args) {
     // add the options again so the launcher can use them for chunky
     cmd.add("--javaOptions");
     StringBuilder javaOptions = new StringBuilder();
-    javaOptions.append("--module-path ");
+    javaOptions.append("--module-path \\\"");
```
leMaik, line 63: Is this intended to literally add `\"` instead of `"` to the string?

jackjt8: It is required to escape twice due to how the options are given to launcher for the second invocation of the launcher that then give them to the actual chunky. Escaping once, while it does construct the correct `javaOptions`, results in the command being feed into chunky lacking the `"`.

leMaik: This would be a good comment. :+1: 
Maybe like this?

```
// escape the path twice to make the second launcher passing the options to Chunky work if it contains spaces
javaOptions.append("\\\"" + javafxDir.toAbsolutePath().toString() + "\\\"");
```

jackjt8: Slightly tweaked the wording and I also adopted the string concatenation you used to clean up the line (I wasn't sure if I could do that with the stringbuilder).

Issue #736, opened 2020-11-24
Better Live Preview


RocketNinja15:
You could add a better live preview. Maybe something base on the EEVE render from blender? This would include more realistic live previews that actually show a general idea of what your export image is.

ProNoob135:
This would entail adding a second, much more complex renderer to chunky. 
A simpler option is to add progressive sampling to the preview like blender has, which would allow it to run at a much better framerate without needing a separate renderer.

RocketNinja15:
Hmm, I like that idea

leMaik:
💯 agree with @ProNoob135 here. Adding a different renderer for the preview doesn't sound reasonable. It was also suggested to use rasterization for the preview and not ray tracing. That would also require a different renderer (and even a different implementation alltogether because we can't just convert _all_ blocks to triangles and render that without optimizing it, i.e. culling, merging trianges, etc.).

Progressive sampling sounds like a good idea, it would just depend on how fast the frame buffer copying is.

HackerDaGreat57:
Couldn't some sort of OpenGL canvas be applied instead of whatever there is right now?

There is a wxGLCanvas widget in wxWidgets that works perfectly, so I know that it is possible to implement OpenGL onto 2D GUI. but I'm not sure if OpenGL is supported in JavaFX.

(I come from C++ and just want to suggest this. I know almost nothing about Java, so please correct my thoughts if I'm wrong.)

RocketNinja15:
How about something like adaptive sampling, simplify, and render regions. Adaptive sampling like blender, will decide areas with less noise and not render that as much, simplify reduces the amount of ao bounces and bounces. Render regions will only render anything inside the viewport, and not calculate anything else outside of that @leMaik

ProNoob135:
The preview has no bounces, so adaptive sampling and "simplify" are meaningless here. Because of how the Octree works, "Render Regions" would increase preview time, not decrease it. (unless i'm misunderstanding)

jackjt8:
Given there are no bounces performed by the preview it is just simple ray casting and returning block colour for a given pixel. Performance is does not scale like you would expect. Sure a smaller chunk selection gives you a smaller octree that is faster to traverse... but it doesn't scale like what you expect. 40k chunks is only a touch faster than 1 million chunks.

We already have a viewport chunk selection option which would cull the selection but the issue then is with 1) the time it takes to generate an octree, 2) the memory footprint for it, and 3) the GC.

Issue #314, opened 2015-04-12
Cannot use resolution above 12800x12800


ghost:
whenever I try to it renders in 12800x12800 anyways. I have 10gb ram allocated.

leMaik:
12800 x 12800 = 491520000… seems to be the same issue as #492 then

StanleyMines:
I believe this issue is distinct from #492, as the canvas size of 12.8k by 12.8k (even tripled) gives 491m doubles (12.8^2 times 3 color values), as leMaik showed above, however that is less than `Integer#MAX_VALUE`, and I believe this is more of the silent JavaFx crash in the background due to too-large canvases, which was noted originally in #27.

Also see #786 and more specifically: https://github.com/chunky-dev/chunky/pull/786#issuecomment-773684014

leMaik:
@StanleyMines Allright, I'll reopen this then

Issue #27, opened 2013-01-19
Changing canvas size sometimes does not resize the preview window


llbit:
Sometime changing the canvas size does not work as intended. For example the double canvas size button may not work, and it does not work to increase the canvas size in large increments.

StanleyMines:
This is caused by JavaFX canvas size being too large: #492.

This will also be (pretty much) fixed, as soon as #786 is complete.

(Could we possibly mark this as duplicate and close this to use #492 issue thread instead?)

StanleyMines:
> ~This is caused by JavaFX canvas size being too large: #492~
> 
> ~This will also be (pretty much) fixed, as soon as #786 is complete.~
> 
> ~(Could we possibly mark this as duplicate and close this to use #492 issue thread instead?)~

Woops, actually, this issue is unique from #492 as 492 discusses larger canvases that exceed `Integer#MAX_VALUE / 3` in pixel count, not canvases which are smaller and still cause JavaFx to fail silently, like #314 did.

See: https://github.com/chunky-dev/chunky/pull/786#issuecomment-773684014

leMaik:
Since Chunky 2.x doesn't have a preview window anymore but instead a preview tab that doesn't need to be resized, I'll close this issue.

PR #786, opened 2021-01-19
Larger Canvas Size Support

### LINKED ISSUE #492: java.lang.NegativeArraySizeException & java.lang.ArrayIndexOutOfBoundsException when Rendering Images with "High" Resolution
Hi,

I'm currently trying to render Images with high resolutions and face some Issues.
When trying to render in a resolution higher than 35520x19980 the problems beginn to show up.
1. When Trying to render in 36480x20520 Chunky crashes after scene Loading with this Error
```
Loading scene took 0m 1s

Unchecked exception caused Chunky to close.
java.lang.NegativeArraySizeException
        at se.llbit.chunky.renderer.scene.Scene.initBuffers(Scene.java:366)
        at se.llbit.chunky.renderer.scene.Scene.importFromJson(Scene.java:2535)
        at se.llbit.chunky.renderer.scene.Scene.fromJson(Scene.java:2379)
        at se.llbit.chunky.renderer.scene.Scene.loadDescription(Scene.java:2501)
        at se.llbit.chunky.renderer.scene.Scene.loadScene(Scene.java:473)
        at se.llbit.chunky.renderer.scene.SynchronousSceneManager.loadScene(SynchronousSceneManager.java:146)
        at se.llbit.chunky.main.Chunky.doHeadlessRender(Chunky.java:159)
        at se.llbit.chunky.main.Chunky.main(Chunky.java:208)

Failed to launch Chunky. Command used:
/usr/lib/jvm/java-8-oracle/jre/bin/java -Xmx153600m -Dchunky.home=/home/user/.chunky -classpath /home/user/.chunky/lib/chunky-core-1.4.5.jar:/home/user/.chunky/lib/commons-math3-3.2.jar se.llbit.chunky.main.Chunky -texture /home/user/txtpack.zip -target 1 -render resolutiontest
```

2. When trying to render even Larger Images with a resolution of 61440x34560 the scene is successfull loaded, but chunky crashed while rendering with this error
```Loading octree took 0m 0s 1)
Loading regions took 0m 0s1)
Loading entities took 0m 0s)
Loading chunks: 100.0% (1,024 of 1,024)
Loading chunks took 0m 3s 1)
Finalizing octree took 0m 1s
Loading scene took 0m 13s
Rendering: 0.0% (0 of 1) [ETA=1011703407:30:07]

Render worker 1 crashed with uncaught exception.
java.lang.ArrayIndexOutOfBoundsException: 2075259000
        at se.llbit.chunky.renderer.RenderWorker.work(RenderWorker.java:135)
        at se.llbit.chunky.renderer.RenderWorker.run(RenderWorker.java:71)
```

Both Issues are showing up regardless of the Map I choose.

I'm using Chunky 1.4.5 and java version 1.8.0_181 64 Bit on a debian 9.5
Regarding the HW: I'm running a Xeon w-2145 with 256 GB ram.
If you need any more Information please let me know.

Thanks for your help.

kind regards 
Max

StanleyMines:
Primary goal is to fix #492 which caused crashes at various "large" resolutions for different people.

* SampleBuffer class to hold the sample buffer and alpha buffer of render as double[][] instead of double[]
* Changed BitmapImage class to use int[][] instead of int[], to be consistent with SampleBuffer
* Made BitmapImage.data protected for better encapsulation.
* Refactor things to work with the new SampleBuffer and BitmapImage. And use the getters and setters instead of directly accessing array (which is faster in most cases too).
* Fixed wrong comment in FloatingPointCompressor class for decompression logic
* PngFileWriter now takes BitmapImage instead of int[]; PfmFileWriter does similar internally.
* BitmapImage function local variables refactored to correct names.
* BitmapImage#vFlipped() now uses System.arraycopy and is thus much faster.

Known Issues:
* Denoiser plugin no longer can directly access double[], and will need to be updated to use the SampleBuffer (and while changing that, may as well switch to the existing PfmFileWriter)
* Limits "Render Preview" canvas size in RenderCanvasFx class to prevent strange JavaFx crashes. This should be changed in the future. The current state can only view the top left corner of the render. I might change this to only show the middle 4k x 4k section, but ideally we should have some zoomed out preview which isn't so big that it crashes (and doesnt take much CPU time away from actual rendering).


NotStirred:
As discussed on discord, this is a temporary solution at best, and proper render regions would be the better approach.
How much does this PR hurt performance? Some profile results would be nice on multiple setups, I'll get some results on my machine later today.
High sample counts will be the most affected most likely

StanleyMines:
This pull request is marked as a `draft` pull request, as it has a few issues with efficiency as well as compatibility. In its current state, I do not expect it to be accepted, without some decent improvements.

I believe it has done better than master in most of the test cases due to a couple slight changes in a few of the files that were made, so I will be creating a separate pull request to fix a couple of those things, should they prove to be the cause of increased efficiency.

jackjt8:
```960x540 2048 SPP

2.3.0-55-gad38006d0 -  3399s, 312347 SPS

max-res-fix - 421s, 314,944 SPS

2% faster

---

1920x1080 32 SPP

2.3.0-55-gad38006d0 - 435s, 304,495 SPS

max-res-fix - 421s, 314,944 SPS

3.2% faster

---

3840x2160 128 SPP

2.3.0-56-g4419bd7e - 3377s, 314,331 SPS

max-res-fix - 3273s, 324,291 SPS

3% faster
```

At the very least this Draft does not reduce performance.

leMaik:
@StanleyMines This PR (along with #798) is pretty much exactly what I'd need to add non-uniform sample distribution (canvas cropping, selective rendering, …). The performance seems to be slightly better than before, this PR looks pretty good.

Is there anything (other than the `BitmapImage.java` conflict) that would need to be changed before merging this?

chunky/src/java/se/llbit/chunky/renderer/RenderWorker.java
```diff
@@ -133,10 +135,7 @@ private void work(RenderTile tile) {
             sg += ray.color.y;
             sb += ray.color.z;
           }
-          double sinv = 1.0 / (scene.spp + manager.sppPerPass);
-          samples[offset + 0] = (samples[offset + 0] * scene.spp + sr) * sinv;
-          samples[offset + 1] = (samples[offset + 1] * scene.spp + sg) * sinv;
-          samples[offset + 2] = (samples[offset + 2] * scene.spp + sb) * sinv;
+          samples.addSample(x, y, scene.spp, manager.sppPerPass, sr, sg, sb);
```
leMaik, line 138: Maybe we could change this so that the `SampleBuffer` "knows" its spp (or later spp at the specific pixel) so that only `manager.sppPerPass` needs to be passed here as a "weight".

chunky/src/java/se/llbit/chunky/renderer/RenderWorker.java
```diff
@@ -195,19 +192,17 @@ private void work(RenderTile tile) {
...
-          samples[(y * width + x) * 3 + 2] = ray.color.z;
+          samples.setPixel(x,y,ray.color.x,ray.color.y,ray.color.z);
 
           scene.finalizePixel(x, y);
 
           if (firstFrame) {
             if (y % 2 == 0 && x < (width - 1)) {
               // Copy the current pixel to the next one.
-              scene.copyPixel(y * width + x, 1);
+              scene.copyPixel(x, y, 1);
```
leMaik, line 202: :bulb: In the future we could add a `PreviewSampleBuffer` that supports not having a sample for every pixel and uses the color of the nearest neighbor sample to create an image. That would solve #736

chunky/src/java/se/llbit/chunky/renderer/Renderer.java
```diff
@@ -66,7 +67,7 @@
   void withBufferedImage(Consumer<BitmapImage> bitmap);
 
   interface SampleBufferConsumer {
-    void accept(double[] samples, int width, int height);
+    void accept(SampleBuffer samples, int width, int height);
```
leMaik, line 70: The sample buffer already has a `width` and a `height` so this could be removed. Actually the entire `SampleBufferConsumer` could then be replaced with `Consumer<SampleBuffer>`

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
```diff
@@ -0,0 +1,125 @@
...
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with Chunky.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package se.llbit.chunky.renderer.scene;
+
+import se.llbit.math.ColorUtil;
+
+public class SampleBuffer {
+  public final int width;   // = rowSize/3, used for alpha
```
leMaik, line 22: "used for alpha"?

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
```diff
@@ -0,0 +1,125 @@
...
+  public final int rowSize; // = width * 3, used for samples
+
+  public final int height;   // = rowCount
+  public final int rowCount; // = height
+
+  protected final double[][] samples;
+  // protected int[][] spp;     // for non-uniform sample distribution
```
leMaik, line 29: As mentioned above, until we have that, maybe store the spp here, i.e. `protected int spp` to simplify the `addSample` method.

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
```diff
@@ -0,0 +1,125 @@
...
+        System.arraycopy(toCopy.alpha[row], 0, this.alpha[row], 0, width);
+    }
+  }
+
+  public long pixelCount() {
+    return width*(long)height;
+  }
+
+  public long sampleCount() {
+    return 3L*width*height;
```
leMaik, line 66: This is not the sample count. The sample count is `spp * width * height`.

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
```diff
@@ -0,0 +1,125 @@
...
+
+  public void set(int x, int y, int o, double value) {
+    samples[y][x*3+o] = value;
+  }
+
+  public int getArgb(int x, int y) {
+    return ColorUtil.getArgb(get(x,y,0),get(x,y,1),get(x,y,2),0xFF);
+  }
+
+  public void addSample(int x, int y, int old_spp, int spp_increase, double r, double g, double b) {
```
leMaik, line 89:
```suggestion
  public void addSample(int x, int y, int weight, double r, double g, double b) {
```

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
```diff
@@ -0,0 +1,125 @@
...
+    double sinv = 1.0 / (old_spp + spp_increase);
+    set(x,y,0,sinv*(get(x,y,0)*old_spp+r));
+    set(x,y,1,sinv*(get(x,y,1)*old_spp+g));
+    set(x,y,2,sinv*(get(x,y,2)*old_spp+b));
```
leMaik, line 93: Would this be faster if it wouldn't call `set` three times? The index just changes by one for each line, no need to re-calculate everything. Same for `setPixel`? :thinking:

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
```diff
@@ -0,0 +1,125 @@
...
+    set(x,y,2,b);
+  }
+
+
+  public void alphaInit() {
+    alpha = new byte[rowCount][rowSize];
+  }
+
+  public byte alphaGet(int x, int y) {
+    return alpha[y][x];
```
leMaik, line 108: If `alpha == null`, just return `1` (instead of crashing)

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  public void alphaSet(int x, int y, byte value) {
+    alpha[y][x] = value;
```
leMaik, line 112: If `alpha == null`, maybe lazy-initialize it here?

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  public byte[] alphaCompile() {
+    // This fails at an image size of about 46 000 x 46 000. (2100 megapixels)
+    if (((long)width)*height>Integer.MAX_VALUE)
+      throw new IndexOutOfBoundsException("Alpha too large. Cannot fit within a single array. Single arrays should only be used for small images, such as textures or MapTiles, not full renders!");
+
+    byte[] copy = new byte[width*height];
```
leMaik, line 120: Fill `copy` with `1` and return it if `alpha == null`?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -358,8 +357,8 @@ public static void exportToZip(String name, File targetFile) {
   public synchronized void initBuffers() {
     frontBuffer = new BitmapImage(width, height);
     backBuffer = new BitmapImage(width, height);
-    alphaChannel = new byte[width * height];
-    samples = new double[width * height * 3];
+    samples = new SampleBuffer(width,height);
+    samples.alphaInit();
```
leMaik, line 361: If alpha is _always_ enabled anyway, it could just be initialized in the constructor. If not, it could only be initialized if the sky is set to transparent to improve memory usage.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -440,8 +439,8 @@ public synchronized void copyState(Scene other, boolean copyChunks) {
       height = other.height;
       backBuffer = other.backBuffer;
       frontBuffer = other.frontBuffer;
-      alphaChannel = other.alphaChannel;
       samples = other.samples;
+      // alphaChannel = other.alphaChannel;
```
leMaik, line 443: Since alpha is now in the sample buffer, this can be removed

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  public void setPixel(int x, int y, double r, double g, double b) {
+    set(x,y,0,r);
+    set(x,y,1,g);
+    set(x,y,2,b);
+  }
+
+
+  public void alphaInit() {
```
leMaik, line 103:
```suggestion
  public void enableAlpha() {
```

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
```diff
@@ -0,0 +1,125 @@
...
+
+  public byte alphaGet(int x, int y) {
+    return alpha[y][x];
+  }
+
+  public void alphaSet(int x, int y, byte value) {
+    alpha[y][x] = value;
+  }
+
+  public byte[] alphaCompile() {
```
leMaik, line 115:
```suggestion
  public byte[] getRGBAAlphaChannel() {
```

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
```diff
@@ -0,0 +1,125 @@
...
+    set(x,y,1,g);
+    set(x,y,2,b);
+  }
+
+
+  public void alphaInit() {
+    alpha = new byte[rowCount][rowSize];
+  }
+
+  public byte alphaGet(int x, int y) {
```
leMaik, line 107:
```suggestion
  public byte getAlpha(int x, int y) {
```

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
```diff
@@ -0,0 +1,125 @@
...
+    return alpha[y][x];
+  }
+
+  public void alphaSet(int x, int y, byte value) {
```
leMaik, line 111:
```suggestion
  public void setAlpha(int x, int y, byte value) {
```

chunky/src/java/se/llbit/chunky/resources/BitmapImage.java
```diff
@@ -176,4 +171,32 @@ public static BitmapImage concatY(BitmapImage a, BitmapImage b) {
...
+    int[] copy = new int[width*height];
+    for (int row = 0; row<height; row++)
+      System.arraycopy(data[row],0,copy,width*row,width);
+    return copy;
+  }
+
+  @Deprecated
+  public int[] getRow_Unsafe(int row) {
+    return data[row];
+  }
```
leMaik, line 201: Are these methods used anywhere in Chunky? I can update the denoiser if that's the only issue.

chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
```diff
@@ -57,6 +58,16 @@
...
 
+  // This sets the maximum "Render Preview" canvas size to 4k by 4k,
+  // for sizes any larger cause a strange crash in JavaFX.
+  // Currently, this only shows the 4k by 4k top-left corner of renders
+  // which are larger than this. Future task: be able to pan around on
+  // the Render Preview canvas. (Kind-of exists, but again, loading the
+  // whole image leads to mysterious crashes.)
+  // Oh, and also, the crash happens at different minimum canvas sizes
+  // unique to each computer.
+  private static final int MAX_CANVAS_SIZE = 4096;
```
leMaik, line 69: This is fine and shouldn't block us from merging this PR. Currently Chunky just crashes, so this is far better! :+1:

leMaik (changes requested):
I don't see why this shouldn't be merged. It's _great_! :open_mouth:

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
StanleyMines: Alpha buffer has only 1 value for pixel, where as "rowSize" is 3x as much since R G and B samples in the main array. Fixed to be clearer in next commit.

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
StanleyMines: Next commit is partial implementation of this, though `int Scene#spp` still exists and is used in most locations.

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
StanleyMines: Ah, I was going for "number of double stored in this array"

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
StanleyMines: I think this is clearer and much easier to read; I trust that the JIT compiler should be able to figure that out and optimize that. This branch seems to run faster than without anyways, so ¯\_(ツ)_/¯

chunky/src/java/se/llbit/chunky/resources/BitmapImage.java
StanleyMines: toIntArray() is used in `MapTile#drawImage(BitmapImage)`
toIntArray() is used in `FxImageUtil#toFxImage(BitmapImage)`

getRow_Unsafe() is used in `RenderCanvasFx#forceRepaint(BitmapImage)`

chunky/src/java/se/llbit/chunky/renderer/scene/SampleBuffer.java
leMaik: I don't think the JIT will know that the function returns the same result +1 for the last two calls. 
But you're right, it's not that big of a deal.

StanleyMines:
Un-'draft'ed this PR.

Still need to fix test cases, though I think everything else is working fine atm. (@leMaik knows more about what exactly needs to be done.)

chunky/src/java/se/llbit/chunky/renderer/renderdump/FloatingPointCompressor.java
```diff
@@ -149,52 +151,47 @@ public double decodeSingle(byte header, BufferedInputStream input) throws IOExce
...
+    LongConsumer pixelProgress
   ) throws IOException {
     try (BufferedOutputStream out = new BufferedOutputStream(output)) {
-      if (input.length % 3 != 0)
-        throw new IllegalArgumentException("Dump doesn't have a multiple of 3 values");
-
-      int pixels = input.length / 3;
-      int size = pixels - 1;
-
+      long size = input.numberOfPixels() - 1;
```
ShirleyNekoDev, line 159: I like this change! Could you do the same in the `decompress` method?

leMaik: Will do :+1:

StanleyMines:
Note: Hard coded 4096x4096 display instead of full image is not sufficient. This still crashes chunky on my other computer, when using my second monitor.

StanleyMines:
> Note: Hard coded 4096x4096 display instead of full image is not sufficient. This still crashes chunky on my other computer, when using my second monitor.

This seems to depend on the windowed size of the chunky GUI itself. When it is large (5000x3000 ish) it crashes at small sizes (2.5k by 2.5k) when window is small (900x300) it can do a bit larger (5kx5k), although I cannot seem to get larger than 5kx5k even on single display, and minimal window size, at least on this computer. (Java 11.0.3 + JavaFX LTS 11.0.2; as well as java 1.8.0u281)

jackjt8:
Both the UI and Canvas need access to the same Texture from what I can gather with JavaFX. The maximum supported texture size depends on the GPU, driver, and D3D/OpenGL pipeline used.

`-Dprism.verbose=true` should give us a quick readout of `Maximum supported texture size: x` which we could use as the baseline but we would also need to factor the UI size. I'm not sure about a programmatic way we can check the hardware capabilities but at least with OpenGL calling `GetIntegerv(GL_MAX_TEXTURE_SIZE)` would work.

aTom3333:
I've not followed this too closely but what I suggest is to limit the on-screen canvas to have its biggest dimension be in some interval, let's say [1000; 2000] px. You then take the biggest dimension of the real canvas and you divide it by 2 until it falls into the interval. For example, for a (imaginary) 7600*5100 canvas, you do the division twice to get an on-screen canvas of 1900*1275 which javafx supports without issue and each on-screen pixel correspond to 4*4 pixels in the real canvas, that's a round number easy to deal with.
To get the on-screen canvas from the real canvas can be done in the finalizePixel function so that it only reads 1 sample out of x (4 in this example) and write it to the backBuffer.

leMaik:
@StanleyMines I'd like to drop the changes done to `BitmapImage`. They are not required for anything but saving the final image. I'd like to keep this as simple as possible, which would be:

- Have the `SampleBuffer` at full resolution and with variable spp (already implemented now)
- Store number of samples of every pixel in the dump file
- Specify area to be rendered before starting the rendering
- Only display the area that is being rendered, show spp for that area… behave like _this area_ is what is being rendered
- Merge variable-spp dumps (and also merge it after rendering)

This would allow people to manually split their images into tiles and merge them, it could also be automated by automatically generating the json files (we'd need some new property like `"crop": [x,y,w,h]` in it) and would also enable area-based tile splitting in ChunkyCloud.

StanleyMines:
Added Downscaleing to render previews:
- Downscaled preview should no longer crash on 99% of computers, and will display entire render.

Fixed Merging:
- wrong spp indices in SampleBuffer methods
- missing spp coefficient for merge samples
- merge tests didn't set spp for whole scene sample buffer

Future things to consider doing:
- We should really have some way to set RenderCanvasFx#previewShouldSubsample, though I haven't implemented that yet.
- We should probably update preview scale %s to reflect the downscaling. (And this should work great with @alexhliu's new auto-scaling preview PR)
- Preview render should probably also only show up to the max shown size instead of rendering far more pixels than will actually be shown. It's a quick preview after all, is it not?
- (from leMaik's comment) 💡 In the future we could add a `PreviewSampleBuffer` that supports not having a sample for every pixel and uses the color of the nearest neighbor sample to create an image. That would solve #736

StanleyMines:
This should fix the issue described in #492 where a single array cannot hold the whole image, however I just realized that #492 does not mention the issue of `strange JavaFx crashes` as I mentioned in the first comment of this PR, which is the issue that is discussed by #27 and #314. This JavaFx crash has an annoying part to it that it fails and crashes the JavaFx thread, where there is no chunky code on the stack trace at all, with no outwardly noticeable indication of the crash (other than resizing the window might lead to flickering or red boxes, which is not something that I think most people are constantly doing when waiting for canvas size to increase). Since it is entirely in JavaFx code, there isn't an easy way that we can wrap it in a try-catch and either restart it with a smaller canvas or notify the user with a warning/error dialog.

StanleyMines:
Working on adding rendering subareas and spp to dump; converting PR back to draft until that works with dumps and merging. (hopefully tonight or tomorrow)

StanleyMines:
This should complete:

- [x] Store number of samples of every pixel in the dump file
- [x] Specify area to be rendered before starting the rendering
- [x] Only display the area that is being rendered, show spp for that area… behave like this area is what is being rendered

However there are still a few issues: 
 * namely dump is uncompressed and thus larger,
 * and dunno why, but its also many many times slower. (eg 1080p scene is ~10-15 seconds per save/load). I chose to take out compression for this iteration due to having trouble troubleshooting the compressed dumps and thought it might have been a contributor to the slow speed (that might have helped by a bit, but not too sure). Could probably be added back now.
 * loading scenes that are cropped via `scene.json` usually render wrong at first. Something to do with the canvas and double buffering I think, but I can't seem to resolve this. Current fix is to just render one more spp. (Merging doesn't seem to have this issue for some reason)
 
 Since it is getting close to ready and most of the new features work, I'll undraft this, but plan on continuing to try to fix these few issues.

StanleyMines:
The last few commits have been fixes to make it more stable, and fix the problem of extremely slow saving/loading times for dumps. SPS number should also be fixed. SPP is also interleaved with the sample values, and thus merging no longer needs to necessarily read the entire dump before merging the first few pixels. This also means more than two dumps could be merged together, with little to no memory overhead.

Other notes:
Dump is still not compressed, so about a 20% size increase over before, if you include spp being added.

Still need work from before:
> * loading scenes that are cropped via scene.json usually render wrong at first. Something to do with the canvas and double buffering I think, but I can't seem to resolve this. Current fix is to just render one more spp. (Merging doesn't seem to have this issue for some reason)
Which is a bit finicky, but all the features do work.

For the future:
* Add way to edit cropping values without needing to modify scene.json (NOT planned to be in this PR.)

StanleyMines:
Updates:
Merging fixed in more ways than one.
Load Scene UI has been updated.

Things left to do in this PR:
All things set out to do have been accomplished.

For the future (probably not this PR):
* Way to subdivide renders within the UI (Currently requires manual editing of the scene json file)
* Allow compression of dumps (possibly changing order of chunks within file for better compression (deflate on SPP values))

StanleyMines:
@leMaik said:

> Dump compression _needs_ to come back [...]

So I guess there is (at least) one more thing to do here...

StanleyMines:
IntegratedSppDump now uses gzip to compress all of the dump after the header. This is slower than the floating point compressor by a bit, and is far far slower than uncompressed, but is capable of saving ~10% size. Merge testing is currently not functional for the new features though.

Next thing I plan to work on is adding some (jank) UI way to subdivide a scene and possibly move to the next subarea to render. (I'll fix dump merge test once theres a better solution to compression... atm I don't think its worth putting in more effort to get the test to pass (more than implementing the compression), if I know its working right now and its probably going to be changed very soon...)

leMaik:
@StanleyMines Sounds good :clap: Maybe do the UI on a new branch so that we can finally get this feature merged when the compression is ready

PR #788, opened 2021-01-20
BitmapImage Nameing & Array Copy Fix

StanleyMines:
The "flipping" methods' local variables were _flipping_ named rotated!!!

`System.arraycopy` of rows in vertical flip would be faster than individually copying each pixel. (Used in bed & chest texturing)


StanleyMines (commented):
Textures for bed and chests (and trapped chests) seem to work correctly when compared to ingame, for all orientations and colors, as well as with and without texturepack applied in Chunky.

leMaik:
Smart micro optimization for the vertical case :wink:

PR #791, opened 2021-01-22
PackedOctree Javadoc

StanleyMines:
Javadoc'd all methods in PackedOctree with more information and explanation of how it works, so that it will be easier for the next reader to understand. Also updated a few things that were out of date (such as 2 ints per node in a few comments).

Also touched the methods set and storeNode and getNodeIndex in minor ways (others may have had minor refactors).

### DIFF
--- a/chunky/src/java/se/llbit/math/PackedOctree.java
+++ b/chunky/src/java/se/llbit/math/PackedOctree.java
@@ -181,8 +181,7 @@ public class PackedOctree implements Octree.OctreeImplementation {
    * Method also marks returned index as used (by removing from free list or
    * incrementing size)
    *
-   * @return the index at the beginning of a free space in the array of size 16 ints
-   *         (8 nodes)
+   * @return the index at the beginning of a free space in the array of size 8 ints
    */
   private int findSpace() {
     // Look in free list
--- a/chunky/src/java/se/llbit/math/PackedOctree.java
+++ b/chunky/src/java/se/llbit/math/PackedOctree.java
@@ -34,30 +34,21 @@ import static se.llbit.math.Octree.*;
  */
 public class PackedOctree implements Octree.OctreeImplementation {
   /**
-   * The whole tree data is store in a int array
+   * The entirety of the octree data is store in an int array.
    * <p>
-   * Each node is made of several values :
-   * - the node type (could be a branch node or the type of block contained)
-   * - optional additional data
-   * - the index of its first child (if branch node)
+   * Each node is made of a single integer value, which is either posative or negative.
+   * - Positive index -> Branch node; int is the index of first child (the other 7 follow sequentially).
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/math/PackedOctree.java
```diff
@@ -163,30 +202,31 @@ private int findSpace() {
...
 
     // We need to grow the array
-    long newSize = (long) Math.ceil(treeData.length * 1.5);
+    long newSize = (long) Math.ceil(treeData.length * ARRAY_RESIZE_MULTIPLIER);
     // We need to check the array won't be too big
     if(newSize > (long) MAX_ARRAY_SIZE) {
       // We can allocate less memory than initially wanted if the next block will still be able to fit
       // If not, this implementation isn't suitable
-      if(MAX_ARRAY_SIZE - treeData.length > 8) {
+      if(MAX_ARRAY_SIZE - size > 8) {
```
StanleyMines, line 210: Note: there might be a small chance we can fit in a node between the end of used array (size), and MAX_ARRAY_SIZE, which would be missed if split by treeData.length.

chunky/src/java/se/llbit/math/PackedOctree.java
```diff
@@ -246,41 +294,55 @@ private boolean nodeEquals(int firstNodeIndex, Octree.Node secondNode) {
...
     }
-    int finalNodeIndex = treeData[parents[0]] + position;
-    treeData[finalNodeIndex] = -data.type; // Store negation of the type
+    // store type into final node (this specific block coordinate's node)
+    treeData[nodeIndex] = -data.type; // Negation of BlockPalette type stored
 
-    // Merge nodes where all children have been set to the same type.
-    for(int i = 0; i <= parentLevel; ++i) {
+    // Merge nodes where all children have been set to the same type, starting from the bottom.
+    for(int i = 0; i < depth; ++i) {
```
StanleyMines, line 342: Note: This is equivalent and less obscure; loop will break anyways if goes too high.

chunky/src/java/se/llbit/math/PackedOctree.java
```diff
@@ -132,25 +164,32 @@ public PackedOctree(int depth, long nodeCount) {
...
+   *
+   * Checks free list first, then tries to append at the end of the array,
+   * reallocating and extending the array if needed.
+   *
+   * Method also marks returned index as used (by removing from free list or
+   * incrementing size)
    *
-   * @return the index at the beginning of a free space in the array of size 16 ints (8 nodes)
+   * @return the index at the beginning of a free space in the array of size 16 ints
+   *         (8 nodes)
```
aTom3333, line 185:
```suggestion
   * @return the index at the beginning of a free space in the array of size 8 ints
   *         (8 nodes)
```

aTom3333 (approved):
Nice job, just need to hope the implementation won't change too much now

Issue #41, opened 2013-02-07
Render entities


llbit:
It would be great if entities could be rendered.

Entity support would enable rendering these things:

- [ ] Mobs
  - [ ] Copper golem
  - [ ] _placeholder for all other not yet supported mobs_
- [ ] Animals
  - [x] Sheep
  - [x] Cow
  - [x] Pig
  - [x] Chicken
  - [x] Mooshroom
  - [x] Squid
  - [ ] _placeholder for all other not yet supported animals_
- [x] Paintings
- [x] Signs with proper text
- [x] The book hovering over enchanting tables
- [x] The book on lecterns
- [ ] Ender crystals
- [ ] Particles
- [x] Beacon beam effect
- [x] Head blocks

Static objects are a priority due to higher benefit/effort ratio. Mobs would need a posing interface. Initially poses could be coded in scene settings, and some presets can be a simple in-between measure.

llbit:
Current progress:

* ✔Paintings
* ✔Text on signs
* ✔Mob heads

jason-green-io:
yay entities!

could we get a way to disable loading entities in the config file for us headless people?

llbit:
@jason-green-io I have add the `renderActors` scene variable now in the `scene.json` file, which currently does not have a corresponding UI element.

jason-green-io:
Great news, thanks Jesper!

SpartanOverdrive:
I'm guessing this thread is the explanation as to why my armor stands, item frames, etc. never render?

leMaik:
Some animals are now rendered in the latest 2.5.0 snapshots. 🎉

PR #792, opened 2021-01-22
Beam upgrades

073TheBoss:
This is a big one compared to what I normally do. This PR includes changes to Beacon Beams including allowing rotation, saving and loading custom colors, and loading colors from the octree.

default_2021-01-22_12-27-45-500

I've removed the beacon_beam material from the Materials Tab and put that functionality into the Entities Tab to enable the user to change those properties for any beam segment along with the color.

image

Related to #41

### DIFF
--- a/chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
+++ b/chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
@@ -97,6 +97,7 @@ public class BeaconBeam extends Entity implements Poseable {
           firstColor = color;
           foundFirst = true;
         } else {
+          //If this is not the first block to change the color, the new beam color is the average of the first color and the color of this block
           float[] foundColor = new float[3];
           float[] baseColor = new float[3];
 
@@ -172,7 +173,7 @@ public class BeaconBeam extends Entity implements Poseable {
         }
       }
     }
-    return -1;
+    return -1; // Return -1 if block doesn't change beam color
   }
 
   @Override
--- a/chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
+++ b/chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
@@ -1,12 +1,18 @@
 package se.llbit.chunky.entity;
 
+import se.llbit.chunky.block.Beacon;
+import se.llbit.chunky.chunk.BlockPalette;
+import se.llbit.chunky.world.Material;
 import se.llbit.chunky.world.material.BeaconBeamMaterial;
 import se.llbit.json.JsonMember;
 import se.llbit.json.JsonObject;
 import se.llbit.json.JsonValue;
+import se.llbit.math.ColorUtil;
+import se.llbit.math.Octree;
 import se.llbit.math.Quad;
 import se.llbit.math.Transform;
 import se.llbit.math.Vector3;
+import se.llbit.math.Vector3i;
 import se.llbit.math.Vector4;
 import se.llbit.math.primitive.Primitive;
 import se.llbit.util.JsonUtil;
@@ -48,13 +54,12 @@ public class BeaconBeam extends Entity implements Poseable {
   private final JsonObject pose;
   private double scale = 1;
   private int height = 256;
-  private Map<Integer, BeaconBeamMaterial> materials = new HashMap<>();
+  private final Map<Integer, BeaconBeamMaterial> materials = new HashMap<>();
 
   public BeaconBeam(Vector3 position) {
     super(position);
     this.pose = new JsonObject();
     pose.add("all", JsonUtil.vec3ToJson(new Vector3(0, 0, 0)));
-    materials.put(0, new BeaconBeamMaterial(0xFFFFFF));
   }
 
   public BeaconBeam(JsonObject json) {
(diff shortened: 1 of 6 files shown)


chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
```diff
@@ -48,13 +54,12 @@
   private final JsonObject pose;
   private double scale = 1;
   private int height = 256;
-  private Map<Integer, BeaconBeamMaterial> materials = new HashMap<>();
+  private final Map<Integer, BeaconBeamMaterial> materials = new HashMap<>();
```
leMaik, line 57: _fastutils_ has a better map for this: http://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/ints/Int2ObjectOpenHashMap.html

But better ask @NotStirred, he knows this stuff better than me

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
```diff
@@ -65,12 +70,108 @@ public BeaconBeam(JsonObject json) {
 
     JsonObject materialsList = json.get("beamMaterials").object();
     for (JsonMember obj : materialsList.members) {
-      BeaconBeamMaterial mat = new BeaconBeamMaterial(0xFFFFFF);
+      BeaconBeamMaterial mat = new BeaconBeamMaterial(0xF9FFFE);
```
leMaik: Why is this not white anymore?

073TheBoss: I changed it to use the actual values that Minecraft uses to tint the beam, which isn't pure white.
https://minecraft.gamepedia.com/Dye#Color_values

leMaik: `private static final int DEFAULT_BEAM_COLOR`?

073TheBoss: Yeah, in the last commit, it uses a constant in BeaconBeamMaterial

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
```diff
@@ -61,27 +67,135 @@ public BeaconBeam(JsonObject json) {
...
-        .translate(0.5, 0.5, 0.5)
-        .translate(position.x + offset.x, position.y + offset.y, position.z + offset.z));
+  public void loadDataFromOctree(Octree octree, BlockPalette palette, Vector3i origin) {
+    int firstColor = BeaconBeamMaterial.DEFAULT_COLOR;
+    boolean foundFirst = false;
+    this.materials.put(0, new BeaconBeamMaterial(BeaconBeamMaterial.DEFAULT_COLOR));
+
+    //Start i at 1 so the first beacon is not checked.
+    //Stop i at 256 even if the beam is taller because the Octree will wrap the coordinates.
+    for (int i = 1; i < height && (i + position.y) < 256; i++) {
```
leMaik, line 87: This will not work for smaller octrees that use y cut-off or bigger octrees (1.17 world height or future cubic chunks support). You'll have to get the maximum y value from the octree somehow.

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
```diff
@@ -61,27 +67,135 @@ public BeaconBeam(JsonObject json) {
...
-        .scale(scale)
-        .rotateY(allPose.y) // Ignore x and z rotation because it breaks the beam into segments
-        .translate(0.5, 0.5, 0.5)
-        .translate(position.x + offset.x, position.y + offset.y, position.z + offset.z));
+  public void loadDataFromOctree(Octree octree, BlockPalette palette, Vector3i origin) {
+    int firstColor = BeaconBeamMaterial.DEFAULT_COLOR;
+    boolean foundFirst = false;
+    this.materials.put(0, new BeaconBeamMaterial(BeaconBeamMaterial.DEFAULT_COLOR));
+
+    //Start i at 1 so the first beacon is not checked.
```
leMaik, line 85: The beacon _block_ itself?

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
```diff
@@ -61,27 +67,135 @@ public BeaconBeam(JsonObject json) {
...
+          ColorUtil.getRGBComponents(firstColor, baseColor);
+
+          int newColor = ColorUtil.getRGB((foundColor[0] + baseColor[0]) / 2.0f, (foundColor[1] + baseColor[1]) / 2.0f, (foundColor[2] + baseColor[2]) / 2.0f);
+          this.materials.put(i, new BeaconBeamMaterial(newColor));
+        }
+      }
+    }
+  }
+
+  private int colorfromMaterial(Material blockMaterial) {
```
leMaik, line 109: (nit) `getColorFromMaterial`; but maybe even change it to `getColorFromBlock(Block block)`. The fact that `Block` _inherits_ from `Material` is weird anyway and might be refactored to blocks having a `material` field instead.

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
```diff
@@ -61,27 +67,135 @@ public BeaconBeam(JsonObject json) {
...
+    if(blockMaterial instanceof Beacon) {
+      return 0xF9FFFE;
```
leMaik, line 111:
```suggestion
      return BeaconBeamMaterial.DEFAULT_COLOR;
```

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
```diff
@@ -61,27 +67,135 @@ public BeaconBeam(JsonObject json) {
...
+    }
+    if(blockMaterial.name.endsWith("_stained_glass") || blockMaterial.name.endsWith("_stained_glass_pane")) {
```
leMaik, line 113: This could probably be refactored to use a regex at some point, e.g. `(minecraft:)?(.+?)_stained_glass(_pane)?` and then it will have the color in the second group and is also future-proof for json model loading (where all block names will be namespaced).

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
```diff
@@ -61,27 +67,135 @@ public BeaconBeam(JsonObject json) {
...
+      String prefix;
+      if(blockMaterial.name.endsWith("_stained_glass")) {
+        prefix = blockMaterial.name.substring(10, blockMaterial.name.length() - 14);
+      } else {
+        prefix = blockMaterial.name.substring(10, blockMaterial.name.length() - 19);
+      }
+      switch (prefix) {
+        default:
```
leMaik, line 121: Should we log a warning in this case? An unsupported glass color is most likely a new Minecraft feature or a Chunky bug.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1184,6 +1184,10 @@ public synchronized void loadChunks(TaskTracker progress, World world,
       waterOctree.endFinalization();
     }
 
+    for (Entity entity : actors) {
```
leMaik, line 1187: For the sake of completeness, this should probably also be called for entities that are not actors. Although I don't know any entities that would need block information?

leMaik (changes requested):
Nice! :+1: 

Small suggestion for another PR: A reset button would be cool, i.e. load the beam segment colors from the world after they were changed in the materials tab.

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
NotStirred: Yeah having a boxed `int` to `Object` map is never gonna be as fast as the fastutils `Int2ObjectOpenHashMap`

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
073TheBoss: I'll go ahead and add another loop for it, just to future-proof it.

chunky/src/java/se/llbit/chunky/entity/BeaconBeam.java
073TheBoss: Changed to check the Octree.isInside(...) method instead of using 256.

leMaik (approved):

PR #793, opened 2021-01-22
Fix for missing JFX with Actions

jackjt8:
See https://github.com/actions/setup-java/issues/18 and docs for setup-java.


leMaik:
Great! I wonder why it even worked before…

PR #798, opened 2021-01-24
Render dump refactoring

### LINKED ISSUE #783: Dump merging still uses old dump format
The dump format in `saveDump(RenderContext context, TaskTracker progress)` was updated in #756 (with fallback while loading),
but the code in `mergeDump(File dumpFile, TaskTracker taskTracker)`  can only load the old format.

ShirleyNekoDev:
Fixes #783

This refactoring moves all the scene dump processing logic into it's own package.
The 3 primary use cases "load", "save", and "merge" can be found as static methods in the RenderDump class. It also handles the lookup of the correct DumpFormat strategy for the given dump format version.
The DumpFormat class provides different implementations for dump formats (including a registry for different dump format versions).


chunky/src/java/se/llbit/chunky/renderer/renderdump/DumpFormat.java
```diff
@@ -0,0 +1,212 @@
...
+import se.llbit.util.TaskTracker;
+
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.SortedMap;
+import java.util.TreeMap;
+
+abstract class DumpFormat {
+  private static final SortedMap<Integer, DumpFormat> DUMP_FORMATS = new TreeMap<>();
```
leMaik, line 29: Why not use a list instead and the index in the list as version?

chunky/src/java/se/llbit/chunky/renderer/renderdump/DumpFormat.java
```diff
@@ -0,0 +1,212 @@
...
+   */
+  protected int getHeaderProgressSize(Scene scene) {
+    return 1;
+  }
+
+  /**
+   * Used in the calculation of the total loading/saving task size
+   */
+  protected int getBodyProgressSize(Scene scene) {
+    return 1;
```
leMaik, line 69: Typically this will depend on the scene width and height, so the overrides could be removed:

```suggestion
    return scene.width * scene.height;
```

leMaik (changes requested):
The refactoring assumes that every format would have a header and a body. That's probably the case but I think it's a strange abstraction.
The common thing among render dumps are: load, save, merge, get spp and color at x/y coordinate (or maybe get the samples array). There should be an interface that reflects this and can be implemented by all octrees.

chunky/src/java/se/llbit/chunky/renderer/renderdump/DumpFormat.java
ShirleyNekoDev: That's a good idea! Don't know what my reason for a map was 😅

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2032,61 +2034,16 @@ private boolean tryLoadDump(RenderContext context, String fileName, TaskTracker
...
-        }
-      }
-    } catch (IOException e) {
+    Log.info("Loading render dump: " + dumpFile);
+    try {
+      RenderDump.load(dumpFile, this, taskTracker);
+    } catch (IOException | IllegalStateException e) {
       // The render dump was possibly corrupt.
-      Log.warn("Failed to load render dump", e);
+      Log.warn("IO exception while loading render dump!", e);
```
leMaik, line 2043: Don't yell at the user, it's not their fault that there was an error! :smile:

ShirleyNekoDev: I copied the warning message from the other parts of the scene - I guess I did it the wrong way around 😁

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2111,9 +2068,10 @@ public void finalizePixel(int x, int y) {
    * @param result the resulting color values are written to this array
    */
   public void postProcessPixel(int x, int y, double[] result) {
-    double r = samples[(y * width + x) * 3 + 0];
-    double g = samples[(y * width + x) * 3 + 1];
-    double b = samples[(y * width + x) * 3 + 2];
+    int index = (y * width + x) * 3;
+    double r = samples[index];
+    double g = samples[index + 1];
+    double b = samples[index + 2];
```
leMaik, line 2074: Why?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2358,45 +2316,14 @@ public void setBufferFinalization(boolean value) {
...
-      // Update render status.
-      spp += dumpSpp;
-      renderTime += dumpTime;
+    Log.info("Merging render dump: " + dumpFile);
+    try {
+      RenderDump.merge(dumpFile, this, taskTracker);
+      postProcessFrame(taskTracker);
     } catch (IOException e) {
-      Log.info("Render dump not loaded");
+      Log.warn("IO exception while merging render dump!", e);
```
leMaik, line 2324: Same here!

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2358,45 +2316,14 @@ public void setBufferFinalization(boolean value) {
...
     }
+    Log.info("Render dump merged: " + dumpFile);
```
leMaik, line 2326: This should be moved before line 2323. Otherwise it will be printed even if merging failed.

chunky/src/java/se/llbit/chunky/renderer/renderdump/DumpFormat.java
ShirleyNekoDev: This is a remnant from the time in refactoring before adding progress report to the FloatingPointCompressor

aTom3333:
I feel that tracking progress at every pixel should have some noticeable overhead. Have you seen any difference in dump time when changing that? (or ideally done some measurements)
Ideally progress should be tracked for every row/column as it was done for the old format, that's just a bit tricky for the new one because it doesn't have a notion of row or column, it works on a linear stream.

ShirleyNekoDev:
> I feel that tracking progress at every pixel should have some noticeable overhead. Have you seen any difference in dump time when changing that? (or ideally done some measurements)

I did a quick check and noticed a big slow down (more than 10x worse).
Theoretically this shouldn't have that much of an impact for JavaFX, but I didn't check the indirection in the TaskTracker and it's listeners.

> Ideally progress should be tracked for every row/column as it was done for the old format, that's just a bit tricky for the new one because it doesn't have a notion of row or column, it works on a linear stream.

A fix for this will come with my next commit by reporting progress (you named it) only once per row again.

ShirleyNekoDev:
Performance is now back for large dumps - thanks for the warning!

Scene size: 2000x2000

## Original code

- Dump format version 1
```
Loading took 266ms
Saving took 260ms

Loading took 265ms
Saving took 242ms
```

## New code

- Dump format version 1
```
Loading took 273ms
Saving took 260ms

Loading took 274ms
Saving took 261ms

Loading took 287ms
Saving took 276ms
```

- Dump format version 0
```
Loading took 12032ms
Saving took 17039ms

Loading took 12213ms
Saving took 16923ms

Loading took 12111ms
Saving took 16887ms
```

- Dump format version 0 (Without sending progress via pixelProgress)
```
Loading took 12010ms
Saving took 17020ms
```

chunky/src/java/se/llbit/chunky/renderer/renderdump/DumpFormat.java
```diff
@@ -0,0 +1,188 @@
...
+  public void load(
+    DataInputStream inputStream,
+    Scene scene,
+    TaskTracker taskTracker
+  ) throws IOException, IllegalStateException {
+    try (TaskTracker.Task task = taskTracker.task(
+      "Loading render dump",
+      scene.width * scene.height
+    )) {
+      readHeader(inputStream, scene);
```
leMaik, line 52: The header is compressed in the old dump format. This looks like it is incompatible with that. There was some logic in place to detect this and un-read the bytes after detecting the version.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2032,61 +2034,16 @@ private boolean tryLoadDump(RenderContext context, String fileName, TaskTracker
       }
       return false;
     }
-    try(PushbackInputStream input = new PushbackInputStream(new FileInputStream(dumpFile), 4)) {
-      byte[] magicNumber = new byte[4];
-      input.read(magicNumber, 0, 4);
-      if(Arrays.equals(magicNumber, DUMP_FORMAT_MAGIC_NUMBER)) {
```
leMaik, line 2038: This logic must not be removed. This PR breaks compatibility with the old dump format.

leMaik (changes requested):

chunky/src/java/se/llbit/chunky/renderer/renderdump/DumpFormat.java
ShirleyNekoDev: Every method in `DumpFormat` expects a stream in the correct format (see Javadocs on class).
The logic for handling the compression and format version logic can be found in `RenderDump`.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
ShirleyNekoDev: This logic for handling the compression and format version logic can be found in `RenderDump`. :)

chunky/src/java/se/llbit/chunky/renderer/renderdump/DumpFormat.java
leMaik: Indeed… I was blind yesterday.

chunky/src/java/se/llbit/chunky/renderer/renderdump/DumpFormat.java
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  protected void readHeader(
+    DataInputStream inputStream,
+    Scene scene
+  ) throws IOException, IllegalStateException {
+    int width = inputStream.readInt();
+    int height = inputStream.readInt();
+    if (width != scene.width || height != scene.height) {
+      throw new IllegalStateException("Scene size does not match dump size");
```
leMaik, line 68: `IllegalArgumentException` probably :thinking:

chunky/src/test/se/llbit/chunky/renderer/renderdump/RenderDumpTests.java
```diff
@@ -0,0 +1,156 @@
...
+    long renderTime
+  ) {
+    Scene scene = new Scene();
+    scene.setCanvasSize(width, height);
+    scene.spp = spp;
+    scene.renderTime = renderTime;
+    return scene;
+  }
+
+  private static final Map<String, String> testDumps = new HashMap<String, String>() {{
```
leMaik, line 79: Why not add them as test resource files and load them from there instead of encoding them as strings?

leMaik (approved):

chunky/src/java/se/llbit/chunky/renderer/renderdump/DumpFormat.java
ShirleyNekoDev: It is the internal state of the dump and scene which do not match, the given argument (the scene) should always be correct.
That's my reasoning behind IllegalStateException

chunky/src/test/se/llbit/chunky/renderer/renderdump/RenderDumpTests.java
ShirleyNekoDev: This can be done in another PR were JUnit is upgraded and we have a resource folder for tests :)

chunky/src/java/se/llbit/chunky/renderer/renderdump/RenderDump.java
```diff
@@ -0,0 +1,124 @@
...
+   * This overwrites ssp, renderTime and samples in the scene.
+   *
+   * @throws IllegalStateException If the width or height of the scene do not match the width or height in the dump.
+   * @throws IOException           If the dump format is unknown or file access fails
+   */
+  public static void load(
+    InputStream inputStream,
+    Scene scene,
+    TaskTracker taskTracker
+  ) throws IOException, IllegalStateException {
```
leMaik, line 56:
```suggestion
   * @throws IllegalArgumentException If the width or height of the scene do not match the width or height in the dump.
   * @throws IOException           If the dump format is unknown or file access fails
   */
  public static void load(
    InputStream inputStream,
    Scene scene,
    TaskTracker taskTracker
  ) throws IOException, IllegalArgumentException {
```

chunky/src/java/se/llbit/chunky/renderer/renderdump/RenderDump.java
```diff
@@ -0,0 +1,124 @@
...
+   * This overwrites ssp, renderTime and samples in the scene.
+   *
+   * @throws IllegalStateException If the width or height of the scene do not match the width or height in the dump.
+   * @throws IOException           If the dump format is unknown or file access fails
+   */
+  public static void merge(
+    InputStream inputStream,
+    Scene scene,
+    TaskTracker taskTracker
+  ) throws IOException, IllegalStateException {
```
leMaik, line 90:
```suggestion
   * @throws IllegalArgumentException If the width or height of the scene do not match the width or height in the dump.
   * @throws IOException           If the dump format is unknown or file access fails
   */
  public static void merge(
    InputStream inputStream,
    Scene scene,
    TaskTracker taskTracker
  ) throws IOException, IllegalArgumentException {
```

PR #810, opened 2021-02-03
More canvas scaling controls

alexhliu:
Small quality of life improvement:
- Added a "Fit to Screen" scaling option that will fit the render/preview to the screen size.
- Made the preview automatically scroll on canvas scale change to keep the render visible.

image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
+++ b/chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
@@ -71,6 +71,8 @@ public class RenderCanvasFx extends ScrollPane implements Repaintable, SceneStat
   private Vector2 target = new Vector2(0, 0);
   private Tooltip tooltip = new Tooltip();
 
+  private boolean fitToScreen = false;
+
   private RenderStatusListener renderListener;
 
   public RenderCanvasFx(se.llbit.chunky.renderer.scene.Scene scene, Renderer renderer) {
@@ -171,9 +173,21 @@ public class RenderCanvasFx extends ScrollPane implements Repaintable, SceneStat
       if (percent == 100) {
         item.setSelected(true);
       }
-      item.setOnAction(e -> updateCanvasScale(percent / 100.0));
+      item.setOnAction(e -> {
+        updateCanvasScale(percent / 100.0);
+        fitToScreen = false;
+      });
       canvasScale.getItems().add(item);
     }
+
+    RadioMenuItem item = new RadioMenuItem("Fit to Screen");
+    item.setToggleGroup(scaleGroup);
+    item.setOnAction(e -> {
+      fitToScreen = true;
+      updateCanvasFit();
+    });
+    canvasScale.getItems().add(item);
+
     contextMenu.getItems().addAll(setTarget, showGuides, canvasScale);
 
     canvas.setOnMouseClicked(event -> {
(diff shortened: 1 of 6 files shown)


StanleyMines:
Great feature! We might be able to make this the default instead of 100%.

Seems to work pretty good when zooming in to fit screen, but only zooms out a few percent (and not the whole way) when needing to zoom out. Clicking fit-to-screen again causes it to zoom out a few more precent, but still not all the way.

I do like how it automatically re-fits when changing scene canvas size, which is a really nice feature! Though it does not currently re-fit on resizing render preview window, which would be nice to have eventually.

Overall great idea and great work!

StanleyMines (changes requested):
Great feature! We might be able to make this the default instead of 100%.

Seems to work pretty good when zooming in to fit screen, but only zooms out a few percent (and not the whole way) when needing to zoom out. Clicking fit-to-screen again causes it to zoom out a few more precent, but still not all the way.

I do like how it automatically re-fits when changing scene canvas size, which is a really nice feature! Though it does not currently re-fit on resizing render preview window, which would be nice to have eventually.

Once zooming out works consistently, I think this would be ready.

Overall great idea and great work!

chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
```diff
@@ -71,6 +72,8 @@
   private Vector2 target = new Vector2(0, 0);
   private Tooltip tooltip = new Tooltip();
 
+  private boolean fitToScreen = false;
```
leMaik, line 75: I'd say we should enable this by default. What do you think?

leMaik (changes requested):
Great first contribution! :tada: 

As @StanleyMines already pointed out, would you mind looking into that bug? If you don't have the time, we could take a look at it. Everything else looks good. :+1:

StanleyMines:
Zoom out seems to work now, although on new scene, the fit-to-screen is marked as selected, however the actual view is still 100%'s view:
Screenshot 2021-02-03 at 18 53 11

StanleyMines (approved):

chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
alexhliu: Fixed bug in 431e63f
Truly set default in 047598f

jackjt8:
Quick note of the performance:

Old scaling methods did not have any performance overhead. Best Fit seems to increase render times by around 4% (short 13 minute test) while preview is active at least for my test scene and system.

alexhliu:
I just tested performance by directly measuring the `updateCanvasFit` method. It seems to take about 130-140us per call and its called once per sample. 4% is also a really small number so I think the difference was probably just due to randomness in background processes etc.
What I measured: 256 sample render, 259 calls to `updateCanvasFit`, spent 34.1ms in `updateCanvasFit` (4 minute 42 second render)

leMaik:
Same here. Performance with scale to fit is slower than without.

StanleyMines:
> Same here. Performance with scale to fit is slower than without. Why do we need to re-fit it with every sample anyway? The image size doesn't change.

I suggested it be run with repaints, as then every time the frame is resized and interacted with it will update (or if some other event such as a scene property is changed, it will updated). The code doesnt seem to be one that should take up very much time, plus for larger renders, its being run less often (compared to number of samples taken). I would check for a frame/window resize, however I didn't see a listener for that kind of event or anything. If you know of one, feel free to move the resize from repaint to there, or let me/Alex know so it can be changed.

Alternatively we could remove the call from repaint, however that would require the user to click to update each time. (The goal was to put it as behind-the-scenes as possible so that the user can set it to `fit` and not have to think about it anymore.)

leMaik:
Please open an issue, I might look into this when I have more time.

Issue #799, opened 2021-01-28
Ray leakage through diagonal walls (precision)


jackjt8:
Rays can leak through diagonal walls. Below I have tested with emitters however this is noticeable with both sun and sky rays as well.

ie

| air   | solid |
|-------|-------|
| solid | air   |

---

**Test setup**

Emitter which is blocked off:
![image](https://user-images.githubusercontent.com/3740917/106177084-fdba6180-618f-11eb-8f2d-a8471bca1438.png)

The missing corner:
![image](https://user-images.githubusercontent.com/3740917/106177125-0ad75080-6190-11eb-9208-6b9a351d90fb.png)

Example render with ESS:None
![image](https://user-images.githubusercontent.com/3740917/106178813-22afd400-6192-11eb-9da1-1d3f448d0bc8.png)

Example render with ESS:All; Additional emitter sampling is causing more leakage per sample.
![image](https://user-images.githubusercontent.com/3740917/106179018-5f7bcb00-6192-11eb-974c-f7325e32173c.png)

---

Test scene:
[TestScene.zip](https://github.com/chunky-dev/chunky/files/5889193/TestScene.zip)

PR #812, opened 2021-02-04
Increase Ray Precision

StanleyMines:
No noticeable speed cost of increased precision, while reducing rays from hitting the "seams" between blocks, or leaking through diagonals. (Not a true fix, but reduces the problem by ~99%.) These seams are really only noticeable if you zoom in to the edge of a block now, and leaked rays are are fewer and would make much less of an impact (1/100 as many as the few that go through now should make them nigh invisible).

(And while I'm at this super short PR, may as well suppress the only warning, which due to how it is currently used, will never cause crash.)

---

Example of seams (zoomed in to the corner of a grass block):
corner-of-grass-blocks

And if offset variable is increased to unreasonably large values, it shows the same artifact on a larger scale (I think this is 0.1 instead of 0.0001):
block-corners-acting-weird

This is also noticeable where the different faces of a block have large differences in color, such as on a normal map in the denoiser. With the current values, a flat wall will have a spattering of pixels along block seams which are colored as a different direction, for when the ray hits gets in the seam and hits a different face of the block, instead of the outward facing face.

---

This also relates to #799 

For the following scene, the following were rendered with the `epsilon` and `offset` values multiplied by 10, 1, and 0.1 respectively, to compare light leakage through the block diagonals. SPS was similar for all three: ~850,000 for ray depth 4. (no visible difference between ray depth 4 and ray depth 100. I tested it.)

DiagonalLightScene_preview

The following are the exact same view, with only the four variables differing (ratio compared to the middle one in parenthesis):

offset: 0.001 (10x); sun intensity: 15 (.2x); sky intensity: 90 (.2x); epsilon: 0.00005 (10x) -> (10 000 spp at ~850 000 sps)
DiagonalLightScene_E-3_10k

offset: 0.0001 (default); sun intensity: 75; sky intensity: 450; epsilon: 0.000005 (default) -> (60 000 spp at 829 900 sps)
DiagonalLightScene_E-4_60k

offset: 0.00001 (0.1x); sun intensity: 150 (2x); sky intensity: 900 (2x); epsilon: 0.0000005 (0.1x) -> (700 000 spp at 923 100 sps)
DiagonalLightScene_E-5_700k

As you can see, the scene looks similar in brightness and noise levels, while light intensity needed to be greatly increased (5x then ~2.5x) to maintain similar lighting, while also requiring far far more spp to become clear over the noise (6x to 12x), per decimal added to epsilon and offset, while not contributing to sps costs.

---

If someone is not lighting a scene with just light leaking through corners, they would have no reason to increase brightness to absurd levels for the little light that leaks through to light or affect the scene in any meaningful way. With these more precise values, I would expect to see about a 100 to 1 reduction in fireflies caused by this (rays need to fit through a gap that is 100 times smaller), as well as a 100 to 1 reduction in the size of the "seams" between blocks. And from my tests, both listed here and not, there seems to be no noticeable performance impact with this change.

### DIFF
--- a/chunky/src/java/se/llbit/math/Ray.java
+++ b/chunky/src/java/se/llbit/math/Ray.java
@@ -33,9 +33,9 @@ import java.util.Random;
  */
 public class Ray {
 
-  public static final double EPSILON = 0.000005;
+  public static final double EPSILON = 0.00000005;
 
-  public static final double OFFSET = 0.0001;
+  public static final double OFFSET = 0.000001;
 
   /**
    * Ray direction.
--- a/chunky/src/java/se/llbit/util/RingBuffer.java
+++ b/chunky/src/java/se/llbit/util/RingBuffer.java
@@ -54,6 +54,8 @@ public class RingBuffer<T> {
     if (size == 0) {
       throw new NoSuchElementException("Buffer is empty.");
     }
+
+    @SuppressWarnings("unchecked")
     T result = (T) data[remove];
     remove = (remove + 1) % capacity;
     size -= 1;


leMaik:
This looks good to me, but are there any cases where _epsilon_ or _offset_ could be too small now and cause errors or infinite loops?

(If there are, we will find out when CC is updated…)

leMaik (approved):

StanleyMines:
> This looks good to me, but are there any cases where _epsilon_ or _offset_ could be too small now and cause errors or infinite loops?
> 
> (If there are, we will find out when CC is updated…)

Oooo. Thats a good point. I only tested it on scenes of reasonable size, but if scenes are huge, a coordinate + offset might result in the exact same value... Hm...

With java's mantissa size, it looks like epsilon might get lost for scenes on the order of 10^8 blocks in radius (octree would have about a height of 26 to 27 (2^26+2^-25 != 2^26, but 2^28+2^-25 == 2^28 in java (52 bits of mantissa in doubles))).

leMaik:
@StanleyMines That would be about 1.5 * 10^14 chunks. So I guess this is save to merge?

StanleyMines:
Probably good for now, but in the future it could probably be set per scene (based on octree size or something), if it doesn't cost efficiency. Should be good for now though :)

PR #813, opened 2021-02-05
Update defaults for Iron-like blocks

jackjt8:
Defaults
materials_default-64

Updated with metalness and roughness
materials-64


leMaik:
Nice! :+1: Are iron chains already in the list? They should probably be like iron bars…

jackjt8:
> Nice! 👍 Are iron chains already in the list? They should probably be like iron bars…

I forgot chains while I was looking through the creative inventory. It's not that noticeable at a distance.
image

PR #816, opened 2021-02-08
Nishita sky and sky caching system

alexhliu:
Main changes:
- Implemented a Nishita sky renderer based on [this](https://www.scratchapixel.com/lessons/procedural-generation-virtual-worlds/simulating-sky/simulating-colors-of-the-sky).
- Added a selector to switch between the default Preetham sky and the new Nishita sky.
- Added a new entry in the scene description json that stores the sky renderer.
- Added an on-the-fly sky caching system which automatically stores sky colors in a cache as they are requested. Sky texture cache is by `1024 x 1024` by default but can be changed through a slider in the `Advanced` tab (I don't see any interpolation artifacts at the default resolution so this is probably not necessary).

---

Example Nishita sky scene. Rendered at ~470k samples per second. Same scene on latest snapshot renders at ~465k samples per second. (Note: With a minimal cache resolution of 64, I got 500k samples per second with no visible quality loss.)
image

Cache resolution: 1024
image

Cache resolution: 64
image

My branch Preetham
image

Master branch Preetham
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
@@ -1,3 +1,19 @@
+/* Copyright (c) 2021 Chunky contributors
+ *
+ * This file is part of Chunky.
+ *
+ * Chunky is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Chunky is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with Chunky.  If not, see <http://www.gnu.org/licenses/>.
+ */
 package se.llbit.chunky.renderer.scene;
 
 import se.llbit.math.Ray;
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
@@ -24,14 +24,9 @@ import se.llbit.math.Vector3;
  */
 public interface SimulatedSky {
   /**
-   * Update the sun
+   * Update the sun if necessary. Returns true if the sun was updated (and cache needs to be purged).
    */
-  void updateSun(Sun sun);
-
-  /**
-   * Check if the sky needs an update with a new sun.
-   */
-  boolean needUpdate(Sun sun);
+  boolean updateSun(Sun sun);
 
   /**
    * Calculate the sky color for a given ray.
@@ -46,5 +41,5 @@ public interface SimulatedSky {
   /**
    * Get the sky renderer tooltip.
    */
-  String getTooltip();
+  String getDescription();
 }
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
@@ -6,11 +6,16 @@ import se.llbit.math.Vector3;
 /**
  * Interface for simulated skies.
  */
-public interface SkySimulated {
+public interface SimulatedSky {
+  /**
+   * Update the sun
+   */
+  void updateSun(Sun sun);
+
   /**
    * Calculate the sky color for a given ray.
    */
-  Vector3 calcIncidentLight(Ray ray, Sun sun, double horizonOffset);
+  Vector3 calcIncidentLight(Ray ray, double horizonOffset);
 
   /**
    * Get the friendly name.
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
@@ -36,7 +36,7 @@ public interface SimulatedSky {
   /**
    * Calculate the sky color for a given ray.
    */
-  Vector3 calcIncidentLight(Ray ray, double horizonOffset);
+  Vector3 calcIncidentLight(Ray ray);
 
   /**
    * Get the friendly name.
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
@@ -13,6 +13,11 @@ public interface SimulatedSky {
   void updateSun(Sun sun);
 
   /**
+   * Check if the sky needs an update with a new sun.
+   */
+  boolean needUpdate(Sun sun);
+
+  /**
    * Calculate the sky color for a given ray.
    */
   Vector3 calcIncidentLight(Ray ray, double horizonOffset);
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
@@ -26,7 +26,7 @@ public interface SimulatedSky {
   /**
    * Update the sun if necessary. Returns true if the sun was updated (and cache needs to be purged).
    */
-  boolean updateSun(Sun sun);
+  boolean updateSun(Sun sun, double horizonOffset);
 
   /**
    * Calculate the sky color for a given ray.
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
@@ -174,9 +174,12 @@ public class Sky implements JsonSerializable {
   /** Simulated sky mode. */
   private int simulatedSkyMode = 0;
 
+  private SkyBaker skyBaker;
+
   public Sky(Scene sceneDescription) {
     this.scene = sceneDescription;
     makeDefaultGradient(gradient);
+    skyBaker = new SkyBaker(this);
   }
 
   /**
@@ -264,7 +267,8 @@ public class Sky implements JsonSerializable {
         break;
       }
       case SIMULATED: {
-        Vector3 color = skies.get(simulatedSkyMode).calcIncidentLight(ray, horizonOffset);
+        Vector3 color = skyBaker.calcIncidentLight(ray, horizonOffset);
+//        Vector3 color = skies.get(simulatedSkyMode).calcIncidentLight(ray, horizonOffset);
         ray.color.set(color.x, color.y, color.z, 1);
         break;
       }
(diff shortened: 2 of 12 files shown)


alexhliu:
Test seems to fail due to the overhead of allocating the texture array and filling it. Test passes after 1 minute and 10 seconds on my computer (timeout is 1 minute). I deferred array allocation and initialization to the first call which seems to let the test pass without impacting performance.

chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
```diff
@@ -0,0 +1,50 @@
...
+public interface SimulatedSky {
+  /**
+   * Update the sun
+   */
+  void updateSun(Sun sun);
+
+  /**
+   * Check if the sky needs an update with a new sun.
+   */
+  boolean needUpdate(Sun sun);
```
leMaik, line 34: The `updateSun` method could do this check and just do nothing if it doesn't require an update with the new sun.

chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Get the friendly name.
+   */
+  String getName();
+
+  /**
+   * Get the sky renderer tooltip.
+   */
+  String getTooltip();
```
leMaik, line 49: (not important)

```suggestion
  String getDescription();
```

chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
```diff
@@ -162,9 +162,24 @@ public static SkyMode get(String name) {
...
+  /** Simulated skies. */
+  public final static List<SimulatedSky> skies = new ArrayList<>();
+
+  static {
+    skies.add(new PreethamSky());
+    skies.add(new NishitaSky());
+  }
+
+  /** Simulated sky mode. */
+  private int simulatedSkyMode = 0;
```
leMaik, line 174: Plugins could add new sky modes. Should we use the classname to identify them?

chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
```diff
@@ -202,7 +217,6 @@ public void loadSkymap(String fileName) {
    * Set the sky equal to other sky.
    */
   public void set(Sky other) {
-    horizonOffset = other.horizonOffset;
```
leMaik, line 205: Did this only not work for simulated sky? We may want to keep it if it works for skymaps (maybe disable it in the GUI if simulated sky is selected).

chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
```diff
@@ -539,6 +587,10 @@ public SkyMode getSkyMode() {
         sky.add("skybox", array);
         break;
       }
+      case SIMULATED: {
+        sky.add("simulatedSky", simulatedSkyMode);
```
leMaik, line 591: Should the cache resolution also be read from / written to the scene description file?

chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
```diff
@@ -583,6 +634,10 @@ public void importFromJson(JsonObject json) {
         }
         break;
       }
+      case SIMULATED: {
+        setSimulatedSkyMode(json.get("simulatedSky").intValue(simulatedSkyMode));
```
leMaik, line 638: see above

chunky/src/java/se/llbit/chunky/renderer/scene/SkyCache.java
```diff
@@ -0,0 +1,162 @@
...
+  private double interp1D(double x, double x0, double x1, double y0, double y1) {
+    return y0 + (x - x0)*(y1-y0)/(x1-x0);
+  }
+
+  // Calculate the bilinearly interpolated value from the cache.
+  private Vector3 getColorInterpolated(double normX, double normY) {
+    double x = normX * skyResolution;
+    double y = normY * skyResolution;
+    int floorX = (int) QuickMath.clamp(x, 0, skyResolution-1);
+    int floorY = (int) QuickMath.clamp(y, 0, skyResolution-1);
```
leMaik, line 130: Is `QuickMath` actually faster for clamping than `Math`?

chunky/src/java/se/llbit/chunky/renderer/scene/Sun.java
```diff
@@ -224,6 +196,11 @@ private void initSun() {
     emittance.set(color);
     emittance.scale(FastMath.pow(intensity, Scene.DEFAULT_GAMMA));
 
+    Sky sky = ((Scene) scene).sky();
+    if (sky.getSkyMode() == Sky.SkyMode.SIMULATED && sky.getSimulatedSky().needUpdate(this)) {
```
leMaik, line 200: As suggested above, you could just always call `sky.updateSimulatedSky(this)` if the implementation checks internally if an update is actually needed. That would simplify the caller side (and get rid of `needUpdate`)

chunky/src/java/se/llbit/chunky/ui/render/AdvancedTab.java
```diff
@@ -113,6 +116,14 @@ public void initialize(URL location, ResourceBundle resources) {
     fastFog.setTooltip(new Tooltip("Enable faster fog rendering algorithm."));
     fastFog.selectedProperty()
             .addListener((observable, oldValue, newValue) -> scene.setFastFog(newValue));
+    cacheResolution.setName("Sky cache resolution");
+    cacheResolution.setTooltip("Resolution of sky cache.");
```
leMaik, line 120:
```suggestion
    cacheResolution.setTooltip("Resolution of the sky cache. Lower values will use less memory and improve performance but can cause sky artifacts.");
```

chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
```diff
@@ -81,12 +86,33 @@ public SkyTab() throws IOException {
   }
 
   @Override public void initialize(URL location, ResourceBundle resources) {
-    simulatedSettings.getChildren().add(horizonOffset);
-    horizonOffset.setName("Horizon offset");
-    horizonOffset.setTooltip("Moves the simulated horizon.");
```
leMaik, line 86: "the simulated horizon"… so this didn't affect other skies at all and didn't work for simulated sky?

leMaik (changes requested):
The new sky model is awesome! :tada: 

Just a few minor suggestions and questions.

chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
jackjt8: It does do something with the horizon but it's somehow bugged for Preetham and not implemented yet with Nishita. This is more of a temporary fix until both can get fixed/implemented.

Non-zero offset:
image

chunky/src/java/se/llbit/chunky/renderer/scene/SkyCache.java
```diff
@@ -0,0 +1,162 @@
...
+import se.llbit.math.Vector3;
+
+import static java.lang.Math.PI;
+
+public class SkyCache {
+  // Sky texture array, HSV colors
+  private double[][][] skyTexture;
+
+  // Default resolution is 1024x1024. Should be more than enough for any simulated sky.
+  private int skyResolution = 1024;
```
jackjt8, line 32: 8 
image

16
image

32
image

64
image

128
image

While a cache size of 32 is too low and 64 appears to be just right, I think going for 128 as the safer option would be best. Performance with the current default of 1024 should give us the same performance as previous versions Preetham.

chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
jackjt8: So, it the offset did certainly work for Preetham and for whatever reason the copy-paste job of the Preetham code for this PR broke it... Only the situated sky modes use the offset.

0
image

1
image

chunky/src/java/se/llbit/chunky/renderer/scene/SkyCache.java
alexhliu: `Math` does not seem to have an explicit clamping method so you must call both `Math.min` and `Math.max`. `QuickMath` does this with some ternary operators. The difference in speed is probably negligible but `clamp` is more clear and theoretically a bit faster.

chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
alexhliu: I believe it did work with both `Preetham` and `Nishita`, however it had the weird circle in the middle if I implemented it on the sky cache. I would have to reset the sky cache and re-bake the sky every time you changed horizon offset. I can try implementing that and testing it.

chunky/src/java/se/llbit/chunky/renderer/scene/SkyCache.java
alexhliu: A cache size of 128 seems good. I can change the default to that. It should improve performance a lot, especially when it has to rebuild the cache on Nishita sky.

chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
alexhliu: Implemented in b3fb88d

chunky/src/java/se/llbit/chunky/renderer/scene/SimulatedSky.java
alexhliu: Implemented in b3fb88d

chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
alexhliu: I switched it to hold a `SimulatedSky` reference instead which should be more robust to additional sky modes (b3fb88d)

chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
alexhliu: Implemented in b3fb88d

chunky/src/java/se/llbit/chunky/ui/render/AdvancedTab.java
alexhliu: Changed in b3fb88d

chunky/src/java/se/llbit/chunky/renderer/scene/SkyCache.java
alexhliu: Changed cache size to 128 in b3fb88d

chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
alexhliu: Re-added in 727b301. Changing this requires resetting the sky cache.

chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
alexhliu: I readded it in 727b301. Changing the horizon offset requires resetting the sky cache. It seems to also cause issues in Nishita sky with offset > 0.25 or so.

chunky/src/java/se/llbit/chunky/renderer/scene/SkyCache.java
leMaik: Whoops… You're right there is no `Math.clamp`, my bad.

leMaik (approved):

PR #817, opened 2021-02-10
add support for 21w06a tall chunks

NotStirred:
initial support for tall chunks, in my testing it works as expected

Currently I'm setting `GeneralTab#yMin` and `GeneralTab#yMax` bounds to `-64` to `320` for all worlds. If you want this to change per-world, scene would probably need the `GeneralTab` instance passed to it on `Scene#loadChunks` or something

Also a sneaky bug fix in `Chunk@290`, preventing negative section positions

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/render/GeneralTab.java
+++ b/chunky/src/java/se/llbit/chunky/ui/render/GeneralTab.java
@@ -43,6 +43,7 @@ import se.llbit.chunky.ui.RenderCanvasFx;
 import se.llbit.chunky.ui.RenderControlsFxController;
 import se.llbit.chunky.world.EmptyWorld;
 import se.llbit.chunky.world.Icon;
+import se.llbit.chunky.world.World;
 import se.llbit.fxutil.Dialogs;
 import se.llbit.json.JsonObject;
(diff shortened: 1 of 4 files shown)


chunky/src/java/se/llbit/chunky/ui/render/GeneralTab.java
```diff
@@ -205,14 +205,14 @@ public GeneralTab() throws IOException {
     canvasSize.setEditable(true);
     canvasSize.getItems().addAll("400x400", "1024x768", "960x540", "1920x1080");
     canvasSize.valueProperty().addListener(canvasSizeListener);
-    yMax.setRange(0, 256);
+    yMax.setRange(-64, 320);
```
leMaik, line 208: You could access the scene and to `this.scene.world.getVersionId() >= 2694` here.

NotStirred: ah! will do

leMaik: oh you're right, it doesn't. Maybe lin 300 is a better place to add this?

PR #818, opened 2021-02-11
Added accessors/made things public

alexhliu:
This should cover most of the things that are required to hack on a new renderer through a plugin.
 - Added a setter to `Chunky.java` for changing the `RendererFactory`
 - Added accessors for both the world and water octree in `Scene.java`
 - Added a getter for the block palette list in `BlockPalette.java`
 - Added a getter for the octree implementation and made `treeData` public in `PackedOctree.java` and `BigPackedOctree.java`
 - Made the classes in `BVH.java` public and added an accessor for the root node

### DIFF
--- a/chunky/src/java/se/llbit/math/BVH.java
+++ b/chunky/src/java/se/llbit/math/BVH.java
@@ -17,7 +17,7 @@
  */
 package se.llbit.math;
 
-import se.llbit.chunky.plugin.API;
+import se.llbit.chunky.plugin.PluginApi;
 import se.llbit.math.primitive.MutableAABB;
 import se.llbit.math.primitive.Primitive;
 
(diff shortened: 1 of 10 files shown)


chunky/src/java/se/llbit/math/BVH.java
```diff
@@ -43,10 +45,10 @@
     SAH_MA,
   }
 
-
-  static abstract class Node {
-    protected final AABB bb;
-    protected final Primitive[] primitives;
+  @PluginApi
```
leMaik, line 48: Hm… Yes this will break plugins when changed but I wouldn't consider this to be part of the plugin api :thinking:

alexhliu: I replaced these with a javadoc saying that they are public for some plugins but that stability is not guaranteed.

PR #822, opened 2021-02-18
Sky cache bug fixes.

alexhliu:
More aggressive sky cache synchronization, fixed black regions in the sky, fixed Nishita horizon offset. Sun is now drawn if `Draw Sun` is enabled but `Enable Sunlight` is disabled.


leMaik (approved):
Nice :+1:

PR #827, opened 2021-02-22
Fixed freeze on load and added common thread pool.

### LINKED ISSUE #415: Chunky stops responding when loading chunks
When I load selected chunks, chunky stops responding, but still loads the chunks. I am unsure why it does that, and I don't know if you get the same error I do. The selection was about 8000 chunks and I know it was still loading the chunks because of the high CPU usage and the increasing amount of RAM it was taking. And another question, I've been wondering this for a while now, but why does chunky take up so much memory? For a selection of 8000 chunks, it's taking ~7Gb. I don't know anything on this subject, but wouldn't it make sense if the consumed memory were around the same size as the map size? Ex: if the map were 1 Gb shouldn't it allocate 1 Gb of memory? Thanks!

alexhliu:
- Added a common thread pool that is accessible through `Chunky.getCommonThreads()`
- Modified existing methods to use this.
- Modified `ChunkMap.drawViewBounds` to wait for scene synchronization on a separate thread (if it is already waiting, subsequent requests do not create new jobs and instead are ignored).
- This fixes #415

### DIFF
--- a/chunky/src/java/se/llbit/chunky/main/Chunky.java
+++ b/chunky/src/java/se/llbit/chunky/main/Chunky.java
@@ -213,9 +213,10 @@ public class Chunky {
 
     int exitCode = 0;
     if (cmdline.mode != CommandLineOptions.Mode.NOTHING) {
+      commonThreads = new ForkJoinPool(PersistentSettings.getNumThreads());
+
       Chunky chunky = new Chunky(cmdline.options);
       chunky.headless = cmdline.mode == Mode.HEADLESS_RENDER || cmdline.mode == Mode.SNAPSHOT;
-      commonThreads = new ForkJoinPool(PersistentSettings.getNumThreads());
       chunky.loadPlugins();
 
       try {
@@ -330,6 +331,9 @@ public class Chunky {
    * Get the common thread pool.
    */
   public static ForkJoinPool getCommonThreads() {
+    if (commonThreads == null) {
+      commonThreads = new ForkJoinPool(PersistentSettings.getNumThreads());
+    }
     return commonThreads;
   }
 
--- a/chunky/src/java/se/llbit/chunky/main/Chunky.java
+++ b/chunky/src/java/se/llbit/chunky/main/Chunky.java
@@ -62,6 +62,7 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.ForkJoinPool;
 
 /**
  * Chunky is a Minecraft mapping and rendering tool created byJesper Öqvist (jesper@llbit.se).
@@ -96,6 +97,8 @@ public class Chunky {
   private RenderControlsTabTransformer renderControlsTabTransformer = tabs -> tabs;
   private TabTransformer mainTabTransformer = tabs -> tabs;
   private boolean headless = false;
+
+  private static ForkJoinPool commonThreads;
 
   /**
    * @return The title of the main window. Includes the current version string.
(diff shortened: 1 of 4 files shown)


alexhliu:
Forgot to check the tests 🤦‍♂️. Should be fixed now by create a new common thread pool if it is null when requested.

jackjt8 (commented):
Java option `-Dchunky.mapLoaderThreads` can be removed given the UI map loader is now using the common pool.

jackjt8:
Confirmed the freezing has been resolved however a few issues general concerns are noted:

Increased CPU and GPU usage - It seems that this new solution has raised CPU utilisation by upto 5% or so; still bottlenecking on the single threaded loading, but more cores are in use to handle the UI draw. I'm seeing about 20% GPU usage from the OpenJDK Chunky 2.3.0 process which was not there before. I figure the rate that we are loading new chunks currently already is leading to pointless UI updates which may only worsen if we get mutli-threaded loading of any nature in the future. [See video](https://youtu.be/4T3AOksP8DI)

Minor Bug - [Camera icon within Map View blinds on update](https://youtu.be/V-7FYl1Ulg0)

leMaik:
> I figure the rate that we are loading new chunks currently already is leading to pointless UI updates which may only worsen if we get mutli-threaded loading of any nature in the future.

Then we should find out why the UI gets pointlessly updated and fix that.

jackjt8:
> Then we should find out why the UI gets pointlessly updated and fix that.

I'm not sure if it's pointless redraw but but the GPU usage seems to be a bit high for what it is doing. It almost seems like we are getting a new update each time a chunk is loaded and given we are loading at least 100-150 chunks per second. I figure if it is vsync'ed we could still limit the number of times the UI updates especially without user interaction.

alexhliu:
> I'm seeing about 20% GPU usage from the OpenJDK Chunky 2.3.0 process which was not there before.

I just tested this and I don't think this is new to this pull request. On both this and the master branch launched through Intellij, I get about 10% GPU. When I launch the nightly snapshot through the launcher I don't get any GPU usage.

> ...this new solution has raised CPU utilisation by upto 5% or so; still bottlenecking on the single threaded loading, but more cores are in use to handle the UI draw.

This shouldn't add more cores to use in UI draw but it might allow the UI draw to run faster since this step is run later (and should return instantly). That may also explain the extra CPU usage since the UI draw no longer needs to wait for scene synchronization.

leMaik:
@alexhliu Can we throttle the gui updates to only happen every so often at most? Can be done in another pr though, this one is fine

jackjt8:
All tests are performed using a built copy via the launcher. Past nightly builds do not suffer from this issue during loading as the UI doesn't update. Please ensure that you remove the previous default `-Dprism.order=sw` from the launcher else JavaFX will be rendering in software and not hardware (if supported). I wonder if the updates are already tied to the screen refresh rate which would make sense. I am running a 144Hz monitor so if alexhliu is running anything less it might explain the difference we are seeing.

My main concern here is just that using 20% of the GPU just to update a counter/progress bar seems pointless when the majority of users will not be interacting with Chunky at this point or rather that we don't need to update these elements so often. But yes this could be resolved later.

alexhliu:
Ah it was the `-Dprism.order=sw` that was causing my launcher not to use GPU. I don't think the updates are tied to screen refresh rate since I was getting about 100 refreshes per second and I only have a 60hz monitor. It shouldn't be too difficult to rate-limit the refreshes. I'll look into it.

leMaik:
@alexhliu Looking forward to the rate limiting. :+1: I'll merge this PR as is, the fix is good, the rate limiting is just the :cherries: on top.

PR #831, opened 2021-02-23
Ratelimit for javafx refresh when loading chunks

alexhliu:
Currently only methods that are associated with loading chunks are ratelimited. I felt like 3 fps was decent just to show progress without giving the impression that Chunky has crashed, however larger values shouldn't have too big of a performance impact.

Also while I was here I cleaned up some code from my previous pull to just use a `volatile boolean` instead of an `AtomicBoolean`.


leMaik:
Thank you! :clap:

PR #837, opened 2021-02-27
Async chunk load & decomp

NotStirred:
This alternates between two `ChunkData` objects, decompressing the next while the current one is being added to the octree

### DIFF
--- a/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
+++ b/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
@@ -27,12 +27,10 @@ import se.llbit.nbt.Tag;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 
 /**
@@ -42,6 +40,12 @@ import java.util.function.Consumer;
  * The numerical IDs are used to efficiently store the blocks in the octree.
  *
  * This class also manages material properties.
+ *
+ * <p>Before <code>{@link BlockPalette#finalise()}</code> is called, <code>{@link BlockPalette}</code> is thread safe
+ * for N writer and N reader threads. It locks on <code>{@link BlockPalette#put(BlockSpec)}</code>, and has concurrent
+ * data structures to achieve this.</p>
+ *
+ * After <code>{@link BlockPalette#finalise()}</code> is called, it's should be treated as entirely unsafe for > 1 thread
  */
 public class BlockPalette {
   private static final int BLOCK_PALETTE_VERSION = 4;
@@ -55,6 +59,8 @@ public class BlockPalette {
 
   private final Map<BlockSpec, Integer> blockMap;
   private List<Block> palette;
+
+  private ReentrantLock lock = new ReentrantLock();
 
   public BlockPalette(Map<BlockSpec, Integer> initialMap, List<Block> initialList) {
     this.blockMap = initialMap;
@@ -77,8 +83,18 @@ public class BlockPalette {
     this(new ConcurrentHashMap<>(), new CopyOnWriteArrayList<>());
   }
 
+  /**
+   * should be called when no threads are acting on the palette
+   *
+   * replaces the lock with one that doesn't lock
+   * switches the palette list for a non-concurrent one
+   */
   public void finalise() {
     palette = new ArrayList<>(palette);
+    lock = new ReentrantLock() {
+      @Override public void lock() { }
+      @Override public void unlock() { }
+    };
   }
 
   /**
@@ -91,17 +107,26 @@ public class BlockPalette {
     return put(new BlockSpec(tag));
   }
 
+  /**
+   * locks to avoid race conditions between writer threads
+   */
   public int put(BlockSpec spec) {
-    Integer id = blockMap.get(spec);
-    if (id != null) {
+    final ReentrantLock lock = this.lock;
+    lock.lock();
+    try {
+      Integer id = blockMap.get(spec);
+      if (id != null) {
+        return id;
+      }
+      id = palette.size();
+      blockMap.put(spec, id);
+      Block block = spec.toBlock();
+      applyMaterial(block);
+      palette.add(block);
       return id;
+    } finally {
+      lock.unlock();
     }
-    id = palette.size();
-    blockMap.put(spec, id);
-    Block block = spec.toBlock();
-    applyMaterial(block);
-    palette.add(block);
-    return id;
   }
 
   public Block get(int id) {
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -837,7 +848,39 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
...
+            chunkData = chunkData1;
+            nextChunkData = chunkData2;
+          } else {
+            chunkData = chunkData2;
+            nextChunkData = chunkData1;
+          }
+          usingFirstChunkData = !usingFirstChunkData;
+
+          if (i + 1 < chunkPositions.length) { //if has next request next
+            int finalI = i;
```
leMaik, line 877:
```suggestion
            final int finalI = i;
```

Just to be sure/make the intent more clear? :thinking:

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -817,17 +814,31 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
 
     Heightmap biomeIdMap = new Heightmap();
 
-    ChunkData chunkData;
+    ChunkData chunkData1;
```
leMaik, line 817:
```suggestion
    // To speed up chunk loading, we "pipeline" the loading process by loading one chunk while adding the previously loaded chunk to the octree.
    // There seems to be no significant performance benefit when trying to load more chunks at once here.
    ChunkData chunkData1;
```

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -817,17 +814,31 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
...
+    ChunkData chunkData2;
     if (isTallWorld) { //snapshot 21w06a, treat as -64 - 320
-      chunkData = new GenericChunkData();
+      chunkData1 = new GenericChunkData(); // chunk loading will switch between these two, using one asynchronously to load the data
+      chunkData2 = new GenericChunkData(); // while the other is used to add to the octree
```
leMaik, line 821: Oh, this comment already explains what's going on, didn't see it.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -837,7 +848,39 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
 
         loadedChunks.add(cp);
 
-        world.getChunk(cp).getChunkData(chunkData, palette);
+        try {
+          nextChunkDataTask.get();
+        } catch(InterruptedException ignored) { // If except, load the chunk synchronously
+          if(usingFirstChunkData) {
+            world.getChunk(chunkPositions[i]).getChunkData(chunkData1, palette);
```
leMaik, line 855: Is the `BlockPalette` thread-safe, i.e. are we allowed to get blocks from the hash maps while other threads add new blocks to it?

leMaik (commented):
Looking good, nice speedup. Just some concerns regarding thread safety

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
NotStirred: definitely not, I'll fix this today

NotStirred:
I will add the requested `finalI` change along with any others necessary in one go

Regarding cc51506:
I tested many options and found that `ConcurrentHashMap` and `CopyOnWriteArrayList` offered the best performance for the situation `~30%` increase, where locking externally offered a `~%5` decrease

aTom3333:
(I can't comment on some code you didn't change so I'll make a standalone comment and refer to the code by line number).
Even with this change the code is not safe. Sure, independent accesses to the list or to the hash map are now thread safe but coherency between the two isn't enforced.In particular the put method still has a race condition.
Imagine the following scenario: 2 threads want to call this function, thread 1 wants to put `minecraft:dirt` and thread 2 wants to put `minecraft:andesite`. Prior to that neither of those block has already been put into the `BlockPalette` and say the palette already has 10 elements (so `palette.size() == 10`).
Thread 1 runs first and enters the function, block is not already in the hash map so it doesn't exit early, it gets the size of the palette as it will be the id of `minecraft:dirt`, this id is 10 (line 99). The correspondence `minecraft:dirt -> 10` is added in a thread safe manner to the hash map (line 100). Then the thread gets interrupted after that but before adding `minecraft:dirt` to the list (line 103).
Then thread 2 runs and enters the function, not early exit either so the size of the list is queried to be the id of `minecraft:andesite`, this id is 10 (line 99 again)(you can already see the problem). The correspondence `minecraft:andesite -> 10` is added to the hash map (line 100), still in a thread safe manner. Then `minecraft:andesite` is added at the back of the list, creating the correspondence `10 -> minecraft:andesite`, again in a thread safe manner. Thread 2 then does somthing else and eventually thread 1 runs again.
Thread 1 then wants to add its `minecraft:dirt` to the list (line 103). As the list is thread safe, the element is correctly added after the one added in thread 2, creating the correspondence `11 -> minecraft:dirt`.
And here we are, the `BlockPalette` is broken mess, now when a dirt block is read from a chunk, we will store id 10 in the octree (due to the `minecraft:dirt -> 10` correspondence) and when rendering those block, they will be rendered as andesite (due to the `10 -> minecraft:andesite` correspondence).
The 3 operations (querying the size, adding to the hash map, adding to the list) must be done atomically, that's why I advized manual locking on discord.
That being said, I think this is the only method that presents such a race condition, and only in the case of multiple threads calling `put` (ie I think that one thread calling `put` and one or more threads calling `get` is safe) so you only need to lock in this function (if you get thread safe variant of the collection, if you don't you would need to lock in more places)

NotStirred:
I see I misunderstood what you meant in your discord message, thanks for your help

NotStirred:
As discussed at length [on discord](https://discord.com/channels/541221265512464394/545374333883777037/817089641836314655) I added the lock to `BlockPalette#put`, and put in more comments explaining what's happening.
It also no longer uses the common thread pool [so we don't break it](https://discord.com/channels/541221265512464394/545374333883777037/817105240641241148).

Before `BlockPalette#finalise` is called, it is thread safe for `N` writers/readers, afterwards it is not threadsafe at all
(if you can think of a better name for that please do let me know)

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
```diff
@@ -72,7 +80,22 @@ public BlockPalette(Map<BlockSpec, Integer> initialMap, List<Block> initialList)
...
-    this(new HashMap<>(), new ArrayList<>());
+    this(new ConcurrentHashMap<>(), new CopyOnWriteArrayList<>());
+  }
+
+  /**
+   * should be called when no threads are acting on the palette
+   *
+   * replaces the lock with one that doesn't lock
+   * switches the palette list for a non-concurrent one
+   * This is done to not limit render performance once async chunk loading is done
```
leMaik, line 91:
```suggestion
   * This method should be called when no threads are acting on the palette anymore.
   *
   * It replaces the lock with one that does nothing and switches the palette list for a non-concurrent one.
   * This is done to not limit render performance once async chunk loading is done.
```

leMaik (approved):

PR #841, opened 2021-03-05
Fixed Preetham sky rendering bug and sky not being set properly when loading a scene.

### LINKED ISSUE #840: Preetham Sky breaks if Sun altitude is below the horizon
Prior to #816 the sun's altitude was internally limited to 0-90 degrees (ignoring the UI). The Preetham sky breaks if the suns altitude is below 0 degrees (sky eventually turns yellow). The suns altitude should be restricted to 0-90 degrees if using Preetham.

alexhliu:
This fixes #840 by clamping the sky's internal sun position to be always above the horizon and follow the sun properly when above the horizon.

The `SkyCache` is now properly set when loading a scene (previously it would always be in `PreethamSky` until the `Sky&fog` menu is opened).


leMaik:
Thank you for fixing this ✨

PR #847, opened 2021-03-13
Minor optimisations 2

NotStirred:
Some random optimisations I saw looking through the codebase

In my testing outside of the chunky environment (in a similar use-case), there is so much noise in results inside chunky that it's probably unreasonable to draw any conclusions from such small tweaks
The `Pair` implementation is ~9% faster
The `HashMap` implementation is ~7% faster

### DIFF
--- a/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
+++ b/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
@@ -58,7 +58,7 @@ public class WorldMapLoader implements ChunkTopographyListener, ChunkViewListene
     RegionChangeWatcher regionWatcher = new RegionChangeWatcher(this, mapView);
 
     // Start worker threads.
-    RegionParser[] regionParsers = new RegionParser[Integer.parseInt(System.getProperty("chunky.mapLoaderThreads", "3"))];
+    RegionParser[] regionParsers = new RegionParser[Integer.parseInt(System.getProperty("chunky.mapLoaderThreads", "12"))];
     for (int i = 0; i < regionParsers.length; ++i) {
       regionParsers[i] = new RegionParser(this, regionQueue, mapView);
       regionParsers[i].start();
--- a/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
+++ b/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
@@ -58,7 +58,7 @@ public class WorldMapLoader implements ChunkTopographyListener, ChunkViewListene
     RegionChangeWatcher regionWatcher = new RegionChangeWatcher(this, mapView);
 
     // Start worker threads.
-    RegionParser[] regionParsers = new RegionParser[Integer.parseInt(System.getProperty("chunky.mapLoaderThreads", "12"))];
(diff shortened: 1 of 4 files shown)


NotStirred:
I should mention I changed `OctreeImplementation#getWithLevel` signature, so this may cause some issues with plugins that overwrite that

NotStirred:
removed commit that changed behaviour

aTom3333:
A while back, octree nodes contained type and data so it made sense to return a `NodeId` but now we really only need this function to return the type (and the level). So maybe we can change this function to return a `IntIntImmutablePair` to get one less allocation (in case the octree implementation overwrites the method, if not the default implementation still uses `NodeId` so no change there). And while we're dealing with such details maybe try changing the signature to accept a `IntIntPair` as parameter and fill it, to prevent creating a pair at every invocation of the function (even if that's not idiomatic)

chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
```diff
@@ -58,7 +58,7 @@ public WorldMapLoader(ChunkyFxController controller, MapView mapView) {
     RegionChangeWatcher regionWatcher = new RegionChangeWatcher(this, mapView);
 
     // Start worker threads.
-    RegionParser[] regionParsers = new RegionParser[Integer.parseInt(System.getProperty("chunky.mapLoaderThreads", "3"))];
+    RegionParser[] regionParsers = new RegionParser[Integer.parseInt(System.getProperty("chunky.mapLoaderThreads", "12"))];
```
leMaik, line 61: This number is (and was) pretty arbitrary. Should we use the thread count from `Settings` by default?

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -67,7 +68,7 @@
     int getData(NodeId node);
     default void startFinalization() {}
     default void endFinalization() {}
-    default Pair<NodeId, Integer> getWithLevel(int x, int y, int z) {
+    default void getWithLevel(IntIntMutablePair typeAndLevel, int x, int y, int z) {
```
leMaik, line 71: Maybe we should rename the first argument to `out` to make clear that it is used to return a value (or add javadoc)?

chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
NotStirred: commiting this was entirely accidental and is justt what I use to make my debugging faster, Yeah I think that'd be best

chunky/src/java/se/llbit/math/PackedOctree.java
```diff
@@ -366,12 +372,12 @@ public void set(Octree.Node data, int x, int y, int z) {
   /**
    * Gets a NodeID and depth of the node that is (or contains) the specified block.
    *
-   * Note: The creation of the Pair\<\> object seems to be a major time consumer in the actual tracing algorithm. (called from Octree.enterBlock)
+   * @param outTypeAndLevel is the reusable output type and level parameters, this is to save on allocation of {@code org.apache.commons.math3.util.Pair} and {@code PackedOctree.NodeId}
    *
-   * x, y, z are in octree coordinates, NOT world coordinates.
+   * @param x,y,z are in octree coordinates, NOT world coordinates.
```
leMaik, line 377:
```suggestion
   * x, y, z are in octree coordinates, NOT world coordinates.
```

PR #855, opened 2021-03-16
Reduce aliasing artifacts from fit-to-screen.

alexhliu:
When the view window and canvas sizes aren't a good ratio, there are aliasing artifacts. (Image works best fullscreen, notice the repeated grid-like aliasing and smeared pixels).
image

This fixes it by rounding the scale ratio down to the nearest 12.5% (1/8) which reduces artifacts while still allowing a reasonable range in scales.
image


leMaik:
Is there a JavaFX way to scale an image smoothly?

alexhliu:
Changing the JavaFX scaling techniques shouldn't do anything to affect this (I don't know of a way to do it anyways). Any interpolation mode should have this type of aliasing.

PR #858, opened 2021-03-22
Fix null pointer exception when chunk loading times out.

### LINKED ISSUE #853: Chunks not loaded after switching to One/ All emitter sampling
Minecraft version 1.16.5, Chunky Version 2.4.0-98-GC64B83F1 

Some (not all) chunks seem to disappear when selecting One/All emitter sampling.

Debug:

Exception in thread "Scene Manager" java.lang.NullPointerException
	at se.llbit.chunky.renderer.scene.Scene.loadChunks(Scene.java:908)
	at se.llbit.chunky.renderer.scene.Scene.reloadChunks(Scene.java:782)
	at se.llbit.chunky.renderer.scene.SynchronousSceneManager.reloadChunks(SynchronousSceneManager.java:190)
	at se.llbit.chunky.renderer.scene.AsynchronousSceneManager.run(AsynchronousSceneManager.java:81)

alexhliu:
Previously when chunk loading times out, `System.out.println(ignored.getCause().getMessage());` is called. However, `ignored.getCause()` returns null for timeout exceptions. I changed it to be more clear in logging and to use the `Log` utility instead of printing it to `stdout`. This should fix #853.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -904,8 +904,13 @@ public class Scene implements JsonSerializable, Refreshable {
 
         try {
           nextChunkDataTask.get(50, TimeUnit.MILLISECONDS);
-        } catch(TimeoutException | InterruptedException ignored) { // If except, load the chunk synchronously
-          System.out.println(ignored.getCause().getMessage());
+        } catch(TimeoutException | InterruptedException logged) { // If except, load the chunk synchronously
+          if (logged instanceof TimeoutException) {
+            Log.info("Chunk loading timed out.");
+          } else {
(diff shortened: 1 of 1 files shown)


NotStirred:
Is probably better to have a separate `catch` block for each exception, rather than checking `instanceof`

NotStirred (approved):

leMaik:
Why can chunk loading time out in the first place, i.e. why is the loading time limited to 50ms per chunk?

NotStirred:
that was an arbitrary value I set, it could be set to infinite wait time. a single chunk will load in far less time than 50ms
My thinking was if something goes wrong in the runnable to cause the thread to hang for a time, just stop and do it synchronously.

leMaik:
We can keep it, I was just curious :slightly_smiling_face:

PR #859, opened 2021-03-23
Improve scene loading speed drastically.

alexhliu:
> "just using java correctly" - @NotStirred 

My testing shows ~26x faster scene loading simply by using a buffered input stream (9k chunks in 3 seconds).

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -47,6 +47,7 @@ import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 
 import it.unimi.dsi.fastutil.io.FastBufferedInputStream;
+import it.unimi.dsi.fastutil.io.FastBufferedOutputStream;
 import org.apache.commons.math3.util.FastMath;
 import se.llbit.chunky.PersistentSettings;
 import se.llbit.chunky.block.*;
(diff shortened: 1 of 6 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1966,7 +1969,7 @@ private synchronized void saveOctree(RenderContext context, TaskTracker taskTrac
       task.update(1);
       Log.info("Saving octree " + fileName);
 
-      try (DataOutputStream out = new DataOutputStream(new GZIPOutputStream(context.getSceneFileOutputStream(fileName)))) {
+      try (DataOutputStream out = new DataOutputStream(new FastBufferedOutputStream(new GZIPOutputStream(context.getSceneFileOutputStream(fileName))))) {
```
leMaik, line 1972: Shouldn't this be the other way around? Same for the three other lines below where a gzip stream is used.

```suggestion
      try (DataOutputStream out = new DataOutputStream(new GZIPOutputStream(new FastBufferedOutputStream(context.getSceneFileOutputStream(fileName))))) {
```

leMaik (approved):
If only all optimizations where simple like this :sweat_smile:

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
NotStirred: Probably, though I'm not sure it matters, will profile it

alexhliu: If [this](https://stackoverflow.com/questions/1082320/what-order-should-i-use-gzipoutputstream-and-bufferedoutputstream) is to be believed, wrapping the buffer around the gzip stream is the faster way (how we have it currently).

PR #861, opened 2021-03-24
Better dump merging UX.

alexhliu:
- Merging through the GUI makes more sense (you can now merge even if the render is not started, merged frame will be displayed after merging is finished).
- Fixed merging task tracker lag (Apparently `Platform.runLater` does not being spammed with millions of tasks`).
- Improved CLI merging speed.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/RenderManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/RenderManager.java
@@ -138,11 +138,15 @@ public class RenderManager extends AbstractRenderManager implements Renderer {
             finalizeAllFrames = scene.shouldFinalizeBuffer();
             updateRenderState(scene);
             if (reason == ResetReason.SCENE_LOADED) {
+              // Make sure frame is finalized
+              bufferedScene.postProcessFrame(renderTask);
+
               // Swap buffers so the render canvas will see the current frame.
               bufferedScene.swapBuffers();
 
-              // Notify the scene listeners (this triggers a canvas repaint).
+              // Notify the scene listeners.
               sendSceneStatus(bufferedScene.sceneStatus());
+              canvas.repaint();
             }
           });
         }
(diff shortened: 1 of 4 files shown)


chunky/src/java/se/llbit/chunky/renderer/renderdump/DumpFormat.java
```diff
@@ -72,12 +72,16 @@ protected void readSamples(DataInputStream inputStream, Scene scene, IntConsumer
...
+    try (TaskTracker.Task task = taskTracker.task("Merging render dump", scene.canvasHeight())) {
+      int width = scene.canvasWidth();
       int previousSpp = scene.spp;
       long previousRenderTime = scene.renderTime;
 
       readHeader(inputStream, scene);
-      mergeSamples(inputStream, previousSpp, scene, task::update);
+      mergeSamples(inputStream, previousSpp, scene, prog -> {
+        if (prog % width == 0)
+          task.update(width);
```
leMaik, line 83:
```suggestion
          task.update(prog);
```

chunky/src/java/se/llbit/chunky/renderer/RenderManager.java
```diff
@@ -138,11 +138,15 @@ public RenderManager(RenderContext context, boolean headless) {
             finalizeAllFrames = scene.shouldFinalizeBuffer();
             updateRenderState(scene);
             if (reason == ResetReason.SCENE_LOADED) {
+              // Make sure frame is finalized
+              bufferedScene.postProcessFrame(renderTask);
```
leMaik, line 142: Why use the `renderTask`? If we use a new task tracker, the new `ostProcessFrame` overload wouldn't be needed. "Nested" tasks should work fine.

leMaik (approved):

chunky/src/java/se/llbit/chunky/renderer/renderdump/DumpFormat.java
alexhliu: Nice catch. Fixed in f11c3c6. It should be `prog/width` since this is tracking line by line and not pixel by pixel.

PR #862, opened 2021-03-24
Fix render reset on opening Sky & Fog tab.

### LINKED ISSUE #860: Opening the 'Sky & Fog' tab resets and resumes render.
Issue: When opening the Sky & Fog tab, if a render has been started, it will act as if the scene has been changed (resetting render if under 30 seconds, or prompting to reset if over 30 seconds), and resumes the render if it was paused (or rerenders the preview if on preview).

Steps to reproduce: Start a new instance of chunky. Without opening the Sky & Fog tab, either create a new scene or load an existing one. Ensure that render has atleast 1 spp, and it paused. Open the Sky & Fog tab. Render will reset and resume (or prompt to reset if time over 30 seconds).

example

alexhliu:
`fogColor` already had a fix where the listener was removed before updating and re-added after updating. I applied this fix to the other sources of resets. This fixes #860.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
+++ b/chunky/src/java/se/llbit/chunky/ui/render/SkyTab.java
@@ -18,6 +18,8 @@
 package se.llbit.chunky.ui.render;
 
 import javafx.beans.value.ChangeListener;
+import javafx.event.ActionEvent;
+import javafx.event.EventHandler;
 import javafx.fxml.FXML;
 import javafx.fxml.FXMLLoader;
 import javafx.fxml.Initializable;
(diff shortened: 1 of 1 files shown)


alexhliu:
This fixes #860

leMaik (changes requested):
Instead of removing and adding the listerners, could we change the setters to not trigger updates if the value didn't actually change?

alexhliu:
> Instead of removing and adding the listerners, could we change the setters to not trigger updates if the value didn't actually change?

Just tested, this dosen't work. There are probably rounding errors or something introduced when converting to JavaFX color and back. The only way would probably to compare equality with some epsilon (which I don't think is a good idea).

PR #863, opened 2021-03-24
Improved water world mode

### LINKED ISSUE #854: Water plane not below y < 1 (MC 1.17)
Y min clip can already handle Minecraft 1.17 y down to -64, but water planes are impossible for y < 1.
I found the following relevant code which would need to be updated:
 - for the GUI: `WaterTab>>update (line 82, 85)`, `WaterTab>>initialize (line 122)`
 - for the ray tracing: `PreviewRayTracer>>nextIntersection (line 95)`, `Scene>>isInWater (line 2382)`,

ShirleyNekoDev:
- fixes #854
- unify naming of water world mode
- rework config for the water world mode feature
- allow any water plane height (as a double)
- use slider for the water height
- add checkbox for default minecraft water height offset (-0.125 below full block)
- fallback for older scene files
- remove water plane height from persistent settings
- improve preview chunk grid (blue if submerged, darker grid for chunks in octree)

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
@@ -111,7 +111,7 @@ public class PreviewRayTracer implements RayTracer {
 
   private static boolean waterIntersection(Scene scene, Ray ray) {
     if (ray.d.y < 0) {
-      double t = (scene.getWaterPlaneHeight() - ray.o.y - scene.origin.y) / ray.d.y;
+      double t = (scene.getEffectiveWaterPlaneHeight() - ray.o.y - scene.origin.y) / ray.d.y;
       if (t > 0 && t < ray.t) {
         ray.t = t;
         Water.INSTANCE.getColor(ray);
@@ -121,7 +121,7 @@ public class PreviewRayTracer implements RayTracer {
       }
     }
     if (ray.d.y > 0) {
-      double t = (scene.getWaterPlaneHeight() - ray.o.y - scene.origin.y) / ray.d.y;
+      double t = (scene.getEffectiveWaterPlaneHeight() - ray.o.y - scene.origin.y) / ray.d.y;
       if (t > 0 && t < ray.t) {
         ray.t = t;
         Water.INSTANCE.getColor(ray);
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
@@ -19,11 +19,10 @@ package se.llbit.chunky.renderer.scene;
 import se.llbit.chunky.block.Air;
 import se.llbit.chunky.block.MinecraftBlock;
 import se.llbit.chunky.block.Water;
-import se.llbit.chunky.model.WaterModel;
 import se.llbit.chunky.renderer.WorkerState;
-import se.llbit.chunky.world.BlockData;
 import se.llbit.math.Ray;
 import se.llbit.math.Vector3;
+import se.llbit.math.Vector4;
 
 /**
  * @author Jesper Öqvist <jesper@llbit.se>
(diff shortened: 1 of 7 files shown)


chunky/src/java/se/llbit/chunky/ui/render/WaterTab.java
```diff
@@ -92,17 +83,25 @@ public WaterTab() throws IOException {
     waterColor.colorProperty().removeListener(waterColorListener);
     waterColor.setColor(ColorUtil.toFx(scene.getWaterColor()));
     waterColor.colorProperty().addListener(waterColorListener);
+
+    waterPlaneEnabled.setSelected(scene.isWaterPlaneEnabled());
+    waterPlaneHeight.setRange(scene.yClipMin, scene.yClipMax);
+    waterPlaneHeight.set(scene.getWaterPlaneHeight(true));
```
leMaik, line 89: Oh, _that's_ what the method is there for. It's weird that the getter doesn't return what the setter sets… Maybe `getEffectiveWaterPlaneHeight()` for the one that takes the offset into account – what do you think? :thinking:

lib/src/se/llbit/chunky/ui/SliderAdjuster.java
```diff
@@ -73,6 +73,10 @@ public double getMax() {
     return max;
   }
 
+  public Slider getSlider() {
+    return valueSlider;
+  }
+
```
leMaik, line 79: The slider is an implementation detail of this component and shouldn't be exposed to the outside.

leMaik (changes requested):
Nice! :grin: Just a few minor things, mostly nit-picky naming stuff…

chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
```diff
@@ -134,26 +133,60 @@ private static boolean waterIntersection(Scene scene, Ray ray) {
...
+
+  /**
+   * Projects a chunk border pattern onto the bottom plane of the octree (yMin).
+   * Changes colors for chunks inside the octree and submerged scenes.
+   * Use only in preview mode - the ray should hit the sky in a real render.
+   */
   private static boolean mapIntersection(Scene scene, Ray ray) {
-    if (ray.d.y < 0) {
-      double t = (scene.waterHeight - .125 - ray.o.y - scene.origin.y) / ray.d.y;
+    if (ray.d.y < 0) { // ray going below horizon
```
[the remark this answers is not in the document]
ShirleyNekoDev: If we get to this method the ray should already have missed the octree. From my understanding the ray direction below y < 0 indicates that the ray went into infinity. Effectively this case handles everything from the sky below the horizon and displays the chunk grid there.

chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
```diff
@@ -134,26 +133,60 @@ private static boolean waterIntersection(Scene scene, Ray ray) {
...
+      double t = (scene.yMin - ray.o.y - scene.origin.y) / ray.d.y;
```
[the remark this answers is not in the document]
ShirleyNekoDev: I didn't want to overlay any blocks of loaded chunks with this chunk grid plane and therefore moved it to the lowest location. Loaded chunks now perfectly fit into this grid. This works better than putting it at the height of the water plane in my opinion.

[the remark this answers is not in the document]
ShirleyNekoDev: So yes, this changes the old behavior, but the change was intentional.

chunky/src/java/se/llbit/chunky/ui/render/WaterTab.java
```diff
@@ -142,12 +142,12 @@ public void initialize(URL location, ResourceBundle resources) {
     waterPlaneEnabled.selectedProperty().addListener((observable, oldValue, newValue) ->
       scene.setWaterPlaneEnabled(newValue)
     );
+
     waterPlaneHeight.setName("Water height");
     waterPlaneHeight.setTooltip("The default ocean height is " + World.SEA_LEVEL + ".");
     waterPlaneHeight.clampBoth();
     waterPlaneHeight.onValueChange(value -> scene.setWaterPlaneHeight(value));
-    Slider slider = waterPlaneHeight.getSlider();
-    slider.setBlockIncrement(0.5);
```
leMaik: Missed this yesterday somehow. We could add a `setStep` method to our numeric slider.

ShirleyNekoDev: This code sadly does not provide stepping in the way I thought - only for slider movement via arrow keys. We could enable snapping but then values between steps would not be possible. 
I'd say this is not needed in this PR ;)

PR #864, opened 2021-03-25
Fix unnecessary octree resaving.

### LINKED ISSUE #844: Octree re-saves after every save
When hitting save, Chunky will re-save the octree, even if it hasn't been modified:
```
Saving scene Greenfield360-2
Saving octree Greenfield360-2.octree2
Octree saved
Saving render dump: C:\Users\jayjayli\.chunky\scenes\Greenfield360-2\Greenfield360-2.dump
Render dump saved: C:\Users\jayjayli\.chunky\scenes\Greenfield360-2\Greenfield360-2.dump
Scene saved
```

I can confirm this happnes with automatic dump saving too.

This issues occur on `Chunky 2.4.0-91-g5eaee6fa`

Previous versions of Chunky skipped the redundant octree write.

alexhliu:
Setting the octree last serialization timestamp before the file being written to closed was causing issues with the timestamp being a couple of milliseconds behind. Now, the timestamp is only set after the file is closed if saving completed without error. This fixes #844.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -1969,15 +1969,20 @@ public class Scene implements JsonSerializable, Refreshable {
       task.update(1);
       Log.info("Saving octree " + fileName);
 
+      boolean saved = false;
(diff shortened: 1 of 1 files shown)


leMaik:
Good catch! :clap: Didn't see that when I searched for the cause of this bug.

PR #865, opened 2021-03-25
Do not post process PFM exports

TheRealJayjayli:
Skipped post processing for PFM exports - I didn't swap any loops since PFMs should be right side up in most programs.

Hope this works 👀

### DIFF
--- a/chunky/src/java/se/llbit/pfm/PfmFileWriter.java
+++ b/chunky/src/java/se/llbit/pfm/PfmFileWriter.java
@@ -76,6 +76,7 @@ public class PfmFileWriter implements AutoCloseable {
       FloatBuffer floatBuffer = buffer.asFloatBuffer();
 
       // get the row's data as floats from raw pixel data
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/pfm/PfmFileWriter.java
```diff
@@ -75,19 +75,9 @@ private void writePixelData(Scene scene, ByteOrder byteOrder, TaskTracker.Task t
       ByteBuffer buffer = ByteBuffer.allocate(width*3*4).order(byteOrder);
       FloatBuffer floatBuffer = buffer.asFloatBuffer();
 
-      // get the row's data as floats
-      if (scene.postprocess == Postprocess.NONE)
-        // from raw pixel data
+      // get the row's data as floats from raw pixel data
```
leMaik, line 78:
```suggestion
      // get the row's data as floats from raw pixel data
      // (ignore post processing because that would clip the color range and defeat the purpose of HDR)
```

leMaik (approved):
Thank you!

PR #866, opened 2021-03-26
Minor optimisations 3

NotStirred:
Some minor optimisations to the map view loading.
Also fixed not using `SimpleChunkData` when possible

Draft because I'm going to do some more profiling before pushing another change

### DIFF
--- a/chunky/src/java/se/llbit/chunky/chunk/SimpleChunkData.java
+++ b/chunky/src/java/se/llbit/chunky/chunk/SimpleChunkData.java
@@ -16,9 +16,13 @@ import static se.llbit.chunky.world.Chunk.*;
  * 2d Biomes only
  */
 public class SimpleChunkData implements ChunkData {
+  /** These final fields are never written to, and are instead used for quickly setting existing blocks/biomes arrays
+   to all zeros, in {@link SimpleChunkData#clear()} */
+  private static final int[] emptyBlocks = new int[X_MAX * Y_MAX * Z_MAX];
+  private static final byte[] emptyBiomes = new byte[X_MAX * Z_MAX];
 
-  private int[] blocks;
-  private byte[] biomes;
+  private final int[] blocks;
+  private final byte[] biomes;
   private final Collection<CompoundTag> tileEntities;
   private final Collection<CompoundTag> entities;
 
@@ -84,8 +88,9 @@ public class SimpleChunkData implements ChunkData {
   }
 
   @Override public void clear() {
-    blocks = new int[X_MAX * Y_MAX * Z_MAX];
-    biomes = new byte[X_MAX * Chunk.Z_MAX];
+    //Quickly set all values to zero. Explanation here: https://github.com/chunky-dev/chunky/pull/866#issuecomment-808490741
+    System.arraycopy(emptyBlocks, 0, blocks, 0, emptyBlocks.length);
+    System.arraycopy(emptyBiomes, 0, biomes, 0, emptyBiomes.length);
     tileEntities.clear();
     entities.clear();
   }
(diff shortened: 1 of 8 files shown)


NotStirred:
I've been testing the performance of `ChunkData#clear`, as it's called once per-chunk for both the map screen and loading into the octree. (The chunkdata object is reused instead of reallocated)

On 3 JREs I profiled 5 different clearing approaches.
The fastest on all JREs was `System.arraycopy` from a static array of the same size as the blocks/biomes array initialised with all zeros. (The static arrays use about `0.25MiB` together)
Fast copy is an implementation where the first element of the array is set to `0` and is then copied using `System.arraycopy` in a binary fashion throughout the array until full

Zulu8
```
1. System.arraycopy (from static array) average: 0.2749
2. Fast copy average: 0.2835
3. For loop copy average: 0.2865
4. Arrays.Fill average: 0.2892
5. master average: 1.5826
```
openjdk8 hotspot
```
1. System.arraycopy (from static array) average: 0.2636
2. Fast copy average: 0.2766
3. Arrays.Fill average: 0.2786
4. For loop copy average: 0.2789
5. master average: 1.5155
```
openjdk11
```
1. System.arraycopy (from static array) average: 0.2736
2. For loop copy average: 0.2815
3. Fast copy average: 0.2818
4. Arrays.Fill average: 0.2824
5. master average: 0.6643
```

Unless you object to the minor jank @leMaik I'll go with the faster option

leMaik:
@NotStirred Semantically, `Arrays.fill` makes the most sense to me (to fill an array with zeros) and it's almost as fast as the other options.
If you want to go with `System.arraycopy`: Isn't there a special class for empty chunk data already? If so, we could copy the array from there (with a comment on why this is faster, maybe linking your benchmark).

NotStirred:
Sadly `EmptyChunkData` is just that, empty, it returns `0` for all values, and doesn't have any fields

leMaik:
Would be a reasonable place to put the empty array though, if we go with that approach. Not sure if that's worth it (on the other hand… we're talking about 250 KiB here…)

NotStirred:
It's something that will only be used in the `SimpleChunkData` class anyway, which is why I have it there for now
tbh it *really* doesn't matter

leMaik:
As long as it is clear why the code does what it does, it's fine either way. `Arrays.fill` is clear on its own, copying an empty array would need a comment that explains it. I prefer the first for that reason but yeah, we shouldn't overthink it.

NotStirred:
In my testing `SimpleChunkData#clear()` takes up around `1%` of time when loading for the map view, which isn't insignificant

leMaik:
@NotStirred Yes… but the difference between 0.2818 seconds and 0.27… seconds which probably is a microbenchmark prone to errors anyway… it's faster than master, so pick one of them.

NotStirred:
Actually the tests were reasonable, the results I posted were averages over several hundred tests, each consisting of 10k clears of existing `SimpleChunkData` objects. The results should be reliable

leMaik:
Perfect! :+1: As I said, `System.arraycopy` + comment is :100:% fine

chunky/src/java/se/llbit/chunky/chunk/SimpleChunkData.java
```diff
@@ -34,7 +38,7 @@ public SimpleChunkData() {
   }
 
   @Override public int maxY() {
-    return 255;
+    return 256;
```
leMaik, line 41: Why? :smile:

leMaik (changes requested):
Imho the `maxY()` change should be reverted, the rest is great :+1:

NotStirred:
the reason I did the `maxY` change is because with `minY` and `maxY` of `0` inclusive, there is expected to be a block inside the chunkdata, with this change there is not. Basically it works exactly how a list works with a for loop, len `5` actually means `0`-`4`
Though if you want me to revert I can do :)

leMaik:
Oooh… Yes okay, so empty chunks get confusing, I see :sweat_smile: Fine then :+1:

NotStirred:
I had mitigated this previously by having the default min and max at max and min integer, though that was a temporary solution at best and relied on the `ChunkData` implementation to handle it themselves

PR #867, opened 2021-03-27
ETA and progress bar for creating and loading Octrees and BVH.

alexhliu:
This does not seem to impact loading performance.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -1328,36 +1328,40 @@ public class Scene implements JsonSerializable, Refreshable {
 
     chunks = loadedChunks;
     camera.setWorldSize(1 << worldOctree.getDepth());
-    buildBvh();
-    buildActorBvh();
+    try (TaskTracker.Task task = taskTracker.task("Building BVH")) {
+      buildBvh(task);
+      buildActorBvh(task);
+    }
     Log.info(String.format("Loaded %d chunks", numChunks));
   }
 
-  private void buildBvh() {
+  private void buildBvh(TaskTracker.Task task) {
+    task.update("Building world BVH", 1, 0);
     final List<Primitive> primitives = new LinkedList<>();
 
     Vector3 worldOffset = new Vector3(-origin.x, -origin.y, -origin.z);
     for (Entity entity : entities) {
       primitives.addAll(entity.primitives(worldOffset));
     }
-    bvh = new BVH(primitives);
+    bvh = new BVH(primitives, task);
   }
 
-  private void buildActorBvh() {
+  private void buildActorBvh(TaskTracker.Task task) {
+    task.update("Building actor BVH", 1, 0);
     final List<Primitive> actorPrimitives = new LinkedList<>();
     Vector3 worldOffset = new Vector3(-origin.x, -origin.y, -origin.z);
     for (Entity entity : actors) {
       actorPrimitives.addAll(entity.primitives(worldOffset));
     }
-    actorBvh = new BVH(actorPrimitives);
+    actorBvh = new BVH(actorPrimitives, task);
   }
 
   /**
    * Rebuild the actors and the other blocks bounding volume hierarchy.
    */
   public void rebuildBvh() {
-    buildBvh();
-    buildActorBvh();
+    buildBvh(TaskTracker.Task.NONE);
+    buildActorBvh(TaskTracker.Task.NONE);
     refresh();
   }
 
@@ -1365,7 +1369,7 @@ public class Scene implements JsonSerializable, Refreshable {
    * Rebuild the actors bounding volume hierarchy.
    */
   public void rebuildActorBvh() {
-    buildActorBvh();
+    buildActorBvh(TaskTracker.Task.NONE);
     refresh();
   }
 
@@ -2100,10 +2104,9 @@ public class Scene implements JsonSerializable, Refreshable {
         Log.info("Octree loaded");
         calculateOctreeOrigin(chunks);
         camera.setWorldSize(1 << worldOctree.getDepth());
-        task.update("Loading BVH", 2, 1);
-        buildBvh();
-        buildActorBvh();
-        task.update(2);
+        task.update("Loading BVH", 1, 0);
+        buildBvh(task);
+        buildActorBvh(task);
         return true;
       } catch (IOException e) {
         Log.error("Failed to load chunk data!", e);
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -2061,15 +2061,32 @@ public class Scene implements JsonSerializable, Refreshable {
     try (TaskTracker.Task task = taskTracker.task("Loading octree", 2)) {
       task.update(1);
       Log.info("Loading octree " + fileName);
+
+      long length = context.getSceneFile(fileName).length();
+      long updateInterval = length / 1000;
+      final long[] nextUpdate = {0};
+      long startTime = System.currentTimeMillis();
+      task.update(1000, 1);
+
       try {
         long fileTimestamp = context.fileTimestamp(fileName);
         OctreeFileFormat.OctreeData data;
-        try (DataInputStream in = new DataInputStream(new FastBufferedInputStream(new GZIPInputStream(context.getSceneFileInputStream(fileName))))) {
+        try (DataInputStream in = new DataInputStream(new FastBufferedInputStream(new GZIPInputStream(new PositionalInputStream(context.getSceneFileInputStream(fileName), pos -> {
+          if (pos > nextUpdate[0]) {
+            task.update((int) (nextUpdate[0] * 1000.0 / length), startTime);
+            nextUpdate[0] += updateInterval;
+          }
+        }))))) {
           data = OctreeFileFormat.load(in, octreeImplementation);
         } catch (PackedOctree.OctreeTooBigException e) {
           // Octree too big, reload file and force loading as NodeBasedOctree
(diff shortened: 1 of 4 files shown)


chunky/src/java/se/llbit/math/BVH.java
```diff
@@ -341,6 +360,7 @@ private Node constructSAH_MA(Primitive[] primitives) {
         Primitive[] chunk = chunks.pop();
         if (chunk.length < SPLIT_LIMIT) {
           nodes.push(new Leaf(chunk));
+          task.updateInterval(progress += chunk.length, startTime, 16384);
```
leMaik, line 363:
```suggestion
          progress += chunk.length;
          task.updateInterval(progress, startTime, 16384);
```

Also, where does the magic `16384` number come from? Could it be moved to a constant with a name that explains it?

chunky/src/java/se/llbit/util/PositionalInputStream.java
```diff
@@ -0,0 +1,102 @@
...
+import java.io.InputStream;
+import java.util.function.LongConsumer;
+
+public class PositionalInputStream extends InputStream {
+    private InputStream stream;
+    private LongConsumer update;
+    private long count = 0;
+    private long mark;
+    private long markLimit;
+    private boolean canMark;
```
leMaik, line 30: This field could be removed, just call `stream.markSupported()` instead

chunky/src/java/se/llbit/util/PositionalInputStream.java
```diff
@@ -0,0 +1,102 @@
...
+        update.accept(count);
+        return actual;
+    }
+
+    @Override
+    public void reset() throws IOException {
+        if (canMark && count <= markLimit) {
+            stream.reset();
+            count = mark;
+        }
```
leMaik, line 93: This doesn't reset `count` if `stream.markSupported()` returns `false`.

chunky/src/java/se/llbit/util/TaskTracker.java
```diff
@@ -114,6 +113,45 @@ public void update(String task, int target, int done, String eta) {
       this.eta = eta;
       update();
     }
+
+    /** Changes the task name and state. Calculates eta. */
+    public void update(String task, int target, int done, long startTime) {
```
leMaik, line 118: If it changes the task name, it's a new task. :thinking: The automatic ETA calculation is great, though! Could be used in `RenderManager.updateRenderProgress` too once this is merged.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -900,7 +901,7 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
       for (int i = 0; i < chunkPositions.length; i++) {
         ChunkPosition cp = chunkPositions[i];
 
-        task.update(target, done);
+        task.update(target, done, startTime);
```
leMaik, line 904:
```suggestion
        task.update(target, done);
```

Putting the start time in every update is strange… Instead you could set it automatically when the task starts, i.e. when `taskTracker.task()` is called and the `Task` is created.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1304,7 +1306,7 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
             waterTexture.set(cp.x * 16 + x - origin.x, cp.z * 16 + z - origin.z, waterMix);
           }
         }
-        task.update(target, done);
+        task.update(target, done, startTime);
```
leMaik, line 1309:
```suggestion
        task.update(target, done);
```

chunky/src/java/se/llbit/util/TaskTracker.java
```diff
@@ -52,11 +53,9 @@ public TaskTracker(ProgressListener progress, TaskBuilder taskBuilder,
...
   public static class Task implements AutoCloseable {
     public static final Task NONE = new Task(null, null, "None", 1) {
-      @Override protected void update() {
-      }
-
-      @Override public void close() {
-      }
+      @Override protected void update() { }
+      @Override public void close() { }
+      @Override public void update(String task, int target, int done, long startTime) { }
```
leMaik, line 58:
```suggestion
      @Override public void update(String task, int target, int done) { }
```

chunky/src/java/se/llbit/util/TaskTracker.java
```diff
@@ -114,6 +113,45 @@ public void update(String task, int target, int done, String eta) {
       this.eta = eta;
       update();
     }
+
+    /** Changes the task name and state. Calculates eta. */
+    public void update(String task, int target, int done, long startTime) {
+      long etaSeconds = 0;
+      if (done > 0) {
```
leMaik, line 120: The ETA should only be shown if a start time was set.

leMaik (changes requested):
The automatic ETA calculation is a great idea. :+1: However I think that re-using a `Task` for multiple tasks is counter-intuitive. Instead, we should pass around the `TaskTracker` and create new `Task`s as needed. The start time of a task could then be set automatically.

chunky/src/java/se/llbit/util/PositionalInputStream.java
alexhliu: Count should not be reset if this is called and mark is not supported. ([Actually it should probably throw an IOException](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#reset--))

chunky/src/java/se/llbit/math/BVH.java
alexhliu: `16384` was just a random number I chose that was low enough that it would still show progress, but not so high that it would hang / crash due to the amount of calls. The value probably depends on the application so I'm not sure moving it to a global constant is a good idea. I can add a comment explaining this.

chunky/src/java/se/llbit/util/PositionalInputStream.java
alexhliu: I changed `reset` to throw an `IOException` if `mark` has not been called, more bytes have been read than `readlimit`, or if `markSupported` returns `false`.

In these cases, the position of the stream (and if an `IOException` is thrown) depends on the underlying stream so the position is unpredictable. (This is the strictest implementation of the [`InputStream.reset` contract](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#reset--))

chunky/src/java/se/llbit/util/TaskTracker.java
alexhliu: Updating with the ETA is now under `updateEta`. The ETA start time is pulled from the task's `startTime` which is set to the task creation time, or when `Task.updateStartTime` is called.

chunky/src/java/se/llbit/math/BVH.java
alexhliu: I changed the BVH progress to include loading primitives (which takes about half the time). To implement this, the task now goes from 0 to 1000 with progress updating every subdivision (0-500 on loading primitives, 500-1000 on building BVH).

chunky/src/java/se/llbit/util/TaskTracker.java
alexhliu: This is still important for usages where an existing task's name is changed. I don't intend on refactoring all the usages of task in this PR 🙂. However, I did change it to create a new task where relevant.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1328,32 +1326,49 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
...
+
+    double primitiveScaler = 500.0 / primitives.size();
+    bvh = new BVH(primitives, i -> task.updateInterval((int) (i * primitiveScaler) + 500, 1));
   }
 
   private void buildActorBvh(TaskTracker.Task task) {
-    task.update("Building actor BVH", 1, 0);
+    double entityScaler = 500.0 / entities.size();
+    int done = 0;
+    task.update(1000, 0);
```
leMaik, line 1359: What's up with the scaling? Why not just `task.update(entities.size(), 0);`?

chunky/src/java/se/llbit/util/PositionalInputStream.java
leMaik: Thank you :+1: This should ensure that we don't get any surprises when using this stream.

leMaik (approved):
The progress scaling is strange, but that's not a blocker. :+1:

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
alexhliu: The scaling is since we don't actually know how many elements there are going to be. Ideally, it should be `entities.size() + primitives.size()`, but we don't know the primitive length until it's halfway done. In addition, loading primitives takes about half the time which a pure count based approach wouldn't reflect.

leMaik: Oh, I see… then I guess 1000 and 500 are used because they are "big enough" to show some actual progress? Should we make`task.update` accept doubles instead in the future?

alexhliu: An `updatePercent` might be useful. I'll have to see when (if) I refactor all the other usages of TaskTracker.

leMaik: I don't expect you to do that :upside_down_face: Maybe `task.subtask(scale)` which returns a new `Task` that will update the original task whenever it is updated? :thinking: Can be done at a later point though, I'll merge this now. The progress bar improved a lot already thanks to your work :pray:

PR #868, opened 2021-03-28
Bulk octree set

aTom3333:
This PR changes the way blocks are set into the octree during loading. 
Setting a block is quite slow (we need to walk down the tree, make the change and try to merge). Especially when doing work that is undone later like when node are subdivised and later merged back because they were of the same type.
The idea of this PR is to amortized a bit this cost by setting a bunch of blocks at once, here 16\*16\*16. This is done with a new method on octree implementation with a default implementation that simply set block one by one. There is a specialized implementation for Packed.
With this change (and by using Packed octree) I saw a 10% reduction of the total loading time of 9 regions of the greenfield map (128.9s before, 116.3s after)

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -826,7 +826,7 @@ public class Scene implements JsonSerializable, Refreshable {
         return;
       }
 
-      int requiredDepth = calculateOctreeOrigin(chunksToLoad);
+      int requiredDepth = calculateOctreeOrigin(chunksToLoad, false);
 
       // Create new octree to fit all chunks.
       palette = new BlockPalette();
@@ -1381,7 +1381,7 @@ public class Scene implements JsonSerializable, Refreshable {
     refresh();
   }
 
-  private int calculateOctreeOrigin(Collection<ChunkPosition> chunksToLoad) {
+  private int calculateOctreeOrigin(Collection<ChunkPosition> chunksToLoad, boolean centerOctree) {
     int xmin = Integer.MAX_VALUE;
     int xmax = Integer.MIN_VALUE;
     int zmin = Integer.MAX_VALUE;
@@ -1411,7 +1411,15 @@ public class Scene implements JsonSerializable, Refreshable {
     int maxDimension = Math.max(yMax - yMin, Math.max(xmax - xmin, zmax - zmin));
     int requiredDepth = QuickMath.log2(QuickMath.nextPow2(maxDimension));
 
-    origin.set(xmin, 0, zmin);
+    if(centerOctree) {
+      int xroom = (1 << requiredDepth) - (xmax - xmin);
+      int yroom = (1 << requiredDepth) - (yMax - yMin);
+      int zroom = (1 << requiredDepth) - (zmax - zmin);
+
+      origin.set(xmin - xroom / 2, -yroom / 2, zmin - zroom / 2);
+    } else {
+      origin.set(xmin, 0, zmin);
+    }
     return requiredDepth;
   }
 
@@ -2090,7 +2098,7 @@ public class Scene implements JsonSerializable, Refreshable {
         palette.applyMaterials();
         task.update(2);
         Log.info("Octree loaded");
-        calculateOctreeOrigin(chunks);
+        calculateOctreeOrigin(chunks, data.version < 6);
         camera.setWorldSize(1 << worldOctree.getDepth());
         buildBvh();
         buildActorBvh();
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -1397,11 +1397,7 @@ public class Scene implements JsonSerializable, Refreshable {
     int maxDimension = Math.max(yMax - yMin, Math.max(xmax - xmin, zmax - zmin));
     int requiredDepth = QuickMath.log2(QuickMath.nextPow2(maxDimension));
 
-    int xroom = (1 << requiredDepth) - (xmax - xmin);
-    int yroom = (1 << requiredDepth) - (yMax - yMin);
-    int zroom = (1 << requiredDepth) - (zmax - zmin);
-
-    origin.set(xmin - xroom / 2, -yroom / 2, zmin - zroom / 2);
+    origin.set(xmin, 0, zmin);
     return requiredDepth;
   }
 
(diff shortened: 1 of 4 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1397,11 +1411,7 @@ private int calculateOctreeOrigin(Collection<ChunkPosition> chunksToLoad) {
     int maxDimension = Math.max(yMax - yMin, Math.max(xmax - xmin, zmax - zmin));
     int requiredDepth = QuickMath.log2(QuickMath.nextPow2(maxDimension));
 
-    int xroom = (1 << requiredDepth) - (xmax - xmin);
-    int yroom = (1 << requiredDepth) - (yMax - yMin);
-    int zroom = (1 << requiredDepth) - (zmax - zmin);
-
-    origin.set(xmin - xroom / 2, -yroom / 2, zmin - zroom / 2);
```
leMaik, line 1404: This probably shouldn't have been removed.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -893,6 +893,9 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
 
       ChunkPosition[] chunkPositions = chunksToLoad.toArray(new ChunkPosition[0]);
 
+      int[] cubeWorldBlocks = new int[16*16*16];
+      int[] cubeWaterBlocks = new int[16*16*16];
```
leMaik, line 897: We should change the Octree interface so that it can do such optimizations internally. The scene shouldn't need to care about this.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
aTom3333: this piece of code made it so the loaded chunks where place at the middle of the octree. 
For example if you load a 3\*3 chunks area (with a small ymax-ymin), you have a 48\*48 blocks area so you need a octree with depth 6 ie, a 64\*64 blocks octree. With this piece of code, the 48\*48 area area is placed in the [8; 56]\*[8; 56]. Without this piece of code, it is placed in the [0; 48]\*[0; 48] area.
Visually this makes no difference, the difference is that without this translation, chunks are aligned with the nodes of the octree so it is possible to load a whole 16\*16\*16 blocks at a time, if they weren't aligned this would not be possible. (In this case, the piece of code made it so that it would at least be 8-aligned instead of 16-aligned so it would still be possible to set 8\*8\*8 blocks at a time but the more we can do at a time, the better)

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
aTom3333: It seemed better to me, the scene knows when it is setting blocks in bulk (loading) and it is setting blocks individually (finalization).
It is still possible to encapsulate this optimization in the octree (in fact that's what I did initially) but it adds quite a bit of complexity:
 - the octree has to keep a 16\*16\*16 buffer and remember the position this buffer has to be set
 -  every time `set` is called, we check that the set is in the same 16\*16\*16 cube than the previous, if not we actually do the bulk set, emptying the buffer. After that the block is added to the buffer
 - As we want `set` to set blocks individually during the finalization phase, we need a flag that says whether `set` operation should be accumulated for bulk set or done indiviidually, this flag is set to accumulate initially and changed when `startFinalization` is called.
 - If before finalization, some blocks are set in cube 1, then some blocks are set in cube 2 (so the buffer for cube 1 has been flushed and the blocks inserted) and then again some blocks in cube 1 are set, but not the same blocks that were set previously, when the insertion of those new blocks will be done, the insertion can either do as the `setCube` method and replace the whole content (which is bad because only some of the blocks where set so the other should have kept there value), or the buffer needs to also keep track of which block is set or not, and when the insertion is done, the blocks of the buffer that haven't been set need some handling to not lose the previous value (which is not good because this additional handling could be expensive).

So, as I see things, if this was implemented only inside the octree without change to the external interface, we have the choice between a erroneous way of doing things that happens to work today because the cubes are only being set once (but that's fragile) or a correct way that can't take full advantage of the optimization.

With a new method added to the interface like `setCube`, the semantic makes it obvious that the whole cube will be replaced by what we give it (because we can't give it a cube with some blocks are not set) so it can takes full advantages of the optimization without introducing errors when we need to set some blocks in cube while leaving the other untouched (the origianl `set` method is used for that).

Maybe there is another way of doing things that I haven't considered but this seems to be better to me, in fact I'd argue that the chunks should have control on how they insert themselves in the octree, this would make it possible for the future 32\*32\*32 chunks (from cubic chunks) to insert themselves in one operation, but that feels very non -OOP/separation of concerns.

Sorry for the big block of text, I feel like I had a hard time expressing myself and this may not be very clear, don't hesitate to ask any question.

leMaik: Don't worry, I really appreciate the detailed explanation. I was mostly worried about how this optimization (obviously) reduces the separation of concerns a bit. Then again I guess optimizations like this are not possible without doing so – the overhead you described probably makes it more complicated than one "special" set method.

Another way (that would keep these optimizations in the octree implementation): What would be cool (especially for the 32x32x32 chunks) would be if the _octree_ could decide how to insert the blocks. Maybe by passing a function that produces the chunk sections or individual blocks and then the octree builds itself from that? A bit like the BVH constructor that gets a list of entities (or primitives) and decides what to do with them by itself.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
leMaik: Would this then break existing scenes when reloading the chunks?

aTom3333: Sorry for the delay. It breaks existing when loading them (the octree is shifted somewhere else) but reloading the chunks refixes it.
If I want my change and be backward compatible maybe I should store the origin so it is not recomputed when loading an existing scene (and it is recomputed using the old version is the origin information is not there)

leMaik: No problem. I'm fine with having the octree origin not at its center. Maybe we can just bump the octree version and move the old octree when it is loaded?
This won't affect entity positions, right?

aTom3333: For now, only the octree is shifted when loading an existing scene, entities are still correctly relative to the camera (so they are not where they should be relative to the octree)
Looking at the code, I see that the origin is used to place emitter position as well, I don't how if they are shifted as well or not for now

aTom3333:
Now loading an existing scene gives correct result, everything is where is should be relative to each other (camera, octree, entities, emitter positions)

PR #869, opened 2021-03-28
Reduce memory usage of sign texture

aTom3333:
The way SignTexture works is it has 2 textures, one for the background, shared with all the other signs of the same material, and one texture for the text, unique per sign.
The text texture ends up consuming a lot of memory so I changed the way it is stored, instead of using a 32 bpp bitmap, I'm using a 4 bpp bitmap because the color of sign text is already limited to 16 choices. As I also need to represent transparency, I use an additional 1 bpp bitmap that acts as a mask. So in the end, I am using 5 bits per pixels instead of the 32 initially, dividing by a lit more than 6 the memory usage of those texture.
As an additional optimization, sign that don't contain any text don't have a text texture at all, leading to more savings. This is especially good as sign are often use without text for decoration.
Before the change, the retained memory for one SignTexture was 18 706 bytes. After this change it is 3 117 bytes for a sign with text and only 125 bytes for a sign without text.
With a sign heavy map like greenfield, the gain is substantial, ~3.4GB vs ~2.8GB process memory wise.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/resources/BinaryBitmapImage.java
+++ b/chunky/src/java/se/llbit/chunky/resources/BinaryBitmapImage.java
@@ -1,0 +1,55 @@
+/* Copyright (c) 2020-2021 Chunky contributors
+ *
+ * This file is part of Chunky.
+ *
+ * Chunky is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Chunky is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with Chunky.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package se.llbit.chunky.resources;
+
+/**
+ * A image type using 1 bit per pixel
+ */
+public class BinaryBitmapImage {
+  public final int width;
+  public final int height;
+  private final byte[] data;
+
+  public BinaryBitmapImage(int width, int height) {
+    this.width = width;
+    this.height = height;
+    int pixelCount = width*height;
+    int byteCount = (pixelCount + 7) / 8;
+    data = new byte[byteCount];
+  }
+
+  public void setPixel(int x, int y, boolean value) {
+    int index = y * width + x;
+    int byteIndex = index / 8;
+    int bitIndex = index % 8;
+    int bit = 1 << (7 - bitIndex);
+    if(value)
+      // set the bit
+      data[byteIndex] |= bit;
+    else
+      // clear the bit
+      data[byteIndex] &= ~bit;
+  }
+
+  public boolean getPixel(int x, int y) {
+    int index = y * width + x;
+    int byteIndex = index / 8;
+    int bitIndex = index % 8;
+    int bit = 1 << (7 - bitIndex);
+    return (data[byteIndex] & bit) != 0;
+  }
+}
--- a/chunky/src/java/se/llbit/chunky/resources/BinaryBitmapImage.java
+++ b/chunky/src/java/se/llbit/chunky/resources/BinaryBitmapImage.java
@@ -36,7 +36,7 @@ public class BinaryBitmapImage {
     int index = y * width + x;
     int byteIndex = index / 8;
     int bitIndex = index % 8;
-    int bit = 1 << (7 - bitIndex);
+    int bit = 1 << bitIndex;
     if(value)
       // set the bit
       data[byteIndex] |= bit;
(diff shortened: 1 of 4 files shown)


chunky/src/java/se/llbit/chunky/resources/BinaryBitmapImage.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package se.llbit.chunky.resources;
+
+/**
+ * A image type using 1 bit per pixel
+ */
+public class BinaryBitmapImage {
+  public final int width;
+  public final int height;
+  private final byte[] data;
```
leMaik, line 25: I wonder if it would be faster to use `long[]` here. Java's `BitSet` seems to use an array of long.

chunky/src/java/se/llbit/chunky/resources/BinaryBitmapImage.java
```diff
@@ -0,0 +1,55 @@
...
+    int pixelCount = width*height;
+    int byteCount = (pixelCount + 7) / 8;
+    data = new byte[byteCount];
+  }
+
+  public void setPixel(int x, int y, boolean value) {
+    int index = y * width + x;
+    int byteIndex = index / 8;
+    int bitIndex = index % 8;
+    int bit = 1 << (7 - bitIndex);
```
leMaik, line 39: Why `7 - bitIndex`? Wouldn't it work without flipping that too (line 52 would also need to be changed)?

chunky/src/java/se/llbit/chunky/resources/PalettizedBitmapImage.java
```diff
@@ -0,0 +1,52 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with Chunky.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package se.llbit.chunky.resources;
+
+/**
+ * A image type using 4 bit per pixel
```
leMaik, line 20:
```suggestion
 * A image type using 4 bits per pixel, supporting up to 16 colors.
```

chunky/src/java/se/llbit/chunky/resources/SignTexture.java
```diff
@@ -36,74 +38,84 @@
...
-    BitmapImage img = new BitmapImage(width, height);
-    int[] data = img.data;
     int ystart = ymargin;
-    for (JsonArray line : text) {
-      if (line.isEmpty()) {
-        ystart += lineHeight;
-        continue;
+    boolean allEmpty = true;
+    for(JsonArray line : text) {
+      if(!line.isEmpty()) {
```
leMaik, line 54: To catch some some (probably rare) cases:

```suggestion
      if(!line.trim().isEmpty()) {
```

chunky/src/java/se/llbit/chunky/resources/SignTexture.java
```diff
@@ -36,74 +38,84 @@
...
   @Override
   public float[] getColor(double u, double v) {
-    float[] rgba = texture.getColor(u, v);
-    if (rgba[3] == 0) {
+    int x = (int)(u * 96 - Ray.EPSILON);
+    int y = (int) ((1 - v) * 48 - Ray.EPSILON);
+    if(textMask != null && textMask.getPixel(x, y)) {
+      float[] result = new float[4];
+      Color characterColor = Color.get(textColor.getPixel(x, y));
+      ColorUtil.getRGBAComponentsGammaCorrected(characterColor.rgbColor, result);
```
leMaik, line 115: We could store the gamma-corrected color in `textColor`. Slightly slower while loading but slightly faster while rendering.

leMaik (approved):

chunky/src/java/se/llbit/chunky/resources/BinaryBitmapImage.java
aTom3333: My intuition is that `long` is better when you have some functions that operate on multiple bits, for example `and`, `or` and `xor` of `BitSet` or probably 8 time as fast if they use `long` compared to `byte`, but when accessing one bit at a time, I don't think there is any difference.
On the other hand, using `byte` could lead to less memory allocated (but not in this case as the number of bytes allocated is multiple of 8).
That being said, if you want some measurements, I can do that

chunky/src/java/se/llbit/chunky/resources/BinaryBitmapImage.java
aTom3333: You're right, when I thought about the code, I thought of it as an array of bits where bits on the left come before bits on the right but it would work the same

chunky/src/java/se/llbit/chunky/resources/SignTexture.java
aTom3333: I guess you mean storing them in `Color` rather? I'll do that (It would even save me some allocation, yeah!)

leMaik: I was thinking about pre-calculating the gamma-corrected values for the texture

chunky/src/java/se/llbit/chunky/resources/SignTexture.java
aTom3333: line is a JsonArray so it was a bit more involved but that good

PR #870, opened 2021-03-28
Inline uv coordinate of vertices of texture triangles

aTom3333:
While I'm at it, here is another memory saving PR.
I inline the uv coordinate of the vertices of the triangle, this saves 72 byte per triangle (24 for each point: 8bytes from the pointer and 16 bytes from the `Vector2` header).
For entity heavy scene, that's MB worth of memory: for 1 region of greenfield, I had 300k entities, reducing memory by 21MB.
(If only java had value types already...)


leMaik:
I'm merging this because it doesn't affect readability (no vector2 methods were used anyway) so the 21 MB savings per region (of a probably typical city world) justifies it. :+1:

PR #872, opened 2021-04-01
Fix RGB textures with transparent color not working properly on Java 8.

leMaik:
OpenJDK <11 [has a bug][1] where it would ignore the transparent color of RGB PNG files. This was fixed [in OpenJDK 11][2] but is still an issue in Java 8.

This bug causes issues with some resourcepacks that use RGB PNGs with a transparent color, e.g. [Sapixcraft 128x][3] (e.g. the _Azure Bluet_).

|Before|After|
|---|---|
|image|image|

(Note how the issue does not appear on grass blocks where the texture uses 32-bit RGBA PNG)

[1]: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6788458
[2]: https://github.com/AdoptOpenJDK/openjdk-jdk11/commit/e9be6a236f8a752a768b29660482e52fba56e49c
[3]: https://sapixcraft.com/downloads/java/

### DIFF
--- a/chunky/src/java/se/llbit/resources/ImageLoader.java
+++ b/chunky/src/java/se/llbit/resources/ImageLoader.java
@@ -16,6 +16,9 @@
  */
 package se.llbit.resources;
 
+import java.awt.Image;
+import java.awt.Toolkit;
+import java.io.ByteArrayOutputStream;
 import se.llbit.chunky.resources.BitmapImage;
 import se.llbit.log.Log;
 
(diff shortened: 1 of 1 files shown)


leMaik:
Just checked it: The bug does not occur with Chunky 2.3.0 (i.e. without this fix) with OpenJDK 11 on Ubuntu 20.04.

leMaik:
The `Image` loading and waiting is pretty dirty but I didn't find a better way to do that (`Thread.sleep(1)` in the loop is too slow). At least this code doesn't execute in Java >=11 and one day we'll probably drop Java 8 support anyway and we can drop this workaround. :grimacing:

PR #876, opened 2021-04-04
Update camera controls in Help tab

jackjt8:
Camera key bindings work slightly differently with different projections; Controls listed are for standard projection which is typically what people use. Fullscreen has been removed thus `U` does nothing. `K` & `J` only move by 1 unit thus redundant. Added a mention of the `CTRL` modifier key.

### DIFF
--- a/chunky/src/res/se/llbit/chunky/ui/render/HelpTab.fxml
+++ b/chunky/src/res/se/llbit/chunky/ui/render/HelpTab.fxml
@@ -12,7 +12,7 @@
       <padding>
          <Insets bottom="10.0" left="10.0" right="10.0" top="10.0"/>
       </padding>
-      <Label text="Camera key bindings (standard projection):">
+      <Label text="Camera key bindings:">
          <font>
             <Font name="System Bold" size="12.0"/>
          </font>
@@ -23,8 +23,6 @@
       <Label text="D strafe right"/>
       <Label text="R move up"/>
       <Label text="F move down"/>
-      <Label text="K move forward"/>
-      <Label text="J move backward"/>
       <Label/>
       <Text strokeType="OUTSIDE" strokeWidth="0.0" text="Holding SHIFT makes the basic movement keys move 0.1 of the normal speed." wrappingWidth="251.2109375" />
       <Text strokeType="OUTSIDE" strokeWidth="0.0" text="Holding CTRL makes the basic movement keys move 100 of the normal speed." wrappingWidth="251.2109375" />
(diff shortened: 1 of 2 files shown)


chunky/src/res/se/llbit/chunky/ui/render/HelpTab.fxml
```diff
@@ -23,10 +23,10 @@
       <Label text="D strafe right"/>
       <Label text="R move up"/>
       <Label text="F move down"/>
-      <Label text="U toggle fullscreen mode"/>
-      <Label text="K move forward x100"/>
-      <Label text="J move backward x100"/>
+      <Label text="K move forward"/>
+      <Label text="J move backward"/>
```
leMaik, line 27: Maybe just drop J and K then while you're at it.

chunky/src/res/se/llbit/chunky/ui/render/HelpTab.fxml
```diff
@@ -12,7 +12,7 @@
       <padding>
          <Insets bottom="10.0" left="10.0" right="10.0" top="10.0"/>
       </padding>
-      <Label text="Camera key bindings:">
+      <Label text="Camera key bindings (standard projection):">
```
leMaik, line 15: All projections but isometric projection, right? :thinking:

jackjt8: That would be correct.

`(Excluding isometric)`?

chunky/src/res/se/llbit/chunky/ui/render/HelpTab.fxml
jackjt8: Having controls that work but undocumented is a bit annoying.. I could remove mention of them and maintain documentation elsewhere. Otherwise removing the actual bindings would probably be the best option.

chunky/src/res/se/llbit/chunky/ui/render/HelpTab.fxml
leMaik: Moving the camera up/down (and actually also the other directions) actually still work in isometric projection and can be important to stop the camera from clipping into the ground

jackjt8: Confirmed. The isometic camera still has the same controls and they work basically the same.. just a bit differently,

chunky/src/res/se/llbit/chunky/ui/render/HelpTab.fxml
jackjt8: I will look at removing the actual bindings tomorrow in a separate PR. Removing mention of the bindings is probably the first step.

leMaik:
Reason for dropping J and K: They are redundant and (more importantly) unintuitive. Controlling the position with AWSD is what most Minecraft players should be used to do.

PR #884, opened 2021-04-11
Add a y level slider to the map view

leMaik:
As [suggested on Reddit](https://old.reddit.com/r/chunky/comments/mmw1ae/ylevel_view_slider_in_chunky_2/), this adds a control to change the y level in the map view which is useful if you are searching for underground structures to render.

This also changes the reload behavior to not clear the chunk selection anymore.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/world/Chunk.java
+++ b/chunky/src/java/se/llbit/chunky/world/Chunk.java
@@ -187,7 +187,7 @@ public class Chunk {
         BlockPalette palette = new BlockPalette();
         loadBlockData(data, chunkData, palette);
         int[] heightmapData = extractHeightmapData(data, chunkData);
-        updateHeightmap(heightmap, position, chunkData, heightmapData, palette);
+        updateHeightmap(heightmap, position, chunkData, heightmapData, palette, yMax);
         surface = new SurfaceLayer(world.currentDimension(), chunkData, palette, yMax);
         queueTopography();
       } else if (version.equals("1.12")) {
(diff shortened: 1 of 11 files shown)


chunky/src/java/se/llbit/chunky/world/Chunk.java
```diff
@@ -188,7 +188,7 @@ private void loadSurface(Map<String, Tag> data, ChunkData chunkData) {
         loadBlockData(data, chunkData, palette);
         int[] heightmapData = extractHeightmapData(data, chunkData);
```
NotStirred, line 189:
```suggestion
        int[] heightmapData = extractHeightmapData(data, chunkData, yMax);
```

chunky/src/java/se/llbit/chunky/world/Chunk.java
```diff
@@ -188,7 +188,7 @@ private void loadSurface(Map<String, Tag> data, ChunkData chunkData) {
...
         updateHeightmap(heightmap, position, chunkData, heightmapData, palette);
```
NotStirred, line 190:
```suggestion
        updateHeightmap(heightmap, position, chunkData, heightmapData, palette, yMax);
```
and in implementation change `Math.max(1, y - 1)` to `Math.max(1, Math.min(y - 1, yMax)`

NotStirred (changes requested):
Heightmap change, otherwise LGTM

NotStirred:
I may also end up adding a `yMin` for CuC worlds in the future, of course for vanilla worlds it would server no purpose

PR #899, opened 2021-04-16
Better waterworld.

alexhliu:
Switch water-world to use the blockpalette water which allows for changing it's material properties.
Removed `Water.OCEAN_WATER` since it was not being used outside of water-world.
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/block/Water.java
+++ b/chunky/src/java/se/llbit/chunky/block/Water.java
@@ -21,9 +21,6 @@ public class Water extends MinecraftBlockTranslucent {
   public static final Water INSTANCE = new Water(0);
 
   public static final double TOP_BLOCK_GAP = 0.125;
-  
-  // Used only as starting material when camera is submerged.
-  public static final Water OCEAN_WATER = new Water(0, 1 << Water.FULL_BLOCK);
 
   public final int level;
   public final int data;
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
@@ -110,8 +110,14 @@ public class PreviewRayTracer implements RayTracer {
   }
 
   private static boolean waterIntersection(Scene scene, Ray ray) {
+    double t = (scene.getEffectiveWaterPlaneHeight() - ray.o.y - scene.origin.y) / ray.d.y;
+    if (scene.waterPlaneClip) {
+      Vector3 pos = new Vector3(ray.o);
+      pos.scaleAdd(t, ray.d);
+      if (scene.isChunkLoaded((int) pos.x, (int) pos.z))
+        return false;
+    }
     if (ray.d.y < 0) {
-      double t = (scene.getEffectiveWaterPlaneHeight() - ray.o.y - scene.origin.y) / ray.d.y;
       if (t > 0 && t < ray.t) {
         ray.t = t;
         Water.INSTANCE.getColor(ray);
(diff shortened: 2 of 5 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
```diff
@@ -110,18 +110,23 @@ public static boolean nextIntersection(Scene scene, Ray ray) {
   }
 
   private static boolean waterIntersection(Scene scene, Ray ray) {
+    double t = (scene.getEffectiveWaterPlaneHeight() - ray.o.y - scene.origin.y) / ray.d.y;
+    if (scene.waterPlaneClip) {
```
ShirleyNekoDev, line 114: Could this use a getter? 🙈

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -244,6 +244,7 @@
   protected boolean waterPlaneEnabled = false;
   protected double waterPlaneHeight = World.SEA_LEVEL;
   protected boolean waterPlaneOffsetEnabled = true;
+  protected boolean waterPlaneClip = true;
```
ShirleyNekoDev, line 247: I would maybe rename it to waterPlaneChunkClip to make it clear what is getting clipped.

chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
leMaik: This calculates `t` if `ray.d.y == 0`, which is not needed. I guess that never really happens a lot though.

chunky/src/res/se/llbit/chunky/ui/render/WaterTab.fxml
```diff
@@ -31,6 +31,7 @@
           </padding>
           <DoubleAdjuster fx:id="waterPlaneHeight" />
           <CheckBox fx:id="waterPlaneOffsetEnabled" mnemonicParsing="false" text="Lower water by minecraft offset (default)" />
+          <CheckBox fx:id="waterPlaneClip" mnemonicParsing="false" text="Clip the water plane in loaded chunks" />
```
leMaik, line 34: The word "clip" confused me here. How about

```suggestion
          <CheckBox fx:id="waterPlaneClip" mnemonicParsing="false" text="Hide the water plane in loaded chunks" />
```

chunky/src/java/se/llbit/chunky/block/Water.java
```diff
@@ -21,9 +21,6 @@
   public static final Water INSTANCE = new Water(0);
 
   public static final double TOP_BLOCK_GAP = 0.125;
-  
-  // Used only as starting material when camera is submerged.
```
leMaik, line 25: I wonder if this was important. Like is there something that is broken now, e.g. if the camera is submerged?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2468,6 +2479,17 @@ public void setBufferFinalization(boolean value) {
     }
   }
 
+  /**
+   * Query if a position is loaded.
```
leMaik, line 2483:
```suggestion
   * Query if a chunk is loaded.
```

leMaik (approved):

leMaik:
There is a off-by-one error somewhere. This only happens at two sides of the loaded chunks (negative x/z):

image

leMaik:
In a perfect world we could take the water world mode into account and not make the water change height here… but that would require re-loading the chunks when toggling water world mode.

image

PR #906, opened 2021-04-21
Fix dump loading progress indicator for ClassicDumpFormat

### LINKED ISSUE #905: Dump loading progress indicator is broken

ShirleyNekoDev:
- fixes #905 including "test"
- adds default fallback implementation for ProgressListener which ignores eta to allow creation of listeners as lambda

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/renderdump/ClassicDumpFormat.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/renderdump/ClassicDumpFormat.java
@@ -33,7 +33,7 @@ class ClassicDumpFormat extends DumpFormat {
   @Override
   public void readSamples(DataInputStream inputStream, Scene scene, PixelConsumer consumer, IntConsumer pixelProgress)
       throws IOException {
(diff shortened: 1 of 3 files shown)


chunky/src/java/se/llbit/chunky/renderer/renderdump/ClassicDumpFormat.java
```diff
@@ -43,7 +43,7 @@ public void readSamples(DataInputStream inputStream, Scene scene, PixelConsumer
         g = inputStream.readDouble();
         b = inputStream.readDouble();
         consumer.consume(pixelIndex, r, g, b);
-        pixelProgress.accept(pixelIndex);
+        pixelProgress.accept(progress++);
```
leMaik, line 46: Looks like the warning in line 38 also applies to the progress :smile:

ShirleyNekoDev: Indeed 😅

PR #916, opened 2021-04-27
Speed up isChunkLoaded by not checking the color.

leMaik:
Forgot to stage this in the last PR. A world texture contains a chunk texture iff that chunk is loaded, no need to get or check the color.


alexhliu:
image
This has issues with some worlds. Here the water plane clip is off by 1 chunk (it is correct on master).

leMaik:
That's weird. Can you send me that world please?

leMaik:
The bug above is now fixed :+1:

PR #918, opened 2021-04-29
Refactor render manager and add better support for rendering plugins.

alexhliu:
`InternalRenderManager` (implements `RenderManager`) deals with all the listeners, synchronizing scene state, and starts the desired `Renderer` when rendering starts.

`RenderWorkerPool` holds a bunch of `RenderWorker`s which are automatically CPU limited and do rendering tasks.

`Renderer` interface represents a whole renderer. It has 1 callback which is called after every frame and should put jobs on the `RenderWorkerPool`.

`PreviewRenderer` and `PathTracingRenderer` extend `TileBasedRenderer` which adapts the `RayTracer` interface to work with the new renderer system.

Renderers can now be hot-swapped. The `setRendererFactory` methods are deprecated, and create a new renderer and adds it to the renderers. Use the `addRenderer` methods instead.


chunky/src/java/se/llbit/chunky/main/Chunky.java
```diff
@@ -353,28 +343,28 @@ public boolean sceneInitialized() {
   }
 
   @PluginApi
-  public void setRendererFactory(RendererFactory rendererFactory) {
-    this.rendererFactory = rendererFactory;
+  public void setRendererFactory(RenderManagerFactory renderManagerFactory) {
```
leMaik, line 346: Should this be deprecated now? I don't know about other plugins but for me the only reason I needed this was to [replace the render manager with something that can switch the underlying renderer](https://github.com/chunky-dev/chunky-denoiser/blob/master/src/main/java/de/lemaik/chunky/denoiser/DenoiserPlugin.java#L37).

chunky/src/java/se/llbit/chunky/main/Chunky.java
```diff
@@ -405,23 +395,35 @@ public SceneFactory getSceneFactory() {
...
 
   @PluginApi
-  public void setRayTracerFactory(RayTracerFactory rayTracerFactory) {
-    this.rayTracerFactory = rayTracerFactory;
+  @Deprecated
+  public void setPreviewRayTracerFactory(RayTracerFactory previewRayTracerFactory) {
+    if (InternalRenderManager.previewRenderers.containsKey("Plugin Preview Renderer"))
+      Log.error("Attempted to register 2 plugin preview renderers.");
+    InternalRenderManager.previewRenderers.computeIfAbsent("Plugin Preview Renderer", name ->
+        new PreviewRenderer(previewRayTracerFactory.newRayTracer()));
```
leMaik, line 417: Just overwrite the previous renderer and log a warning. That would keep the previous behavior (last loaded plugin wins).

chunky/src/java/se/llbit/chunky/main/Chunky.java
```diff
@@ -405,23 +395,35 @@ public SceneFactory getSceneFactory() {
...
 
   @PluginApi
-  public RayTracerFactory getRayTracerFactory() {
-    return rayTracerFactory;
+  @Deprecated
+  public void setRayTracerFactory(RayTracerFactory rayTracerFactory) {
+    if (InternalRenderManager.renderers.containsKey("Plugin Renderer"))
+      Log.error("Attempted to register 2 plugin renderers.");
+    InternalRenderManager.renderers.computeIfAbsent("Plugin Renderer", name ->
+        new PathTracingRenderer(rayTracerFactory.newRayTracer()));
```
leMaik, line 426: See above.

chunky/src/java/se/llbit/chunky/main/Chunky.java
```diff
@@ -405,23 +395,35 @@ public SceneFactory getSceneFactory() {
...
+    InternalRenderManager.renderers.computeIfAbsent(name, n -> renderer);
   }
 
   @PluginApi
-  public void setRayTracerFactory(RayTracerFactory rayTracerFactory) {
-    this.rayTracerFactory = rayTracerFactory;
+  @Deprecated
+  public void setPreviewRayTracerFactory(RayTracerFactory previewRayTracerFactory) {
+    if (InternalRenderManager.previewRenderers.containsKey("Plugin Preview Renderer"))
+      Log.error("Attempted to register 2 plugin preview renderers.");
```
leMaik, line 415: Instead of logging this error I'd just log a warning (on all these methods) that `setPreviewRayTracerFactory` is deprecated and the plugin should be updated.

chunky/src/java/se/llbit/chunky/main/Chunky.java
```diff
@@ -405,23 +395,35 @@ public SceneFactory getSceneFactory() {
   }
 
   @PluginApi
-  public void setPreviewRayTracerFactory(RayTracerFactory previewRayTracerFactory) {
-    this.previewRayTracerFactory = previewRayTracerFactory;
+  public void addPreviewRenderer(String name, Renderer renderer) {
+    if (InternalRenderManager.previewRenderers.containsKey(name))
+      Log.error("Preview renderer already exists (do you have the same plugin loaded twice?): " + name);
+    InternalRenderManager.previewRenderers.computeIfAbsent(name, n -> renderer);
```
leMaik, line 401: If the same plugin is loaded twice, then better take the _last_ registered renderer, not the first one so that we don't have weird classpath problems

chunky/src/java/se/llbit/chunky/main/Chunky.java
```diff
@@ -405,23 +395,35 @@ public SceneFactory getSceneFactory() {
...
   }
 
   @PluginApi
-  public RayTracerFactory getPreviewRayTracerFactory() {
-    return previewRayTracerFactory;
+  public void addRenderer(String name, Renderer renderer) {
+    if (InternalRenderManager.renderers.containsKey(name))
+      Log.error("Renderer already exists (do you have the same plugin loaded twice?): " + name);
+    InternalRenderManager.renderers.computeIfAbsent(name, n -> renderer);
```
leMaik, line 408: If the same plugin is loaded twice, then better take the last registered renderer, not the first one so that we don't have weird classpath problems

chunky/src/java/se/llbit/chunky/renderer/InternalRenderManager.java
```diff
@@ -0,0 +1,502 @@
+/* Copyright (c) 2012-2016 Jesper Öqvist <jesper@llbit.se>
+ *
```
leMaik, line 2: Chunky contributors, too :wink:

chunky/src/java/se/llbit/chunky/renderer/InternalRenderManager.java
```diff
@@ -0,0 +1,502 @@
...
+        renderStart = System.currentTimeMillis();
+        if (mode == RenderMode.PREVIEW) {
+          // Preview with no CPU limit
+          pool.setCpuLoad(100);
+          render.setPostRender(previewCallback);
+          render.render(this);
+          pool.setCpuLoad(cpuLoad);
+        } else {
+          // Bail early if render is already done
+          if (bufferedScene.spp > bufferedScene.getTargetSpp()) {
```
leMaik, line 274:
```suggestion
          if (bufferedScene.spp >= bufferedScene.getTargetSpp()) {
```

Or does that break anything (post-processing or something)?

chunky/src/java/se/llbit/chunky/renderer/InternalRenderManager.java
```diff
@@ -0,0 +1,502 @@
...
+  }
+
+  private Renderer getPreviewRenderer() {
+    return previewRenderers.getOrDefault(previewRenderer, EMPTY_RENDERER);
+  }
+
+  @Override
+  public String[] getRenderers() {
+    return renderers.keySet().toArray(new String[0]);
+  }
```
leMaik, line 305: This could return a `Collection<String>` instead of creating a new array every time.

chunky/src/java/se/llbit/chunky/renderer/InternalRenderManager.java
```diff
@@ -0,0 +1,502 @@
...
+
+  @Override
+  public String[] getPreviewRenderers() {
```
leMaik, line 308: This could return a `Collection<String>` instead of creating a new array every time.

chunky/src/java/se/llbit/chunky/renderer/PathTracingRenderer.java
```diff
@@ -0,0 +1,71 @@
...
+public class PathTracingRenderer extends TileBasedRenderer {
+  protected RayTracer tracer;
+
+  public PathTracingRenderer(RayTracer tracer) {
+    this.tracer = tracer;
+  }
+
+  @Override
+  public String getIdString() {
+    return "InternalRender";
```
leMaik, line 17:
```suggestion
    return "PathTracingRenderer";
```

chunky/src/java/se/llbit/chunky/renderer/PathTracingRenderer.java
```diff
@@ -0,0 +1,71 @@
...
+
+        int offset = 3 * (y*width + x);
+        sampleBuffer[offset + 0] = (sampleBuffer[offset + 0] * spp + (sr * passinv)) * sinv;
+        sampleBuffer[offset + 1] = (sampleBuffer[offset + 1] * spp + (sg * passinv)) * sinv;
+        sampleBuffer[offset + 2] = (sampleBuffer[offset + 2] * spp + (sb * passinv)) * sinv;
+      });
+
+      manager.pool.awaitEmpty();
+      scene.spp += sppPerPass;
+      if (postRender.getAsBoolean()) return;
```
leMaik, line 68:
```suggestion
      if (postRender.getAsBoolean()) break;
```

I know it does the same but I had to read this twice to see that `return` breaks the loop here

chunky/src/java/se/llbit/chunky/renderer/PreviewRenderer.java
```diff
@@ -0,0 +1,96 @@
...
+public class PreviewRenderer extends TileBasedRenderer {
+  protected RayTracer tracer;
+
+  public PreviewRenderer(RayTracer tracer) {
+    this.tracer = tracer;
+  }
+
+  @Override
+  public String getIdString() {
+    return "InternalPreview";
```
leMaik, line 19:
```suggestion
    return "PreviewRenderer";
```

chunky/src/java/se/llbit/chunky/renderer/PreviewRenderer.java
```diff
@@ -0,0 +1,96 @@
...
+        if (sampleNum == 0 && x < (width - 1)) {
+          sampleBuffer[offset + 3] = state.ray.color.x;
+          sampleBuffer[offset + 4] = state.ray.color.y;
+          sampleBuffer[offset + 5] = state.ray.color.z;
+        }
+      });
+
+      manager.pool.awaitEmpty();
+      task.update(i+1);
+      if (postRender.getAsBoolean()) return;
```
leMaik, line 93:
```suggestion
      if (postRender.getAsBoolean()) break;
```

chunky/src/java/se/llbit/chunky/renderer/RenderManager.java
```diff
@@ -18,443 +19,128 @@
...
-    for (SceneStatusListener listener : sceneListeners) {
-      listener.sceneStatus(status);
-    }
-  }
+  void shutdown();
 
-  @Override public void shutdown() {
-    interrupt();
+  interface SampleBufferConsumer {
+    void accept(double[] samples, int width, int height);
```
leMaik, line 144: This will be cleaner once we have an actual object for the sample buffer :partying_face: (#786)

chunky/src/java/se/llbit/chunky/renderer/RenderWorkerPool.java
```diff
@@ -0,0 +1,151 @@
+/* Copyright (c) 2012-2019 Jesper Öqvist <jesper@llbit.se>
+ *
```
leMaik, line 2: + chunky contributors

chunky/src/java/se/llbit/chunky/renderer/RenderWorkerPool.java
```diff
@@ -0,0 +1,151 @@
...
+  public void awaitEmpty() throws InterruptedException {
+    synchronized (progress) {
+      while (progress.get() != localProgress.get()) {
+        progress.wait();
+      }
+    }
+  }
+
+  public void interrupt() {
+    Arrays.stream(workers).forEach(Thread::interrupt);
```
leMaik, line 149: A good old `for(Thread worker : workers)` would be fine here, I guess

chunky/src/java/se/llbit/chunky/renderer/Renderer.java
```diff
@@ -1,101 +1,36 @@
...
- * @author Jesper Öqvist <jesper@llbit.se>
- */
 public interface Renderer {
-  void setSceneProvider(SceneProvider sceneProvider);
-
-  void setCanvas(Repaintable canvas);
-
   /**
-   * Instructs the renderer to change its CPU load.
+   * Get a short id string representing this renderer.
```
leMaik, line 7: Then we also need a method to get a longer name (for the selection)?

chunky/src/java/se/llbit/chunky/renderer/Renderer.java
```diff
@@ -1,101 +1,36 @@
...
-
-  RenderStatus getRenderStatus();
-
-  /**
-   * Start up the renderer.
-   *
-   * <p>This should start all worker threads used by the renderer.
-   */
-  void start();
+  void render(InternalRenderManager manager) throws InterruptedException;
```
leMaik, line 29: is this _always_ an `InternalRenderManager`? What if a plugin replaced the `renderManagerFactory`?

chunky/src/java/se/llbit/chunky/renderer/scene/TileBasedRenderer.java
```diff
@@ -0,0 +1,105 @@
...
+    int tileWidth = manager.context.tileWidth();
+
+    if (prevWidth != width || prevHeight != height) {
+      prevWidth = width;
+      prevHeight = height;
+      cachedTiles.clear();
+
+      for (int i = 0; i < width; i += tileWidth) {
+        for (int j = 0; j < height; j += tileWidth) {
+          cachedTiles.add(new RenderTile(i, FastMath.min(i + tileWidth, width),
```
leMaik, line 99: Unrelated to this PR: Is `FastMath.min`/`.max` actually faster than `Math.min`/`max`?

chunky/src/java/se/llbit/chunky/ui/render/AdvancedTab.java
```diff
@@ -192,6 +183,14 @@ public PictureExportFormat fromString(String string) {
       scene.setPreventNormalEmitterWithSampling(newvalue);
       PersistentSettings.setPreventNormalEmitterWithSampling(newvalue);
     });
+
+    rendererSelect.setTooltip(new Tooltip("The renderer to use on a final render."));
```
leMaik, line 187:
```suggestion
    rendererSelect.setTooltip(new Tooltip("The renderer to use for rendering."));
```

(although that sounds strange too)

chunky/src/java/se/llbit/chunky/ui/render/AdvancedTab.java
```diff
@@ -192,6 +183,14 @@ public PictureExportFormat fromString(String string) {
...
+    rendererSelect.getSelectionModel().selectedItemProperty().addListener((observable, oldValue, newValue) ->
+        controller.getRenderManager().setRenderer(newValue));
+
+    previewSelect.setTooltip(new Tooltip("The renderer to use when previewing."));
```
leMaik, line 191:
```suggestion
    previewSelect.setTooltip(new Tooltip("The renderer to use for the preview."));
```

chunky/src/java/se/llbit/chunky/ui/render/AdvancedTab.java
```diff
@@ -226,11 +225,19 @@ public void setController(RenderControlsFxController controls) {
...
       if(shutdownAfterCompletedRender()) {
         // TODO: rewrite the shutdown alert in JavaFX.
         new ShutdownAlert(null);
       }
     });
+
+    if (rendererSelect.getItems().isEmpty()) {
+      RenderManager renderManager = controller.getRenderManager();
+      rendererSelect.getItems().addAll(renderManager.getRenderers());
+      rendererSelect.getSelectionModel().select(renderManager.getRendererName());
```
leMaik, line 238: We shouldn't confuse the _name_ and the "short" _id_ of a renderer. The ID should be used to identify the renderer internally, the name should only be used for display (and may change without breaking existing scenes).

The other selection boxes have a similar problem (except for the postprocessing mode), so that's fine for now. Make sure to store the _id_ in the scene description though, not the display name.

leMaik (changes requested):
Already said that on Discord but this looks really good. :+1: 
Can't wait to clean up the denoiser.

chunky/src/java/se/llbit/chunky/renderer/Renderer.java
alexhliu: I was thinking that the `Renderer` interface is only for plugins that fit into the existing Chunky code well. If a plugin were to change so many things that it needs to override the `RenderManager` other renderers probably wouldn't work in the first place (or it wouldn't make sense to have other renderers).

Essentially, this allows simple renderer plugins to use specific things like the `RenderWorkerPool` or `bufferedScene` without forcing low level renderer plugins to provide it or implement "dummy" versions (if that makes any sense).

leMaik: This sounds reasonable 👍
Maybe call it `DefaultRenderManager` then?

chunky/src/java/se/llbit/chunky/renderer/scene/TileBasedRenderer.java
leMaik: As benchmarked by @alexhliu  https://microbenchmarks.appspot.com/runs/5888757a-84de-4d1d-9da4-c77536213259

chunky/src/java/se/llbit/chunky/renderer/InternalRenderManager.java
alexhliu: Nice catch, mistranslated the original code.

chunky/src/java/se/llbit/chunky/main/Chunky.java
alexhliu: New behavior:
* INFO level log saying `setPreviewRayTracerFactory` is deprecated
* WARN level log warning 2+ plugin preview renderers are installed
* Last loaded plugin wins

chunky/src/java/se/llbit/chunky/renderer/PathTracingRenderer.java
```diff
@@ -0,0 +1,96 @@
...
+  protected RayTracer tracer;
+
+  public PathTracingRenderer(RayTracer tracer, String name, String id) {
+    this.tracer = tracer;
+    this.nameString = name;
+    this.idString = id;
+  }
+
+  @Override
+  public String getIdString() {
```
leMaik, line 36:
```suggestion
  public String getId() {
```

leMaik (changes requested):
So close! :tada: Just the id/name stuff mostly (drop the display name for now if you want and use the ID everywhere… I can add the display name for all things (post processing, octree, renderers) in another PR.

leMaik (approved):

chunky/src/java/se/llbit/chunky/renderer/Renderer.java
```diff
@@ -17,85 +16,52 @@
...
 
   /**
-   * Shut down the renderer.
-   *
-   * <p>This should interrupt all worker threads used by the renderer.
+   * This is called when the scene is reset. The default implementation does nothing.
+   * This is for {@code Renderer}s which need to export the scene data in some way.
    */
-  void shutdown();
+  default void sceneReset(DefaultRenderManager manager, ResetReason reason) {}
```
alexhliu, line 66: Add a method to see if the RenderManager should finalize the frames in the postRender callback.

leMaik:
As discussed on Discord, the selected renderer ID should be saved in the scene description and restored on load. If it doesn't exist anymore, use the default one. That's also what we do for post-processing and bvh.

PR #920, opened 2021-05-03
correct isBlockOnEdge checks

### LINKED ISSUE #911: yMax clipping breaks octree optimization
When yMax clipping is used, blocks get replaced with _any_ even if they are at the top of the octree.

image

As found by  @Lohfrum-Faleign

NotStirred:
Fixes #911

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -1014,8 +1014,8 @@ public class Scene implements JsonSerializable, Refreshable {
                 int cubeIndex = (cz * 16 + cy) * 16 + cx;
 
                 // Change the type of hidden blocks to ANY_TYPE
-                boolean notOnEdge = !chunkData.isBlockOnEdge(cx, y, cz) && !(y <= yMin || y >= yMax - 1);
-                boolean isHidden = notOnEdge
+                boolean onEdge = chunkData.isBlockOnEdge(cx, y, cz) || y <= yMin || y >= yMax - 1;
+                boolean isHidden = !onEdge
                         && palette.get(chunkData.getBlockAt(cx + 1, y, cz)).opaque
                         && palette.get(chunkData.getBlockAt(cx - 1, y, cz)).opaque
                         && palette.get(chunkData.getBlockAt(cx, y + 1, cz)).opaque
@@ -1080,7 +1080,7 @@ public class Scene implements JsonSerializable, Refreshable {
                       // Move plain water blocks to the water octree.
                       octNode = palette.airId;
 
-                      if(notOnEdge) {
+                      if(!onEdge) {
                         // Perform water computation now for water blocks that are not on th edge of the chunk
                         // Test if the block has not already be marked as full
                         if(((Water) palette.get(waterNode)).data == 0) {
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1133,7 +1133,7 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
                       }
                     }
                     cubeWaterBlocks[cubeIndex] = waterNode;
-                  } else if(cy + 1 < yMax && block instanceof Lava) {
+                  } else if(y + 1 < yMax && block instanceof Lava) {
```
NotStirred, line 1136: There is no way this was correct, I can move to a separate PR if requested

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1014,7 +1014,7 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
                 int cubeIndex = (cz * 16 + cy) * 16 + cx;
 
                 // Change the type of hidden blocks to ANY_TYPE
-                boolean notOnEdge = !chunkData.isBlockOnEdge(cx, cy, cz);
+                boolean notOnEdge = !chunkData.isBlockOnEdge(cx, y, cz) && !(y <= yMin || y >= yMax - 1);
                 boolean isHidden = notOnEdge
```
leMaik, line 1018:
```suggestion
                boolean onEdge = chunkData.isBlockOnEdge(cx, y, cz) || y <= yMin || y >= yMax - 1;
                boolean isHidden = !onEdge
```

I don't not think that this isn't not not harder to read.

leMaik (changes requested):

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
NotStirred: I was thinking the EXACT same thing, I didn't change it because `notOnEdge` has usages below, though I can just flip those too ;)

leMaik: Let's flip it then :+1:

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1080,7 +1080,7 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
                       // Move plain water blocks to the water octree.
                       octNode = palette.airId;
 
-                      if(notOnEdge) {
+                      if(!onEdge) {
```
leMaik, line 1083: Water finalization must also be performed if `y == yMin` or `y == yMax -1`. Only x/z edge (i.e. chunk crossing) is important here. This was the case before because `chunkData.isBlockOnEdge` (and `onEdge`) ignored the y coordinate.

leMaik (changes requested):

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
aTom3333: Should be good, it is done during the octree finalization, https://github.com/chunky-dev/chunky/blob/423ee0176b6b8da6d57166ca385d2fb78c1f5ec3/chunky/src/java/se/llbit/chunky/renderer/scene/OctreeFinalizer.java#L52

leMaik: Allright, maybe we could optimize this at some point in the future… Water finalization could happen here for y edge blocks too

PR #924, opened 2021-05-06
Make ChunkPosition#get much faster

NotStirred:
eliminates one hashmap get
also uses `computeIfAbsent` insead of `get` nullcheck `put` which is also faster

### DIFF
--- a/chunky/src/java/se/llbit/chunky/world/ChunkPosition.java
+++ b/chunky/src/java/se/llbit/chunky/world/ChunkPosition.java
@@ -44,25 +44,12 @@ public class ChunkPosition {
     return get(x >> 5, z >> 5);
   }
 
-  private static final Map<Integer, Map<Integer, ChunkPosition>> map = new ConcurrentHashMap<>();
+  //not using synchronized Long2ReferenceOpenHashMap due to using one lock. ConcurrentHashMap locks on buckets
+  private static final Map<Long, ChunkPosition> positions = new ConcurrentHashMap<>();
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/chunky/world/ChunkPosition.java
```diff
@@ -44,25 +44,12 @@ public ChunkPosition regionPosition() {
...
-    }
-
-    ChunkPosition chunkPosition = submap.get(z);
-    if (chunkPosition == null) {
-      chunkPosition = new ChunkPosition(x, z);
-      submap.put(z, chunkPosition);
-    }
-
-    return chunkPosition;
+    return positions.computeIfAbsent((x & INT_BIT_MASK) | (z & INT_BIT_MASK) << 32, (position) -> new ChunkPosition(x, z));
```
leMaik, line 52: Compared to `getLong()`… does this shift the wrong value? `x` is supposed to be the highest 32 bits, not z.

NotStirred: Oh interesting, didn't realise chunky differed to vanilla here
Yeah sure I'll switch it

NotStirred: will also inline some constants as discussed on discord

leMaik: Didn't even know how MC did it until now

Issue #931, opened 2021-05-10
Chunk Loading Error


TheRealJayjayli:
Sorry for the undescriptive title, I don't know what to call this -

```
Exception in thread "Scene Manager" java.lang.ArrayIndexOutOfBoundsException: -779
	at se.llbit.math.PackedOctree.findSpace(PackedOctree.java:208)
	at se.llbit.math.PackedOctree.insertTempTree(PackedOctree.java:421)
	at se.llbit.math.PackedOctree.insertTempTree(PackedOctree.java:423)
	at se.llbit.math.PackedOctree.insertTempTree(PackedOctree.java:423)
	at se.llbit.math.PackedOctree.insertTempTree(PackedOctree.java:423)
	at se.llbit.math.PackedOctree.setCube(PackedOctree.java:508)
	at se.llbit.math.Octree.setCube(Octree.java:787)
	at se.llbit.chunky.renderer.scene.Scene.loadChunks(Scene.java:1198)
	at se.llbit.chunky.renderer.scene.SynchronousSceneManager.loadFreshChunks(SynchronousSceneManager.java:163)
	at se.llbit.chunky.renderer.scene.AsynchronousSceneManager.lambda$loadFreshChunks$2(AsynchronousSceneManager.java:139)
	at se.llbit.chunky.renderer.scene.AsynchronousSceneManager.run(AsynchronousSceneManager.java:81)
```

I also saw `Exception in thread "Scene Manager" java.lang.ArrayIndexOutOfBoundsException: -844` (`-844` instead of `-779`)

I had trouble finding the exact chunks - so I have attatched the region files, a small selection from IHOU2. I have also attatched some screenshots of what I see in the Render Preview:

![image](https://user-images.githubusercontent.com/52780721/117696483-40410080-b1b9-11eb-96b0-875cb8f15c37.png)

![image](https://cdn.discordapp.com/attachments/543090398671929344/841356921437028352/unknown.png)

TheRealJayjayli:
Oops. Forgot the Region Files;
[ihou2.zip](https://github.com/chunky-dev/chunky/files/6453781/ihou2.zip)

leMaik:
Fixed by #933

Issue #923, opened 2021-05-04
Octree bugs


leMaik:
Original comment at https://github.com/chunky-dev/chunky/issues/911#issuecomment-825998174

- Octree y origin is wrong (see #908)
- Sometimes after changing y clipping I get a stack overflow in `PackedOctree.finalizationNode`
- Sometimes after changing y clipping the octree doesn't load properly until the chunks are reloaded multiple times (and even that sometimes doesn't fix it)
- The bug seen in the picture above also happens on the bottom of the loaded area if ymin clipping is enabled

All of this are regressions in 2.4.0

leMaik:
Fixed by #933

PR #933, opened 2021-05-12
Various octree bugfixes

### LINKED ISSUE #908: Inaccurate Y max Calculation when Y min>0
Was trying to display single layers of chunks so that I can make a schematic of a 1 chunk house I built.
I set the Y min to 73 to cut off everything below the bottom layer of the house, then I set the Y max to 74 and hit reload chunks. Instantly the layer of blocks gets teleported 53 blocks down to y=20. However if I keep Y Min at 0 and just adjust the Y Max alone, it makes the layer cuts accurately.

Y Min = 73, Y Max = 256
image
Y Min =73, Y Max = 74
image
As shown in this image, the Polished Diorite that was originally at y= 73.5 Is now at y=20.5
image

But if you leave Y Min at 0, Adjusting the Y Max works as normal.
Here y min=0, y max=85
image

aTom3333:
Should fix every bug listed in #923 
Should also fix #908 
Some detail about the reason the bugs were occurring and their fixes are in each commit message

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -981,7 +981,7 @@ public class Scene implements JsonSerializable, Refreshable {
             double y = pos.get(1).doubleValue();
             double z = pos.get(2).doubleValue();
 
-            if (y >= yClipMin && y <= yClipMax) {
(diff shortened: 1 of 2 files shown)


leMaik (approved):

aTom3333:
It seems that this also fixes #931

leMaik:
Great, thank you! :partying_face:

PR #901, opened 2021-04-17
Refactor block models

leMaik:
This PR refactors the block models so that they are reusable and the intersection code is not scattered across the models. This will allow implementing normal maps and specular maps and re-using the models e.g. by the OpenCL plugin.

Huge thanks to @alexhliu for converting most of the blocks.

Missing or known wrong blocks:

- [ ] Brewing Stand
- [ ] Hopper
- [ ] Flower pot
- [ ] GlassPane
- [ ] Cauldron
- [ ] HoneyBlock
- [ ] SlimeBlock
- [ ] Log
- [ ] Stem
- [ ] OrientedTexturedBlock (e.g. barrels)
- [ ] TopBottomOrientedTexturedBlock (e.g. loom, cartography table)

If you spot a wrong block in this branch, please let me know!

Issue #891, opened 2021-04-13
[water] Extreme Refraction Angles -> Black


IceSelkie:
When refracting the rays through the surface of water, if the angle is extreme enough the ray seems to get lost. If water opacity is above 0, it will give the color of "water" there, and at 0 it will be pure black.

Close to the surface of water:
![water blackness](https://user-images.githubusercontent.com/4878558/114609913-2eb91700-9c54-11eb-8ee2-a8ad90c5d390.png)

Zoomed in to one of the areas (towards the top left; decreased FoV):
![water blackness_2](https://user-images.githubusercontent.com/4878558/114609948-3b3d6f80-9c54-11eb-933a-e4fb273406a1.png)

One one-hundred-thousanth of a block from the edge of a glass pane:
![glasspane no black](https://user-images.githubusercontent.com/4878558/114610931-4fce3780-9c55-11eb-9662-97ccc1ffb82f.png)

leMaik:
Water: Maybe the ray intersects with the water surface twice (because it's close to parallel to the water surface) but Chunky assumes that it has left the water at that point. Not sure, just an assumption.

PR #935, opened 2021-05-12
Fix water artifact when looking at water on the horizon/with low incidence angle

aTom3333:
Normal mapping can introduce issue if not careful:
image
This image is diagram of what happens when a ray hits a surface that uses normal mapping and wants to reflect off of this surface.
Plain solid lines are surface, either real surface or fake surface simulated by normal mapping, dotted lines with arrow are normals, the black one is the real geometry normal, the red one is the normal obtained by normal mapping. solid lines with arrow are light paths.
The blue light path is the incoming ray, it intersect with the surface and wants to reflect off of it. 
The gray light path is the reflected ray we would have if there is no normal mapping, just using the real geometry surface.
The red light path is the reflected path if we are using the normal from the map, this is what is done when using normal mapping. As you can see this is problematic in this case because the reflected ray will be on the wrong side of the real geometry surface (black plain solid line) which can lead to visual artifacts.
The blue light path is my proposed solution, it forces the light ray to be on the right side of the surface, at a very low angle.
This is technically a hack but so is normal mapping in the first place, without normal mapping the situation where we hit surface from the back would be impossible so I believe this is a good solution.

The way this is implemented is that I keep track of both the real and displaced normal and when computing direction for reflection or refraction, I check to see if the direction is on the right side, if not, I correct that.
This means I had to change a lot of code that sets the normals.

Example of render where this kind of artifacts where visible:
Before:
sunset-before

After:
sunset-after

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
@@ -95,7 +95,7 @@ public class PathTracer implements RayTracer {
       Material currentMat = ray.getCurrentMaterial();
       Material prevMat = ray.getPrevMaterial();
 
-      if (!scene.stillWater && ray.getN().y != 0 &&
+      if (!scene.stillWater && ray.getNormal().y != 0 &&
           ((currentMat.isWater() && prevMat == Air.INSTANCE)
               || (currentMat == Air.INSTANCE && prevMat.isWater()))) {
         WaterModel.doWaterDisplacement(ray);
@@ -199,13 +199,13 @@ public class PathTracer implements RayTracer {
               double directLightG = 0;
               double directLightB = 0;
 
-              boolean frontLight = reflected.d.dot(ray.getN()) > 0;
+              boolean frontLight = reflected.d.dot(ray.getNormal()) > 0;
 
               if (frontLight || (currentMat.subSurfaceScattering
                   && random.nextFloat() < Scene.fSubSurface)) {
 
                 if (!frontLight) {
-                  reflected.o.scaleAdd(-Ray.OFFSET, ray.getN());
+                  reflected.o.scaleAdd(-Ray.OFFSET, ray.getNormal());
                 }
 
                 reflected.setCurrentMaterial(reflected.getPrevMaterial(), reflected.getPrevData());
@@ -214,7 +214,7 @@ public class PathTracer implements RayTracer {
 
                 Vector4 attenuation = state.attenuation;
                 if (attenuation.w > 0) {
-                  double mult = QuickMath.abs(reflected.d.dot(ray.getN()));
+                  double mult = QuickMath.abs(reflected.d.dot(ray.getNormal()));
                   directLightR = attenuation.x * attenuation.w * mult;
                   directLightG = attenuation.y * attenuation.w * mult;
                   directLightB = attenuation.z * attenuation.w * mult;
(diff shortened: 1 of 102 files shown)


leMaik:
The "hacky" solution seems reasonable to me :+1: I'll keep it in mind when I work on normal map support for other textures again.

Does this also fix the issue when looking at the surface from below (#891)?

aTom3333:
I want to say yes but I'm not succeeding at making a similar underwater scene for testing, my attempts are super dark underwater

leMaik:
Try increasing water visibility

aTom3333:
It wasn't working before but it is now
Master:
sunset-underwater-100-master

This PR before the last commit:
sunset-underwater-30-broken

Now:
sunset-underwater-100-fixed

leMaik:
This needs to be added to #901 (where it would also require less changed because there are only a few model intersection methods).

Would you like to look into that? :pray: I'm not sure if I should merge this in the meantime.

aTom3333:
Pull request moved to #939

leMaik:
Since #901 will take a while to be implemented, I'd like to merge this instead.

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
```diff
@@ -95,12 +95,12 @@ public static boolean pathTrace(Scene scene, Ray ray, WorkerState state, int add
...
       Material prevMat = ray.getPrevMaterial();
 
-      if (!scene.stillWater && ray.n.y != 0 &&
+      if (!scene.stillWater && ray.getN().y != 0 &&
           ((currentMat.isWater() && prevMat == Air.INSTANCE)
               || (currentMat == Air.INSTANCE && prevMat.isWater()))) {
         WaterModel.doWaterDisplacement(ray);
         if (currentMat == Air.INSTANCE) {
-          ray.n.y = -ray.n.y;
+          ray.invertNormal();
```
leMaik, line 103: Previously, only `y` was inverted. Now all components of the normal are inverted. Is this intentional, was the code wrong before?

chunky/src/java/se/llbit/math/Ray.java
```diff
@@ -494,4 +530,44 @@ public int getPrevData() {
   public int getCurrentData() {
     return currentData;
   }
+
+  public Vector3 getN() {
```
leMaik, line 534:
```suggestion
  /**
   * Get the normal of the previously hit surface.
   */
  public Vector3 getNormal() {
```

chunky/src/java/se/llbit/math/Ray.java
```diff
@@ -494,4 +530,44 @@ public int getPrevData() {
...
+    return n;
+  }
+
+  public Vector3 getGeomN() {
```
leMaik, line 538:
```suggestion
  /**
   * Get the geometric normal of the previously hit surface. When using normal maps, this is the normal of the geometry that was hit, not taking the normal map into account.
   */
  public Vector3 getGeometryNormal() {
```

leMaik (approved):
Not sure about the abbreviations, but yeah… I'd merge it :+1:

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
aTom3333: It was intentional because it makes more sense (I think) like this. If you think as normal mapping simulating displaced geometry, you want to invert the whole normal when working with the underside of the surface.
But in practice, it didn't have any perverse effect, the worse being that technically if you looked at both side of some water surface, they would not correspond to each other

PR #938, opened 2021-05-13
Add support for Minecraft 1.12 or older worlds

### LINKED ISSUE #553: Support 1.12 and 1.13+ worlds in the same Chunky version
A lot of things changed in _the flattening_ and the new updated shiny version of Chunky doesn't support 1.12 and older worlds anymore. That leads to inconsistencies (1.12 Chunky doesn't have the new UI) and bug fixes not being applied to both versions.

By abstracting away from the world format and refactoring some things, it should be possible to open 1.12 _and_ 1.13+ worlds in the same Chunky version one day.

leMaik:
Closes #553 

This PR adds support for legacy (i.e. 1.12 or lower) worlds to Chunky. It also fixes some bugs the blocks had in Chunky 1.4.5 and adds some features for legacy blocks that were not in 1.4.5.

_I'll add a list of changes to legacy blocks vs. 1.4.5 here later._

The way this works is that we map the legacy blocks to 1.13+ block tags while loading the chunks. This doesn't work for blocks that depend on the surrounding blocks (e.g. fence) or blocks that used to have tile entities but don't have them anymore (e.g. bed). These blocks get wrapped into an `UnfinalizedLegacyBlock` class that is then called during octree finalization and can replace itself in the octree depending on the surrounding blocks (or just create a custom entity depending on data, which is what banners and wall banners do).

All block connections currently match Minecraft 1.12 behavior. Some connections were different in earlier versions. We could add a way to select the "target Minecraft version" in a future PR.

Known issues: Fire is not finalized because Chunky currently doesn't support that for 1.13+ worlds either.


chunky/src/java/se/llbit/chunky/block/legacy/UnfinalizedLegacyBlock.java
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Replace the current block in the given finalization state by the finalized version of this
+   * block. The state is used for context, i.e. surrounding blocks.
+   *
+   * @param state Current finalization state
+   */
+  public abstract void finalizeBlock(FinalizationState state);
+
+  protected static boolean hasName(Material material, String... names) {
```
alexhliu, line 84: Maybe remove this and replace with `LegacyBlockUtils.getName()`? I think its a bit more clear and simpler to use. Currently, only `Chest` and `PumpkinStem` use it.

chunky/src/java/se/llbit/chunky/block/legacy/blocks/Chest.java
```diff
@@ -0,0 +1,53 @@
...
+            createTag(data, data == 2 ? "left" : "right"));
+        return;
+      }
+    } else {
+      // facing north or south
+      if (hasName(state.getMaterial(0, 0, -1), name, "#legacy_" + name)) {
+        state.replaceCurrentBlock(
+            createTag(data, data == 5 ? "right" : "left"));
+        return;
+      } else if (hasName(state.getMaterial(0, 0, 1), name, "#legacy_" + name)) {
```
alexhliu, line 35: Replace `hasName()` with `LegacyBlockUtils.getName()`?

chunky/src/java/se/llbit/chunky/block/legacy/blocks/Chest.java
```diff
@@ -0,0 +1,53 @@
+package se.llbit.chunky.block.legacy.blocks;
+
+import se.llbit.chunky.block.FinalizationState;
+import se.llbit.chunky.block.legacy.LegacyBlocks;
+import se.llbit.chunky.block.legacy.UnfinalizedLegacyBlock;
+import se.llbit.nbt.CompoundTag;
+
+public class Chest extends UnfinalizedLegacyBlock {
```
alexhliu, line 8: Rename to `LegacyChest` to be more clear?

chunky/src/java/se/llbit/chunky/block/legacy/blocks/MelonStem.java
```diff
@@ -0,0 +1,35 @@
+package se.llbit.chunky.block.legacy.blocks;
```
alexhliu, line 1: Rename to `LegacyMelonStem`?

chunky/src/java/se/llbit/chunky/block/legacy/blocks/PumpkinStem.java
```diff
@@ -0,0 +1,35 @@
+package se.llbit.chunky.block.legacy.blocks;
```
alexhliu, line 1: Rename to `LegacyPumpkinStem`?

chunky/src/java/se/llbit/chunky/block/legacy/blocks/SnowCoverableBlock.java
```diff
@@ -0,0 +1,34 @@
+package se.llbit.chunky.block.legacy.blocks;
```
alexhliu, line 1: Rename to `LegacySnowCoverableBlock`? or something to that effect?

chunky/src/java/se/llbit/chunky/block/legacy/blocks/Vine.java
```diff
@@ -0,0 +1,30 @@
+package se.llbit.chunky.block.legacy.blocks;
```
alexhliu, line 1: Rename to `LegacyVine`?

chunky/src/java/se/llbit/chunky/block/legacy/blocks/PumpkinStem.java
```diff
@@ -0,0 +1,35 @@
...
+
+  public PumpkinStem(String name, CompoundTag tag) {
+    super(name, tag);
+  }
+
+  @Override
+  public void finalizeBlock(FinalizationState state) {
+    // pumpkin stem points to adjacent pumpkin or carved pumpkin (but not jack-o-lantern)
+    for (BlockFace side : sides) {
+      if (hasName(state.getMaterial(side), "minecraft:pumpkin", "minecraft:carved_pumpkin")) {
```
alexhliu, line 23: Remove `hasName()` and replace with `LegacyBlockUtils.getName()` and `.equals` or just `.endsWith`?

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
```diff
@@ -135,6 +135,12 @@ public Block get(int id) {
     return palette.get(id);
   }
 
+  public Tag getTag(int id) {
+    return blockMap.entrySet().stream().filter(item -> item.getValue() == id)
+        .map(item -> item.getKey().getTag())
+        .findFirst().orElse(null);
```
alexhliu, line 141: Replace with a `foreach` loop? This stream is very hard to comprehend.

alexhliu (commented):
`UnfinalizedLegacyBlock.hasName` can probably be removed. The older legacy blocks should probably get a `Legacy` prefix to help differentiate them better from the non-legacy blocks.

chunky/src/java/se/llbit/chunky/block/BlockFace.java
```diff
@@ -0,0 +1,63 @@
...
+      case "south":
+        return SOUTH;
+      case "west":
+        return WEST;
+      case "up":
+        return UP;
+      case "down":
+        return DOWN;
+      default:
+        throw new IllegalArgumentException("Unknown BlockFace name: " + name);
```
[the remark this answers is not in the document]
leMaik: `fromName` is only used once currently… Should I still add a `static Map<String, BlockFace>`?

chunky/src/java/se/llbit/chunky/block/legacy/UnfinalizedLegacyBlock.java
leMaik: Yeah… when I added this I thought it would be used more (it was before you added `LegacyBlockUtils.getName`)

chunky/src/java/se/llbit/chunky/block/BlockFace.java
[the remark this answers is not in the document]
alexhliu: Sorry misread it. It's probably fine as is.

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
leMaik: Is it? It's just a simple filter/map, maybe I'm just more used to this stuff… I find it way easier to read than a for loop

alexhliu: I find it challenging to understand, I think a simple `foreach` and `if` would be very obvious as to what it does.

leMaik: Okay, changed it :+1:

alexhliu (approved):
Looks good 👍

PR #939, opened 2021-05-13
 Fix water artifact when looking at water on the horizon/with low incidence angle (bis)

aTom3333:
This is #935 but on top of #901.
(see #935 for explanation)


leMaik:
This was included in #901

PR #940, opened 2021-05-13
Save the window position/size on close and restore that on startup

aTom3333:
Small quality of life feature (maybe more for me that for real users).
I was annoyed with chunky starting maximized every time I start it (often while debugging) when I don't want it to cover the whole screen.
I was not able to test this with multiple screen, if someone can try that, I'd appreciate it. I'm especially worried about the situation where the chunky window is put on the second screen, close (saving its position), the second screen is unplugged and chunky reopened, I don't know where chunky would go in this situation

### DIFF
--- a/lib/src/se/llbit/chunky/PersistentSettings.java
+++ b/lib/src/se/llbit/chunky/PersistentSettings.java
@@ -20,6 +20,7 @@ import java.io.File;
 
 import se.llbit.chunky.renderer.RenderConstants;
 import se.llbit.chunky.resources.SettingsDirectory;
+import se.llbit.fxutil.WindowPosition;
 import se.llbit.json.JsonArray;
 import se.llbit.json.JsonValue;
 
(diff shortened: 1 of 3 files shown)


leMaik:
> I'm especially worried about the situation where the chunky window is put on the second screen, close (saving its position), the second screen is unplugged and chunky reopened, I don't know where chunky would go in this situation

At least on elementary OS (Xorg) it opens on the "other" screen with the size/maximized state it was closed with on the unplugged screen. So this "just works", not sure about other platforms.

leMaik (approved):
Looked good, I just moved the settings handling into the settings class and the restore and check logic into a new `WindowPosition` class.

@aTom3333 are the changes okay? :upside_down_face:

lib/src/se/llbit/chunky/PersistentSettings.java
```diff
@@ -85,7 +86,7 @@ private PersistentSettings() {
     changeSettingsDirectory(directory);
   }
 
-  private static void save() {
+  public static void save() {
```
leMaik, line 89: I swear I've seen this change in at least three PRs now :rofl:

aTom3333:
I can't approve via github as it is my own PR but this good for me

PR #943, opened 2021-05-15
Fix clipping when using parallel projection

### LINKED ISSUE #942: Parallel Projection front plane clipping
When using projection mode "Parallel", depending on camera rotation and FoV, parts of the scene near the camera get clipped resulting in black spots in the final render or an unidentifiable mess in the preview renderer.

Example render where mountains get clipped:
image
Same scene in preview mode:
image

aTom3333:
Fix #942

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -681,16 +681,23 @@ public class Scene implements JsonSerializable, Refreshable {
 
     if(camera.getProjectionMode() == ProjectionMode.PARALLEL
       && worldOctree.isInside(state.ray.o)) {
+      // When in parallel projection, push the ray origin back so the
+      // ray start outside the octree to prevent ray spawning inside some blocks
       int limit = (1 << worldOctree.getDepth());
       Vector3 o = state.ray.o;
       Vector3 d = state.ray.d;
       double t = 0;
+      // simplified intersection test with the 6 planes that form the bounding box of the octree
       t = Math.min(t, -o.x/d.x);
       t = Math.min(t, (limit-o.x)/d.x);
       t = Math.min(t, -o.y/d.y);
       t = Math.min(t, (limit-o.y)/d.y);
       t = Math.min(t, -o.z/d.z);
       t = Math.min(t, (limit-o.z)/d.z);
+      // set the origin to the farthest intersection point behind
+      // I theory, we only would need to set it to the closest intersection point behind
+      // but this doesn't because matter because the Octree.enterOctree function
+      // will do the same amount of math for the same result no matter what the exact point is
       o.scaleAdd(t, d);
     }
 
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -678,6 +679,28 @@ public void rayTrace(RayTracer rayTracer, WorkerState state) {
...
+      // simplified intersection test with the 6 planes that form the bounding box of the octree
+      t = Math.min(t, -o.x/d.x);
+      t = Math.min(t, (limit-o.x)/d.x);
+      t = Math.min(t, -o.y/d.y);
+      t = Math.min(t, (limit-o.y)/d.y);
+      t = Math.min(t, -o.z/d.z);
+      t = Math.min(t, (limit-o.z)/d.z);
+      // set the origin to the farthest intersection point behind
+      // I theory, we only would need to set it to the closest intersection point behind
+      // but this doesn't because matter because the Octree.enterOctree function
```
leMaik, line 699:
```suggestion
      // but this doesn't matter because the Octree.enterOctree function
```

leMaik:
Two bugs with this, I'm afraid…

1. Create a new scene, change projection mode to "Parallel" → Renderer freezes (and doesn't recover even when changing the projection mode back)
2. Create new scene, change camera orientation and _then_ look down → World is invisible (depending on the camera rotation)

aTom3333:
Oops, probably division by 0 causing havoc, will look at that

ShirleyNekoDev:
Bug 1 is not from this PR - renderer freeze is on master

aTom3333:
Nah, there was indeed an additional bug in this PR that caused freezing too.
Forgot about bug 2 but I can't reproduce, are you sure the scene is not behind the camera?

leMaik:
@aTom3333 You are right, bug 2 happened because my camera was below zero. The first bug is still there, but not on master.

leMaik:
@ShirleyNekoDev 
image

Works fine on master, breaks with this PR

ShirleyNekoDev:
Thank you for you explanation, I found another bug (#945) with your description ;)

ShirleyNekoDev:
Btw. the problem behind the freeze (infinite loop) is the same loop exit condition:
https://github.com/chunky-dev/chunky/blob/e15ee28fabf3cade8e477659f1dc39f3c3382b9b/chunky/src/java/se/llbit/math/Octree.java#L528-L643

enterBlock has a while(true) loop which exits only when outside of the octree or a block was found
But because we have x,y,z = 0 the octree check is always false (( lx != 0 || ly != 0 || lz != 0 with lx = x >>> depth...)

aTom3333:
> The first bug is still there, but not on master.
Even with the division by 0 fix?

leMaik:
@aTom3333 Yes, maybe related to this? https://github.com/chunky-dev/chunky/issues/945#issue-892476115

Freeze:
image

aTom3333:
Oh I just reproduced it, I can investigate

aTom3333:
I think @ShirleyNekoDev was right, this bug is also there on master, the bug happens even when the new code inside the branch is not taken (and the branch conditon is read only so it can't have an impact either). The bug is in Octree but the fix might be tricky so I'll open another PR

Issue #945, opened 2021-05-15
(Preview) Renderer freezes on "New scene from selection" with empty selection


ShirleyNekoDev:
Tested in Chunky 2.3.0, 2.4.0-snapshot, master

Camera defaults orientation points down (-90 0 0) which results in an infinite loop in enterBlock here:
https://github.com/chunky-dev/chunky/blob/e15ee28fabf3cade8e477659f1dc39f3c3382b9b/chunky/src/java/se/llbit/math/Octree.java#L528-L643

Variables at the end of the loop:
```
x=0, y=0, z=0, depth=1,
lx=0, ly=0, lz=0,
t=NaN, distance=+Inf, invDx=+Inf
```

The while(true) loop never ends because no block can be found that can be entered, but at the same time we never went outside of the octree (lx, ly, lz = 0 because we shift 0).

leMaik:
I'd say this is a UI bug. How about this?

![image](https://user-images.githubusercontent.com/5544859/118369792-0f8c0d00-b5a5-11eb-841e-e203d588aa8f.png)

ShirleyNekoDev:
It is possible to create an empty scene when switching directly to render preview without selecting chunks in the map.
I use this for testing custom entities for plugins ;)

PR #948, opened 2021-05-15
Fix the freeze when looking down and  in parallel projection

aTom3333:
Explanation is in the comment :D

### DIFF
--- a/chunky/src/java/se/llbit/math/Octree.java
+++ b/chunky/src/java/se/llbit/math/Octree.java
@@ -440,17 +440,17 @@ public class Octree {
 
     // If we have 0 < ray.o.x < octree_size, meaning the ray is correctly aligned in
     // x to intersect with the octree, [tXMin ; tXMax] will have the values
-    // [-Infinity ; Infinity] meaning its union with the interval [tMin ; tMax]
+    // [-Infinity ; Infinity] meaning its intersection with the interval [tMin ; tMax]
     // will keep the value [tMin ; tMax].
 
     // On the other hand if ray.o.x < 0 or if ray.o.x > octree_size then both
     // tXMin and tXMax will have the value -Infinity (resp. +Infinity)
-    // Meaning the interval [tXMin ; tXMax] (and [tMin ; tMax] once it is updated to be the union)
+    // Meaning the interval [tXMin ; tXMax] (and [tMin ; tMax] once it is updated to be the intersection)
     // will be reduced to a single values: -Infinity (resp. +Infinity)
     // (this description is not mathematically rigorous as infinity is not really a value
     // but is enough here).
     // As ray.d is not a null vector, at least one of its component is not 0
-    // and will have an associated interval with finite bounds, meaning the union of those interval
+    // and will have an associated interval with finite bounds, meaning the intersection of those interval
     // will give an empty set (in practice this is the condition
     // `if ((tMin > tXMax) || (tXMin > tMax)) return false;`) meaning no intersection is possible.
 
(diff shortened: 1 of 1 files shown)


leMaik:
This fixes the regression I was seeing in #943

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -425,26 +425,79 @@ public boolean isInside(Vector3 o) {
...
+
+    // On the other hand if ray.o.x < 0 or if ray.o.x > octree_size then both
+    // tXMin and tXMax will have the value -Infinity (resp. +Infinity)
+    // Meaning the interval [tXMin ; tXMax] (and [tMin ; tMax] once it is updated to be the union)
+    // will be reduced to a single values: -Infinity (resp. +Infinity)
+    // (this description is not mathematically rigorous as infinity is not really a value
+    // but is enough here).
+    // As ray.d is not a null vector, at least one of its component is not 0
+    // and will have an associated interval with finite bounds, meaning the union of those interval
+    // will give an empty set (in practice this is the condition
```
leMaik, line 454: A union with an interval never gives an empty set. Do you mean _intersection_ instead of _union_ here? Same for the first paragraph? :thinking:

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -425,26 +425,79 @@ public boolean isInside(Vector3 o) {
...
+    // `if ((tMin > tXMax) || (tXMin > tMax)) return false;`) meaning no intersection is possible.
+
+    // Lastly the other edge case (literally) is when we have ray.o.x == 0 or
+    // ray.o.x == octree_size, meaning the ray is right on the edge and is
+    // neither really outside nor inside.
+    // In this case tXMin or tXMax will be NaN (and the other will be +/- Infinity but
+    // that will not pose any problem as seen previously).
+    // Every comparison involving a NaN return false, this means that when doing the union
```
leMaik, line 462:
```suggestion
    // Every comparison involving a NaN returns false, this means that when doing the intersection
```

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -425,26 +425,79 @@ public boolean isInside(Vector3 o) {
...
+    // with [tMin ; tMax] (in practice the `if (tXMin > tMin) tMin = tXMin;` and
+    // `if (tXMax < tMax) tMax = tXMax;`), the interval will not be changed, acting
+    // as if the [tXMin ; tXMax] interval was [-Infinity ; +Infinity] and not disrupting
+    // following computations.
+    // Additionally, given how the condition to test whether the union will be empty before
+    // updating it is written, (the `if ((tMin > tXMax) || (tXMin > tMax))`), it will
+    // evaluate to false and not exit the function. This means that the ray is
+    // considered to be intersecting with the octree (if the other coordinate fulfill the condition
```
leMaik, line 470:
```suggestion
    // considered to be intersecting with the octree (if the other coordinate fulfills the condition
```

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -425,26 +425,79 @@ public boolean isInside(Vector3 o) {
...
+    // to be intersecting as well).
+    // If instead we would like rays on the edge of the octree not be considered intersecting,
```
leMaik, line 472:
```suggestion
    // If instead we would like rays on the edge of the octree not to be considered intersecting it,
```

leMaik (approved):

leMaik:
Intersection: [-20; 10] ∩ [-5; 20] = [-5; 10]
Union: [-20; 10] ∪ [-5; 20] = [-20; 20]

chunky/src/java/se/llbit/math/Octree.java
aTom3333: You're right I mixed up, it's intersection everywhere

ShirleyNekoDev:
This also fixes the issue with any projection described in #945 🎉

PR #952, opened 2021-05-16
Add a way for plugin to save data in a scene

aTom3333:
Add a json field in the scene that can be used by plugin to store data that relates to a scene

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -3373,17 +3373,38 @@ public class Scene implements JsonSerializable, Refreshable {
 
   /**
    * Add additional data
+   * Additional data is not used by chunky but can be used by plugins
    */
   @PluginApi
-  public void addData(String name, JsonValue value) {
+  public void setData(String name, JsonValue value) {
     additionalData.add(name, value);
   }
 
   /**
    * Retrieve additional data
+   * Additional data is not used by chunky but can be used by plugins
    */
   @PluginApi
   public JsonValue getData(String name) {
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -3358,4 +3370,20 @@ public void setAnimationTime(double animationTime) {
   public double getAnimationTime() {
     return animationTime;
   }
+
+  /**
+   * Add additional data
+   */
+  @PluginApi
+  public void addData(String name, JsonValue value) {
```
leMaik, line 3378:
```suggestion
  public void setData(String name, JsonValue value) {
```

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -3358,4 +3370,41 @@ public void setAnimationTime(double animationTime) {
...
+   */
+  @PluginApi
+  public JsonObject getAdditionalData() {
+    return additionalData;
+  }
+
+  /**
+   * Sets the whole additional data object
+   * Additional data is not used by chunky but can be used by plugins
+   * Prefer using `setData` to not remove the additional of another plugin
```
leMaik, line 3404: Oh, I'm sorry, my comment was bad. I didn't mean to add methods to get/set the whole object but just rename `getData` to something that makes clear what "data" is instead. Not sure if replacing the entire data is useful to plugins… maybe it is?

```suggestion
   * Prefer using {@link #setData} to not remove the additional of another plugin
```

leMaik (approved):

PR #953, opened 2021-05-16
Handle unsupported post-processors more gracefully when exporting a tiff

leMaik:
This logs a warning and saves without post-processing.

I also moved the default post-processor into the scene because it's more of a default value for a scene than something related to the post-processing itself.

### DIFF
--- a/chunky/src/java/se/llbit/tiff/TiffFileWriter.java
+++ b/chunky/src/java/se/llbit/tiff/TiffFileWriter.java
@@ -16,17 +16,17 @@
  */
 package se.llbit.tiff;
 
+import java.io.DataOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
-
+import java.io.IOException;
+import java.io.OutputStream;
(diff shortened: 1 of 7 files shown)


chunky/src/java/se/llbit/tiff/TiffFileWriter.java
```diff
@@ -233,7 +237,8 @@ public void write32(Scene scene, TaskTracker.Task task) throws IOException {
       task.update(height, y);
       for (int x = 0; x < width; ++x) {
         double[] pixel = new double[3];
-        scene.postProcessPixel(x, y, pixel);
+        ((PixelPostProcessingFilter) filter)
+            .processPixel(width, height, scene.getSampleBuffer(), x, y, scene.getExposure(), pixel);
```
aTom3333, line 241: Maybe remove the Scene.postProcessPixel function as I think this was the only call site

aTom3333 (approved):

PR #954, opened 2021-05-16
Move bulk of chunk loading out of Scene

Derongan:
Refactor to simplify the responsibilities of the scene class.

* Moved the bulk of the logic within loadChunks to a dedicated ChunkLoader class which performs chunk loading and returns data back in a wrapper object to scene.
* Move chunk reading into dedicated class that abstracts away the async read write.
* Move shared code from `calculateOctreeOrigin` and `calcCenterCamera` to separate method.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/ChunkLoader.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/ChunkLoader.java
@@ -1,3 +1,19 @@
+/* Copyright (c) 2021 Chunky contributors
+ *
+ * This file is part of Chunky.
+ *
+ * Chunky is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Chunky is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with Chunky.  If not, see <http://www.gnu.org/licenses/>.
+ */
 package se.llbit.chunky.renderer.scene;
 
 import static java.lang.Math.min;
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/ChunkLoader.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/ChunkLoader.java
@@ -1,0 +1,604 @@
+package se.llbit.chunky.renderer.scene;
+
+import static java.lang.Math.min;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Executors;
+import se.llbit.chunky.PersistentSettings;
+import se.llbit.chunky.block.Air;
+import se.llbit.chunky.block.Block;
+import se.llbit.chunky.block.Lava;
+import se.llbit.chunky.block.Water;
+import se.llbit.chunky.chunk.BlockPalette;
+import se.llbit.chunky.chunk.GenericChunkData;
+import se.llbit.chunky.chunk.SimpleChunkData;
+import se.llbit.chunky.entity.ArmorStand;
+import se.llbit.chunky.entity.Entity;
+import se.llbit.chunky.entity.Lectern;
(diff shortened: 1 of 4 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -17,11 +17,17 @@
  */
 package se.llbit.chunky.renderer.scene;
 
+import static java.lang.Math.max;
+import static java.lang.Math.min;
```
leMaik, line 21: Looking good so far but please don't do this

chunky/src/test/se/llbit/chunky/renderer/scene/ChunkReaderTest.java
```diff
@@ -0,0 +1,186 @@
...
+      mockChunk = new MockChunk(this);
+    }
+
+    @Override
+    public synchronized Chunk getChunk(ChunkPosition pos) {
+      return mockChunk;
+    }
+  }
+
+  private static final class MockChunkData implements ChunkData {
```
leMaik, line 108: Instead of micking empty chunk data you could use `EmptyChunkData`. In the future we could maybe even mock some water scenarios (or load _actual_ chunks) and test if finalization works?

leMaik:
Nice first PR and step to make the scene class smaller. :rocket: :+1:

chunky/src/test/se/llbit/chunky/renderer/scene/ChunkReaderTest.java
Derongan: EmptyChunkData doesn't provide a public constructor and instead provides a single instance. I want to assert that the two chunk data objects are properly being swapped back and forth, so I need different instances.

chunky/src/java/se/llbit/chunky/renderer/scene/ChunkReader.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @Override
+    public int minY() {
+      return chunkData.minY();
+    }
+
+    @Override
+    public int maxY() {
+      return 0;
```
aTom3333, line 134: typo over here, you want to return chunkdata.maxY()

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1424,47 +918,63 @@ public void rebuildActorBvh() {
...
+      return new Vector3i(bounds.xmin - xroom / 2, -yroom / 2, bounds.zmin - zroom / 2);
     } else {
       // Note: Math.floorDiv rather than integer division for round toward -infinity
-      origin.set(xmin, Math.floorDiv(yMin, 16) * 16, zmin);
+      return new Vector3i(bounds.xmin, Math.floorDiv(yMin, 16) * 16, bounds.zmin);
     }
-    return requiredDepth;
+  }
+
+  static int calculateOctreeRequiredDepth(BlockBounds bounds, int yMax, int yMin) {
```
leMaik, line 975: This could be moved into the `Octree` class

chunky/src/java/se/llbit/chunky/renderer/scene/ChunkLoader.java
```diff
@@ -0,0 +1,620 @@
...
+      TaskTracker taskTracker, boolean loadActors) {
+    BlockBounds bounds = Scene.calculateBounds(chunksToLoad);
+    int requiredDepth = Scene.calculateOctreeRequiredDepth(bounds, yMax, yMin);
+
+    ChunkLoadResult result = new ChunkLoadResult(octreeImplementation, requiredDepth,
+        emitterSamplingStrategy, gridSize);
+
+    try (Task task = taskTracker.task("(1/6) Loading regions")) {
+      task.update(2, 1);
+      //TODO is this necessary? Are we intentionally mutating origin here?
```
leMaik, line 145: Yes it is… also you're this now mutates the *result* of this method, so that's fine.

leMaik (changes requested):
Looks good so far, maybe we could move more stuff and get rid of the now package-internal methods (and move them into appropriate classes if there are reusable enough or make them private otherwise)?

chunky/src/java/se/llbit/chunky/renderer/scene/ChunkLoader.java
Derongan: Wups, meant to follow up in the review not leave as a TODO.

Ok, good to know it was intentional. Since the original mutating method was used in two places, one where the origin change was used immediately and here where only the depth info was used, I was concerned it was a minor bug causes by all of the lovely lovely side effects.

I'll remove the TODO

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
Derongan: Good point

PR #956, opened 2021-05-19
Add strategy interfaces for perfomring emitter sampling and fog rendering. Decouple logic from PathTracer and PreviewTracer

Derongan:
This PR modifies how the PathTracer handles emitter sampling and fog. Instead of directly implementing fog/sampling within a ray tracer, delegate to a Sampler or FogStrategy.

Additionally I have moved static methods defined in PreviewRayTracer to a utility `RayTracers` class to untangle tracer types.

TODO

- [ ] Verify that the minor increase in method calls does not hurt rayTrace performance. JIT should handle this just fine, but have to be sure.
- [ ] Allow plugins to provide custom fog handling strategies
- [ ] Better differentiate between skyFog and er not sky fog? settings in UI and allow choosing a fog strategy

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/WorkerState.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/WorkerState.java
@@ -16,16 +16,13 @@
  */
 package se.llbit.chunky.renderer;
 
-import se.llbit.math.Ray;
-import se.llbit.math.Vector4;
-
 import java.util.Random;
+import se.llbit.math.Ray;
 
 /**
  * State for a render worker.
  */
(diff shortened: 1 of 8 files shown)


ShirleyNekoDev:
Hi there,
is this PR still active?
The current changes look good but I'm not sure about the open TODOs!
I'd suggest that we split this up and create PRs for each change (`RayTracers` refactor, fog, sampling), to get them merged early :)

Derongan:
I'm still alive but got fairly busy with other things. Just recently been back in chunky so I may have some time to check some off

PR #957, opened 2021-05-19
Reduce memory usage of ChunkTexture

aTom3333:
Back with my memory usage obsession, this time I changed ChunkTexture to be stored as byte instead of float and doing the conversion to linear on the fly.
I think I didn't touch it last time I changed several classes because iirc NotStirred wanted to change how it's done. But  WorldTexture (due to ChunkTexture) is now the biggest object in memory (when using optimized octrees) and the change is simple that it's worth it even if ChunkTexture is changed in a month (especially if this makes it to chunky 2.4 but not the changes NotStirred wanted to make).
I've not measured a performance drop due to this (but it was a small scale test).
The format on disk is unchanged.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/world/ChunkTexture.java
+++ b/chunky/src/java/se/llbit/chunky/world/ChunkTexture.java
@@ -16,6 +16,8 @@
  */
 package se.llbit.chunky.world;
 
+import se.llbit.math.ColorUtil;
+
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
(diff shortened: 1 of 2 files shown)


NotStirred:
+1 to this, the changes I would *like* to make are much more of a rewrite, and would be at least 2.5
For a ***very*** large world I actually used a 20GiB heap just loading the world texture (it was `51200*563200`)

leMaik (approved):
Looks good, and thanks to the lookup table this probably doesn't affect rendering performance too much.

PR #959, opened 2021-05-20
Texture image compression

aTom3333:
Another memory usage reduction PR.
This time I changed Texture to use a generic `Image` interface instead of directly `BitmapImage` and I made a another implementation of this interface that is compressed with a palette.
Of course the compressed implementation is slower to access pixel so the idea is to only use it for rarely accessed texture. What I've done is to use it for texture that are mainly related to player/armor stand like equipment and player head.
With this the majority of texture access are still fast but some memory is saved.
When loading 4 region of Greenfield I have:
before: retained size of `BitmapImage` is 108MB
after: retained size of `BitmapImage`+`CompressedImage` is 69MB (63MB + 6MB).
That's around a 10MB reduction per region for Greenfield. (of course, a map without entities wouldn't see improvements).
I have not measured any performance drop (but small scale test as well while the camera is far away from the ground to get a bird eye view so not a lot of rays must have intersected with entities).

### DIFF
--- a/chunky/src/java/se/llbit/chunky/resources/CompressedImage.java
+++ b/chunky/src/java/se/llbit/chunky/resources/CompressedImage.java
@@ -5,6 +5,11 @@ import it.unimi.dsi.fastutil.ints.IntArrayList;
 import it.unimi.dsi.fastutil.longs.LongArrayList;
 import se.llbit.math.QuickMath;
 
+/**
+ * Image using a palette compression to reduce memory usage.
+ * The size of the palette/number of bit each pixel takes
+ * is computed based on the number of colors used in the image to compress
+ */
 public class CompressedImage implements Image {
   private int width;
   private int height;
@@ -51,6 +56,12 @@ public class CompressedImage implements Image {
     return new BitmapImage(width, height, uncompressedData);
   }
 
+  /**
+   * Do the compression.
+   * The input image is first analyzed to count the number of color and determine
+   * the bit depth of the image.
+   * Then the image is compressed by storing the palette indexes
+   */
   private void compressFrom(Image source) {
     IntArrayList paletteBuilder = new IntArrayList();
     Int2IntOpenHashMap colorToIndex = new Int2IntOpenHashMap();
@@ -78,6 +89,7 @@ public class CompressedImage implements Image {
     pixelPerLong = 64 / bitDepth;
     mask = ((1L << bitDepth) - 1); // bitDepth 1s
 
+    // build the image
     LongArrayList compressedDataBuilder = new LongArrayList();
 
     int pixelIndex = 0;
--- a/chunky/src/java/se/llbit/chunky/resources/CompressedImage.java
+++ b/chunky/src/java/se/llbit/chunky/resources/CompressedImage.java
@@ -1,0 +1,96 @@
+package se.llbit.chunky.resources;
+
+import it.unimi.dsi.fastutil.ints.Int2IntOpenHashMap;
+import it.unimi.dsi.fastutil.ints.IntArrayList;
+import it.unimi.dsi.fastutil.longs.LongArrayList;
+import se.llbit.math.QuickMath;
+
+public class CompressedImage implements Image {
+  private int width;
+  private int height;
+  private int[] palette;
+  private int bitDepth;
+  private int pixelPerLong;
+  private long[] compressedData;
+  private long mask;
+
+  CompressedImage(Image source) {
+    compressFrom(source);
+  }
+
+  @Override
+  public int getWidth() {
+    return width;
+  }
+
+  @Override
+  public int getHeight() {
+    return height;
+  }
+
+  @Override
+  public int getPixel(int x, int y) {
+    return getPixel(y * width + x);
+  }
+
+  @Override
+  public int getPixel(int index) {
+    int longIndex = index / pixelPerLong;
+    int indexInLong = index % pixelPerLong * bitDepth;
+    int paletteIndex = (int) ((compressedData[longIndex] >>> indexInLong) & mask);
+    return palette[paletteIndex];
+  }
+
+  @Override
+  public BitmapImage asBitmap() {
+    int pixelCount = width * height;
+    int[] uncompressedData = new int[pixelCount];
+    for(int i = 0; i < pixelCount; ++i) {
+      uncompressedData[i] = getPixel(i);
+    }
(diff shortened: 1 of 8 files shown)


chunky/src/java/se/llbit/chunky/resources/Image.java
```diff
@@ -0,0 +1,31 @@
+package se.llbit.chunky.resources;
+
+/**
+ * Generic image giving read only access
+ */
+public interface Image {
```
leMaik, line 6: 👍 for this interface!
Maybe we could reduce it even more so that `Texture` can implement it too?

chunky/src/java/se/llbit/chunky/resources/Image.java
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * Get the pixel at the given index. Index is computed as y * width + x
+   */
+  int getPixel(int index);
+
+  /**
+   * Convert the image to a BitmapImage
+   */
+  BitmapImage asBitmap();
```
leMaik, line 30: `@PluginApi`?

chunky/src/java/se/llbit/chunky/resources/Image.java
leMaik: Nope, bad idea. A texture might contain multiple textures internally later (for normal maps, specular maps, metalness maps).

chunky/src/java/se/llbit/chunky/resources/Image.java
aTom3333: I did not really think of this interface as being used/useful for plugin

leMaik: The OpenCL plugin needs to get the textures somehow. I don't think any other plugin needs it so maybe we could not make it official api. :-)

aTom3333: Oh, in this case it should probably rather be Texture.getBitmap that is exposed

chunky/src/java/se/llbit/chunky/resources/Image.java
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * Get the pixel at position (x, y)
+   */
+  int getPixel(int x, int y);
+
+  /**
+   * Get the pixel at the given index. Index is computed as y * width + x
+   */
+  int getPixel(int index);
```
leMaik, line 25: Should we drop this method from the `Image` interface? The usages of it could as well use the `getPixel(x, y)`, which would even simplify the code.

chunky/src/java/se/llbit/chunky/resources/Texture.java
```diff
@@ -1050,28 +1049,27 @@ public Texture(String resourceName) {
...
-    width = image.width;
-    height = image.height;
+    width = image.getWidth();
+    height = image.getHeight();
     float[] pixelBuffer = new float[4];
     for (int y = 0; y < height; ++y) {
       for (int x = 0; x < width; ++x) {
         int index = width * y + x;
-        ColorUtil.getRGBAComponentsGammaCorrected(data[index], pixelBuffer);
+        ColorUtil.getRGBAComponentsGammaCorrected(image.getPixel(index), pixelBuffer);
```
leMaik, line 1072:
```suggestion
        ColorUtil.getRGBAComponentsGammaCorrected(image.getPixel(x, y), pixelBuffer);
```

chunky/src/java/se/llbit/chunky/resources/Texture.java
```diff
@@ -1134,7 +1132,7 @@ public void getColor(Ray ray) {
     if(useAverageColor)
       return avgColorFlat;
     float[] result = new float[4];
-    ColorUtil.getRGBAComponentsGammaCorrected(image.data[width*y + x], result);
+    ColorUtil.getRGBAComponentsGammaCorrected(image.getPixel(width*y + x), result);
```
leMaik, line 1135:
```suggestion
    ColorUtil.getRGBAComponentsGammaCorrected(image.getPixel(x, y), result);
```

chunky/src/java/se/llbit/chunky/resources/Texture.java
```diff
@@ -1213,19 +1211,22 @@ public boolean isEmptyTexture() {
...
   }
 
-  public BitmapImage getBitmap() {
-    return image;
+  /**
+   * Replace the current image by a compressed version of the image
+   */
+  public void compress() {
+    if(!(image instanceof CompressedImage))
+      image = new CompressedImage(image);
```
leMaik, line 1230: When would we not want to compress a texture? Should this just be done in `setTexture`?

chunky/src/java/se/llbit/chunky/resources/CompressedImage.java
```diff
@@ -0,0 +1,108 @@
...
+ * The size of the palette/number of bit each pixel takes
+ * is computed based on the number of colors used in the image to compress
+ */
+public class CompressedImage implements Image {
+  private int width;
+  private int height;
+  private int[] palette;
+  private int bitDepth;
+  private int pixelPerLong;
+  private long[] compressedData;
```
leMaik, line 19: is `int[]` without shifting logic faster? Textures aren't _that_ big in Minecraft (and not even with high-res resourcepacks)

chunky/src/java/se/llbit/chunky/resources/Texture.java
aTom3333: Reading from a compressed texture is slower. When every texture access is slower, this has an impact on the rendering speed. I think it is undesirable to compress frequently used textures. When only rarely used texture are compressed, the slow down is not noticable

chunky/src/java/se/llbit/chunky/resources/CompressedImage.java
aTom3333: Even with int shifts would be need because each pixel can take an arbitrary number of bits (but for a given image this number is the same for all pixels).
Additionally a pixel is never stored between 2 variables, this means there are some wasted bits when the number of bit per pixel is not a divisor of 64 (or 32 if using int). For example is there is 7 bits per pixels, we can store 4 pixels in a int with 4 bits wasted, so 8 pixels are stored in 8 bytes. If using long, we can store 9 pixels with only 1 wasted bit. Biger types are better in this case.

leMaik (approved):

PR #961, opened 2021-05-22
Cache TextureMaterial to prevent their proliferation 

aTom3333:
A number of entities need to create `TextureMaterial` to wrap their `Texture` when calling `primitives`. The issue is that a lot of those `TextureMaterial` are created from the same `Texture` meaning they are distinct instances that are identical.

I have added a `TextureMaterial` cache to ensure a single `TextureMaterial` instance exist for a given `Texture`, doing this divides the number of `TextureMaterial` instances by 10, saving memory.
Additionally, by not using the cache for `Texture`s that I know aren't shared (like the `SignTexture`, cf second commit), the cache is not filled with entries that won't bring any benefits and stay small.

To implement the cache I needed a hash map where the values were weak references (the built-in `WeakHashMap` has weak keys but strong values) so I added a new dependency on apache common collections. You decide if that is acceptable.

For 1 region of greenfield, I went from 90k `TextureMaterial` instances to 9k going from 5MB to 500kB.
For 4 regions of greenfield, I went from 160k `TextureMaterial` instances to 16k, going from 9MB to 900kB.

The additional memory overhead of the cache is was only of 35kB for 1 region of greenfield (didn't measure for 4 regions).

The savings don't seem to scale that much and aren't that big but that's probably still worth it.

Another benefit of reducing the number of instances of `Material` is for an hypothetical BVH implementation that wants to store the primitives compressed by replacing the material by an index into a palette of materials, having less instances of `Material` would reduce the size of this palette.


chunky/src/java/se/llbit/chunky/entity/StandingBanner.java
```diff
@@ -468,6 +468,6 @@ public static Material getBannerTexture(JsonObject design) {
     }
     Texture texture = new Texture(tinted);
     TextureCache.put(design, texture);
-    return new TextureMaterial(texture);
+    return TextureMaterial.getForTexture(texture);
```
leMaik, line 471: The texture is created two lines further up and will always be different and can't be cached this way.

chunky/src/java/se/llbit/chunky/world/material/TextureMaterial.java
```diff
@@ -16,11 +16,31 @@
...
+  public static TextureMaterial getForTexture(Texture texture) {
+    TextureMaterial mat = cache.get(texture);
+    if(mat == null) {
+      mat = new TextureMaterial(texture);
+      cache.put(texture, mat);
+    }
+    return mat;
+  }
+
   public TextureMaterial(Texture texture) {
```
leMaik, line 44: If you want to make sure that the cache is used, you could make this constructor private and check for sign textures by `instance of`` in `getForTexture`.

Good candidates for no caching would probably we skull textures, too.

chunky/build.gradle
```diff
@@ -8,6 +8,7 @@ archivesBaseName = 'chunky-core'
 dependencies {
   compile 'it.unimi.dsi:fastutil:8.4.4'
   compile 'org.apache.commons:commons-math3:3.2'
+  compile 'org.apache.commons:commons-collections4:4.4'
```
leMaik, line 11: Looks fine to me. 🤔 They don't have the primitive maps that we take from fastutil, right?

chunky/src/java/se/llbit/chunky/entity/StandingBanner.java
aTom3333: You're right, I replaced without thinking about it

chunky/src/java/se/llbit/chunky/world/material/TextureMaterial.java
aTom3333: There are some other uses of the constructor that I left untouched, mainly the ones where the `MaterialTexture` was static and always shared without help from the cache.

Will look at skull textures

chunky/build.gradle
aTom3333: No, they don't. 
One thing I forgot to say is that google guava provides the weak map as well but between the 2, I choose the smaller dependency. If someday guava is added (and nothing else from commons collection is used), this could be removed

chunky/src/java/se/llbit/chunky/world/material/TextureMaterial.java
aTom3333: If i'm reading the code correctly, custom skull textures are not supported, only the built-in steve (and monsters) so for now it makes sense to cache it

leMaik: Custom skull textures are supported since 2.3.0

aTom3333: Oh yeah sorry, I was reading SkullEntity instead of HeadEntity

leMaik:
The weak *keys* part… The existing `TextureCache` is buggy then :scream:

chunky/build.gradle
```diff
@@ -14,6 +14,7 @@ configurations {
 dependencies {
   implementation 'it.unimi.dsi:fastutil:8.4.4'
   implementation 'org.apache.commons:commons-math3:3.2'
+  implementation 'org.apache.commons:commons-collections4:4.4'
```
leMaik, line 17: Reminder to self: I'll have to add this library to the `latest.json` and `snapshot.json` and host it.

leMaik (approved):
This could now be extended to also cache the textures of hanging signs.

PR #965, opened 2021-05-23
Prevent normal being (0, 0, 0) which leads to lock in the renderer

### LINKED ISSUE #963: Renderer locks when clipping into a block while rendering
Steps to reproduce:

1. Start rendering
2. Fly around, hit blocks
3. At some point, the renderer locks up (infinite loop with high CPU usage)

@alexhliu found out that the ray direction is `NaN` when that happens.

aTom3333:
cf discussion on discord.
I set a value to the normal after the ray is created. The actual value doesn't matter, the only case where it isn't replaced is when the camera is inside a block and the screen will be black.
Fix #963


leMaik:
Thank you and @alexhliu for quickly debugging this and for the quick fix! :+1:

PR #968, opened 2021-05-24
Fix paintings with 1.14+ resource packs

### LINKED ISSUE #927: Paintings don't work with 1.14+ resourcepacks
Paintings textures were changed in 1.14 so they are not found by chunky anymore.

Instead of using a single texture with all paintings, the textures are now split into multiple files in `/assets/minecraft/textures/painting/`. The border and back of a painting is in `back.png`.

leMaik:
Fixes #927 

- Every painting now uses its own `Texture` and I added an adapter for loading <1.14 textures that crops the paintings from the atlas texture.
- The back texture is 1x1 block in 1.14+ but used to be 4x4 blocks. For ease of use in Chunky (triangles don't repeat textures), the new texture gets tiled so that it is a 4x4 blocks texture and can be used just like the old texture.
- I increased the frame thickness so that it matches the (1.16.5) ingame look, i.e. there is no space between the painting and the wall it is attached to and the painting has a thickness of 1/16th of a block.


chunky/src/java/se/llbit/chunky/entity/PaintingEntity.java
```diff
@@ -37,108 +39,105 @@
...
+    paintings.put("Void", new Painting(Texture.paintingVoid, 2, 2));
+    paintings.put("minecraft:void", new Painting(Texture.paintingVoid, 2, 2));
+    paintings.put("SkullAndRoses", new Painting(Texture.paintingSkullAndRoses, 2, 2));
+    paintings.put("minecraft:skull_and_roses", new Painting(Texture.paintingSkullAndRoses, 2, 2));
+    paintings.put("Wither", new Painting(Texture.paintingWither, 2, 2));
+    paintings.put("minecraft:wither", new Painting(Texture.paintingWither, 2, 2));
+    paintings.put("Fighters", new Painting(Texture.paintingFighters, 4, 2));
+    paintings.put("minecraft:fighters", new Painting(Texture.paintingFighters, 4, 2));
+    paintings.put("Skeleton", new Painting(Texture.skeleton, 4, 3));
+    paintings.put("minecraft:skeleton", new Painting(Texture.skeleton, 4, 3));
```
leMaik, line 128: :bug:

PR #976, opened 2021-05-25
Load entities checkbox

### LINKED ISSUE #896: Add button for Load entities
It would be nice to see a check box for `Load entities`, similar to the one currently used for `Load players`, allowing you to switch entities off during loading.

Stenodyon:
Resolves #896.
This also fixes a bug where the "This takes effect the next time a new scene is created" popup would show even when the user didn't click on the checkbox.


leMaik:
This is great :+1: But I wonder if it would be better to allow toggling it per entity, i.e. with some popover that shows up when clicking the button with a menu of checkboxes. Not sure how hard that would be

Stenodyon:
I'll see what I can do

Stenodyon:
How does this look ? Am I missing entity types ?
entities

leMaik:
Awesome! Way better than what i had in mind

Stenodyon:
Not all entity types are supported but I added the ones I saw most often in the entity tab. Support for additional entity types is easy to add by adding a checkbox to the menu and implementing the `shouldLoad` function on the corresponding entity type.

chunky/src/java/se/llbit/chunky/entity/ArmorStand.java
```diff
@@ -620,4 +621,9 @@ public static Entity fromJson(JsonObject json) {
   @Override public JsonObject getGear() {
     return gear;
   }
+
+  @Override
+  public boolean shouldLoad() {
+    return PersistentSettings.getLoadArmorStands();
```
aTom3333, line 627: I feel like this should not be a persistent setting but something saved in the scene. If I create a sceneA with entities, then create a sceneB without entities, then load sceneA, I want my entities to load.
Like other settings saved in the scene, the default value could still be from the persistent settings in addition.
Maybe this means you need to change the `shouldLoad` interface to take something as parameter (the whole feels overkill but is maybe the simplest if you just add some booleans to the scene, maybe do something a bit better organized and put those boolean in some kind of `EntityLoadingOption` class or I don't know)

chunky/src/res/se/llbit/chunky/ui/render/GeneralTab.fxml
```diff
@@ -1,18 +1,11 @@
...
-<?import javafx.scene.control.ScrollPane?>
-<?import se.llbit.chunky.ui.IntegerAdjuster?>
+<?import javafx.geometry.*?>
+<?import javafx.scene.control.*?>
+<?import javafx.scene.layout.*?>
+<?import se.llbit.chunky.ui.*?>
 
-<?import javafx.scene.layout.FlowPane?>
-<fx:root type="javafx.scene.control.ScrollPane" xmlns="http://javafx.com/javafx/8.0.40" xmlns:fx="http://javafx.com/fxml/1">
+<fx:root type="javafx.scene.control.ScrollPane" xmlns="http://javafx.com/javafx/11.0.10-internal" xmlns:fx="http://javafx.com/fxml/1">
```
aTom3333, line 8: I have no idea if/how the namespace is used by javafx but checked that this still works with java 8 give that you've replaced 8.something with 11.something. (or maybe we don't care because soon java will be bundled?)

chunky/src/java/se/llbit/chunky/entity/ArmorStand.java
Stenodyon: Aren't entities stored in the scene anyway? These checkboxes only affect the loading of entities when loading chunks, but not when loading a scene from disk.

chunky/src/res/se/llbit/chunky/ui/render/GeneralTab.fxml
Stenodyon: I'll revert these namespaces, just in case.

chunky/src/java/se/llbit/chunky/entity/ArmorStand.java
aTom3333: Yes they are but maybe I want to reload the chunks after reloading sceneA because I want to use emitter sampling or I changed something in my minecraft world or whatever. 
Maybe let's wait to see wait @leMaik thinks

chunky/src/res/se/llbit/chunky/ui/render/GeneralTab.fxml
leMaik: A name space that ends with -internal doesn't sound like we should use it regardless of the Java version 🤔

leMaik, line 8:
```suggestion
<fx:root type="javafx.scene.control.ScrollPane" xmlns="http://javafx.com/javafx/8.0.40" xmlns:fx="http://javafx.com/fxml/1">
```

chunky/src/java/se/llbit/chunky/entity/ArmorStand.java
leMaik: Just saw this now (after rebasing the PR)… I agree with @aTom3333 on this one (reloading the chunks of an existing scene shouldn't cause entities to disappear). 
Saving this in the scene also has the nice benefit that the scene will still work as expected when the octree is removed and it is opened.

Stenodyon:
I think I have something decent, however I'm getting a bug where the per-scene preferences are not respected when loading chunks for some reason? I feel like the preferences set in the UI are not propagated to the right Scene instance.

leMaik:
@Stenodyon Looks good :+1: You need to (shallow) copy the new property in [Scene#copyState](https://github.com/chunky-dev/chunky/blob/master/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java#L436), that should fix the issue

alexhliu:
I haven't looked too far in the code, but generally you need to have a `Scene.refresh()` to synchronize all the scenes (but this will reset the render, it's dumb I know). Octree and bvh have the same problem. I've already opened an issue about this.

leMaik:
Sorry it took so long, but it's finally merged. :tada: Thank you so much for your work, @Stenodyon! :clap:

PR #979, opened 2021-05-27
Prevent map view from attempting to render as fast as possible

NotStirred:
locked to 24fps, can make it configurable if requested
Here is greenfield 2021 example:
before
https://user-images.githubusercontent.com/16853282/119767438-b5495100-beae-11eb-9728-1285352979f7.mp4
after
https://user-images.githubusercontent.com/16853282/119767456-bed2b900-beae-11eb-932a-7ec4af3b73ea.mp4

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/ChunkMap.java
+++ b/chunky/src/java/se/llbit/chunky/ui/ChunkMap.java
@@ -350,7 +350,7 @@ public class ChunkMap implements ChunkUpdateListener, ChunkViewListener, CameraV
         mapBuffer.updateView(view);
         mapBuffer.redrawView(mapLoader, chunkSelection);
         repaintDeferred();
-      }, (1000000000/VIEW_MAX_UPDATES) - (System.nanoTime() - lastUpdate), TimeUnit.NANOSECONDS);
+      }, (1_000_000_000 / VIEW_MAX_UPDATES) - (System.nanoTime() - lastUpdate), TimeUnit.NANOSECONDS);
     }
   }
 
--- a/chunky/src/java/se/llbit/chunky/ui/ChunkMap.java
+++ b/chunky/src/java/se/llbit/chunky/ui/ChunkMap.java
@@ -107,7 +107,7 @@ public class ChunkMap implements ChunkUpdateListener, ChunkViewListener, CameraV
 
   private static final int VIEW_MAX_UPDATES = Math.max(Integer.parseInt(System.getProperty("chunky.mapviewfps", "30")), 1);
 
-  private volatile long lastUpdate = System.nanoTime();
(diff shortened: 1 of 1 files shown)


NotStirred:
For dragging the map around SUPER fast I can't notice it, but it's possible that moving the view really fast, to separate it from the actual position. This will of course fix itself on the next frame, unless you instantly stop the mouse

ShirleyNekoDev:
You could probably try to lock this to the display refresh rate for smoother panning (https://docs.oracle.com/javase/8/docs/api/java/awt/DisplayMode.html)

alexhliu:
A cleaner solution would be changing the map view from using `repaintDirect` to `repaintDeferred()` [here](https://github.com/chunky-dev/chunky/blob/881773c665a977d1a93f4ef1721e5f135ba642f2/chunky/src/java/se/llbit/chunky/ui/ChunkMap.java#L345). This will force the map to wait for the first repaint to finish, before attempting to repaint again (dropping repaint requests that occur while the map is not repainted yet).

NotStirred:
I see, though this still will attempt to repaint as fast as possible, just not cancelling a previous repaint to start the next.
Is this the behaviour we want? It works perfectly, though for a map view anything above 24/30fps is wasted cpu time that on a low end system would be better used actually computing the map tiles themselves

NotStirred:
A rate limited deferred repaint may be best

chunky/src/java/se/llbit/chunky/ui/ChunkMap.java
```diff
@@ -337,12 +340,18 @@ public void renderView(File targetFile, ProgressTracker progress) {
...
+      viewUpdateScheduled = true;
+      executor.schedule(() -> {
+        lastUpdate = System.nanoTime();
+        viewUpdateScheduled = false;
+
+        this.view = view;
+        mapBuffer.updateView(view);
+        mapBuffer.redrawView(mapLoader, chunkSelection);
+        repaintDeferred();
+      }, (1000000000/VIEW_MAX_UPDATES) - (System.nanoTime() - lastUpdate), TimeUnit.NANOSECONDS);
```
leMaik, line 353:
```suggestion
      }, (1_000_000_000 / VIEW_MAX_UPDATES) - (System.nanoTime() - lastUpdate), TimeUnit.NANOSECONDS);
```

(after counting the zeros twice before realizing that this is 10^9)

leMaik (approved):

leMaik:
After some discussion on Discord, the solution was to not over-think this and set the default fps to 30 for now (but make it adjustable). We can always revisit this in the future. :+1:

NotStirred:
adjustable by property `chunky.mapviewfps`

PR #983, opened 2021-05-28
Move none to the top of the post processor combobox and add a separator.

leMaik:
The post processors are sorted alphabetically but displaying none in the middle doesn't make sense.

This improves it to look like this:
image


chunky/src/java/se/llbit/chunky/ui/render/PostprocessingTab.java
```diff
@@ -97,4 +107,20 @@ public PostProcessingFilter fromString(String string) {
       controller.getCanvas().forceRepaint();
     });
   }
+
+  /**
+   * Fake post processing filter that is also a seperator for the combobox.
+   */
+  private static class PostprocessingSeparator extends Separator implements PostProcessingFilter {
```
leMaik, line 114: Note: This is only required to make it type-safe. We could just ignore the generic type of `getItems()` but that would require us to ignore warnings.

PR #990, opened 2021-06-02
Fix preview not updating under certain plugin circumstances.

alexhliu:
The preview callback was checking the wrong `Renderer`'s `auotPostProcess()` method.


leMaik:
4+ eyes looking over every PR and yet why don't catch this 👀

PR #994, opened 2021-06-03
Deduplicate WorldTextures to save a lot of memory.

alexhliu:
Some benchmarks:
Greenfield:
|    | Heap used |
|---|-----------|
| master | 4,718,981,222 bytes |
| deduped | 4,369,123,080 bytes |

4 regions of a real world:
|   | `WorldTexture` size | Number of `ChunkTexture`s |
|---|-------------------|----------------------------|
| master | 27,787,634 bytes |  36,864 textures |
| deduped | 2,271,314 bytes  | 1,786 textures|

### DIFF
--- a/chunky/src/java/se/llbit/chunky/world/WorldTexture.java
+++ b/chunky/src/java/se/llbit/chunky/world/WorldTexture.java
@@ -33,7 +33,7 @@ public class WorldTexture {
 
   private final Long2ObjectOpenHashMap<ChunkTexture> map = new Long2ObjectOpenHashMap<>();
 
-  private boolean endFinalized = false;
+  private boolean compacted = false;
 
   /**
    * Timestamp of last serialization.
@@ -47,10 +47,14 @@ public class WorldTexture {
    * @throws IllegalStateException if this WorldTexture has already been finalized.
    */
   public void set(int x, int z, float[] frgb) {
-    if (endFinalized) throw new IllegalStateException("Attempted to modify a finalized WorldTexture.");
-
     long cp = ((long) x >> 4) << 32 | ((z >> 4) & 0xffffffffL);
     ChunkTexture ct = map.get(cp);
+
+    // This chunk texture might be used in other places so we must create a copy
+    if (compacted) {
+      ct = new ChunkTexture(ct);
+    }
+
     if (ct == null) {
       ct = new ChunkTexture();
       map.put(cp, ct);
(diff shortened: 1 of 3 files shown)


chunky/src/java/se/llbit/chunky/world/WorldTexture.java
```diff
@@ -98,16 +103,43 @@ public void store(DataOutputStream out) throws IOException {
...
     for (int i = 0; i < numTiles; ++i) {
       int x = in.readInt();
       int z = in.readInt();
       ChunkTexture tile = ChunkTexture.load(in);
+
+      if (textureCache.containsKey(tile)) {
+        tile = textureCache.get(tile);
+      } else {
+        textureCache.put(tile, tile);
+      }
```
leMaik, line 117: This should use `computeIfAbsent` instead

chunky/src/java/se/llbit/chunky/world/WorldTexture.java
```diff
@@ -42,8 +44,11 @@
    * Set color at (x, z)
    *
    * @param frgb RGB color components
+   * @throws IllegalStateException if this WorldTexture has already been finalized.
    */
   public void set(int x, int z, float[] frgb) {
+    if (endFinalized) throw new IllegalStateException("Attempted to modify a finalized WorldTexture.");
+
```
leMaik, line 51: I don't think this is needed. Adding new tiles after "finalization" will just make it un-deduped. Deduplication isn't a strict requirement for using the world texture.

chunky/src/java/se/llbit/chunky/world/WorldTexture.java
```diff
@@ -98,16 +103,43 @@ public void store(DataOutputStream out) throws IOException {
...
       texture.map.put(((long) x) << 32 | (z & 0xffffffffL), tile);
     }
     return texture;
   }
 
+  /**
+   * Deduplicate this {@code WorldTexture} to save memory. This also makes this read-only.
+   */
+  public void endFinalization() {
```
leMaik, line 127:
```suggestion
  public void compact() {
```

leMaik (changes requested):

chunky/src/java/se/llbit/chunky/world/WorldTexture.java
alexhliu: Ok. I changed it so that if it is already "finalized", it will duplicate the texture (since it could be used by other chunks) and modify the new texture.

PR #1003, opened 2021-06-26
Better error handling in Asynchronous Scene manager.

alexhliu:
Exceptions are now logged and shown to the user.

Out of memory errors get a special message:
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
@@ -91,7 +91,7 @@ public class AsynchronousSceneManager extends Thread implements SceneManager {
         Log.error("Chunky has run out of memory! Increase the memory given to Chunky in the launcher.");
       }
 
-      Log.error(e);
+      Log.error("Scene manager has crashed due to uncaught exception:", e);
       throw e;
     }
   }
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
@@ -86,6 +86,13 @@ public class AsynchronousSceneManager extends Thread implements SceneManager {
       }
     } catch (InterruptedException ignored) {
       // Interrupted.
+    } catch (Throwable e) {
+      if (e instanceof OutOfMemoryError) {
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
```diff
@@ -86,6 +86,13 @@ public void setOnChunksLoaded(Runnable onChunksLoaded) {
       }
     } catch (InterruptedException ignored) {
       // Interrupted.
+    } catch (Throwable e) {
+      if (e instanceof OutOfMemoryError) {
+        Log.error("Chunky has run out of memory! Increase the memory given to Chunky in the launcher.");
+      }
+
+      Log.error("Scene manager has crashed due to uncaught exception:", e);
```
leMaik, line 94:
```suggestion
      Log.error("Scene manager has crashed due to an uncaught exception. Chunky may not work properly until you restart it. If you think this is a bug, please report it to the developers.", e);
```

leMaik: …or maybe only show the "report bug" thing if the error is _not_ an out of memory error...

chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
```diff
@@ -86,6 +86,13 @@ public void setOnChunksLoaded(Runnable onChunksLoaded) {
       }
     } catch (InterruptedException ignored) {
       // Interrupted.
+    } catch (Throwable e) {
+      if (e instanceof OutOfMemoryError) {
+        Log.error("Chunky has run out of memory! Increase the memory given to Chunky in the launcher.");
+      }
```
leMaik, line 92: Should we only print one error (= only one tab) per error?

alexhliu: A stacktrace is probably helpful for advanced users. For example, an OOM in octree = use octree plugin. OOM in bvh = use bvh plugin.

leMaik: it sure is, but maybe put the stack trace into the same message so that it's still a single tab?

leMaik:
Nice, no more telling people to restart Chunky with debug console and try to reproduce the crash anymore! :clap:

PR #1004, opened 2021-07-02
Fix additional data and post processing progress.

### LINKED ISSUE #999: Scene's AdditionalData system is flawed.
Attempting to overwrite existing data will cause duplicate members to appear.
image

Of lesser importance, the add interface can be somewhat confusing as `JsonObject`s must be created. Maybe override with the most commonly used types?
image

alexhliu:
This closes #999


leMaik:
Thank you! :clap:

PR #1007, opened 2021-07-08
Stop render control buttons from causing Chunky to freeze while loading a scene.

### LINKED ISSUE #832: Clicking on the render start, pause, and stop buttons while chunks are loading causes a freeze.
Clicking on these buttons (and probably more) will cause Chunky to lockup until the chunks are finished loading. This is likely due to the fact that these button's callbacks all call `Scene` synchronized methods.

alexhliu:
* Added an `enqueueTask` api to AsynchronousSceneManager which allows for synchronized scene modifications to happen off the JavaFx thread.
* Add a query to to Scene to check if it is currently loading a scene.
* Adjust render control buttons to use the `enqueueTask` api and ignore clicks while the Scene is loading.

This closes #832 (Pausing loading should probably be separate since it would be quite complex)

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
@@ -28,7 +28,7 @@ import se.llbit.util.TaskTracker;
 import java.io.File;
 import java.io.IOException;
 import java.util.Collection;
-import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.LinkedBlockingQueue;
 
 /**
  * This scene manager is used for asynchronous loading and saving of scenes.
@@ -40,13 +40,13 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 public class AsynchronousSceneManager extends Thread implements SceneManager {
 
   private final SynchronousSceneManager sceneManager;
-  private final ConcurrentLinkedQueue<Runnable> taskQueue;
+  private final LinkedBlockingQueue<Runnable> taskQueue;
 
   public AsynchronousSceneManager(RenderContext context, RenderManager renderManager) {
     super("Scene Manager");
 
     sceneManager = new SynchronousSceneManager(context, renderManager);
-    taskQueue = new ConcurrentLinkedQueue<>();
+    taskQueue = new LinkedBlockingQueue<>();
   }
 
   public SceneProvider getSceneProvider() {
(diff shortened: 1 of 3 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
```diff
@@ -74,15 +76,13 @@ public void setOnChunksLoaded(Runnable onChunksLoaded) {
...
-          }
-        }
-        currentTask.run();
-        synchronized (this) {
-          currentTask = null;
+        synchronized (taskQueue) {
+          while (taskQueue.isEmpty())
+            taskQueue.wait();
         }
+        Runnable task = taskQueue.poll();
```
leMaik, line 83: Could the explicit wait and locking logic be eliminated by using a `LinkedBlockingQueue` and `Runnable task = taskQueue.take()`?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -558,6 +560,8 @@ public synchronized void saveScene(RenderContext context, TaskTracker taskTracke
    */
   public synchronized void loadScene(RenderContext context, String sceneName, TaskTracker taskTracker)
       throws IOException {
+    isLoading = true;
```
leMaik, line 563:
```
isLoading = true;
try {
  // loading logic
} finally {
  isLoading = false;
}
```

Just to be safe?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -3312,6 +3321,10 @@ public String getPreviewRenderer() {
     return previewRenderer;
   }
 
+  public boolean getIsLoading() {
```
leMaik, line 3324:
```suggestion
  public boolean isLoading() {
```

chunky/src/java/se/llbit/chunky/ui/ChunkyFxController.java
```diff
@@ -689,13 +689,22 @@ public synchronized void exportZip(File targetFile, ProgressTracker progress) {
     copyFrameBtn.setOnAction(this::copyCurrentFrame);
     start.setGraphic(new ImageView(Icon.play.fxImage()));
     start.setTooltip(new Tooltip("Start rendering."));
-    start.setOnAction(e -> scene.startRender());
+    start.setOnAction(e -> {
+      if (!scene.getIsLoading())
+        asyncSceneManager.enqueueTask(() -> scene.startRender());
+    });
```
leMaik, line 695: Could we listen to `isLoading` changes and enable/disable the buttons?

PR #1012, opened 2021-07-26
Buffered Dump IO

StanleyMines:
This buffers the Input/Output streams for dumps, allowing dump formats that take many small operations to be much much faster. May help on existing dumps by just a little.

(I thought this was added a while ago.)

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/renderdump/RenderDump.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/renderdump/RenderDump.java
@@ -16,10 +16,10 @@
  */
 package se.llbit.chunky.renderer.renderdump;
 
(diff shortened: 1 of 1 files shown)


leMaik:
What's the difference between `FastBufferedInputStream` and `BufferedInputStream`? Is it worth not using the built-in one here?

StanleyMines:
> What's the difference between `FastBufferedInputStream` and `BufferedInputStream`? Is it worth not using the built-in one here?

Not too sure. The Fast one seems to have some other features, but also caused a crash when I added the others as Fast, so I left the existing one as Fast, and changed the new ones to just using the default Java one.

leMaik:
Can you make both ones the Java ones just to make sure we don't introduce regressions with this? :pray:

leMaik:
Als always, thank you! :+1:

PR #1013, opened 2021-07-26
Downscale JavaFX Scene Preview

StanleyMines:
This takes the JavaFX preview window and downscales the preview view to fit within the JavaFX canvas sizes, thus avoiding crashes that might otherwise happen (such as #314 and #27).

Most of this was pulled out of #786. This should solve #314 and solve #27.

This allows for renders of sizes larger than 4k on all computers without crashing (as opposed to crashing on different computers at different sizes above 4k). This makes the new effective largest size 26.7k square, or 18* 1920x1080 (34560 x19440) before crashing due to maxing out the SampleBuffer (This can be solved by replacing the double[] with a double[][], as done in #786)

Here is what the canvas scale dropdown looks like now when the canvas is large enough to be scaled (previously, only had the % value on the right): 
  A screenshot showing the new Canvas Scale dropdown.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
+++ b/chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
@@ -44,11 +44,13 @@ import se.llbit.chunky.PersistentSettings;
 import se.llbit.chunky.renderer.*;
 import se.llbit.chunky.renderer.RenderManager;
 import se.llbit.chunky.renderer.scene.Camera;
+import se.llbit.chunky.renderer.scene.Scene;
 import se.llbit.chunky.resources.BitmapImage;
 import se.llbit.math.Vector2;
 import se.llbit.util.Pair;
 
 import java.nio.IntBuffer;
+import java.util.ArrayList;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
@@ -69,7 +71,7 @@ public class RenderCanvasFx extends ScrollPane implements Repaintable, SceneStat
   private static final int REDUCED_CANVAS_MAX_SIZE = 4096; // TODO: set via command line/launcher arg?
   private boolean previewShouldSubsample = true;
 
-  private final se.llbit.chunky.renderer.scene.Scene renderScene;
+  private final Scene renderScene;
 
   private WritableImage image;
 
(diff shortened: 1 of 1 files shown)


StanleyMines:
An addition to this could be to crop the displayed area instead of downscaling. Will look into this.

StanleyMines:
Most of this is from commit 33152565 in #786 (minor changes have been added since).

leMaik:
tbh I don't get what the new context menu options do. Also the preview is terribly slow for high resolutions. Maybe we could always render the preview at the preview size and only render the real image at full resolution? :thinking:

StanleyMines:
> tbh I don't get what the new context menu options do. Also the preview is terribly slow for high resolutions. Maybe we could always render the preview at the preview size and only render the real image at full resolution? 🤔

render at preview size would be... more difficult. I can look into changing it to that later.

The menu with the size percents is telling you what its displaying it as is effectively 400% if it were the full image, but since it was downscaled its only zooming 200% on the downscaled image, as an example. I wanted to convey both values, because otherwise when you set the zoom, it may appear blurry or pixelated at a low zoom, and you might get confused about canvas size vs preview size (like I did at one point). idk how to otherwise convey both values.

(If you are below the size where it scales the preview though, it will just have the standard %s and not mention downscaling.)

jackjt8:
image

12800x12800 ~22,650.7MB

Well.. I can't really test the upper bounds of this fix due to how much memory it is using so far but it doesn't crash.... available options are a bit complex.

StanleyMines:
Any ideas on how to make the options better?

jackjt8:
As lemaik has pointed out the scaling options are confusing..
image

Surely instead of downscaling it should read `upscaled`/`zoomed` if above 100%? Would be a bit less confusing.

StanleyMines:
> As lemaik has pointed out the scaling options are confusing..
> [image]
> 
> Surely instead of downscaling it should read `upscaled`/`zoomed` if above 100%? Would be a bit less confusing.

hmmm yeah that wording is a bit strange. It should be describing that the preview is zoomed to 400%, but due to downscaling, that 400% is down scaled 100% (or same zoom level as the output file will be), but since its zoom is 400% in javafx, its only displaying 1/16th of the pixels.

Does that make sense? I didnt really know how else to describe it in a couple words...

PR #1015, opened 2021-07-26
Scene Picker UI: Cleanup & Sort By Last Modified

### LINKED ISSUE #437: Sort scenes by most recently used
Scenes should be sorted by most recently used/edited in the scene manager.

IceSelkie:
- Default sorting of scenes is by most recent at the top, just as you've always wanted. You're welcome :)
 - Rebalanced column widths (Can now fit the default scene name)
 - ~~some stuff for cropping scenes, currently unused and/or commented out. Will be used once sub-scene tiling/cropping comes in #786)~~

Closed #437

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
+++ b/chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
@@ -124,7 +124,10 @@ public class SceneChooserController implements Initializable {
     });
     sizeCol.setCellValueFactory(data -> {
       SceneListItem scene = data.getValue();
-      return new ReadOnlyStringWrapper(scene.dimensions);
+      if (scene.cropping!=null)
+        return new ReadOnlyStringWrapper(scene.dimensions + scene.cropping);
+      else
+        return new ReadOnlyStringWrapper(scene.dimensions);
     });
     sppCol.setCellValueFactory(data -> {
       SceneListItem scene = data.getValue();
@@ -212,6 +215,11 @@ public class SceneChooserController implements Initializable {
      * Whether this scene description file is a backup file and the original .json is missing.
      */
     private final boolean isBackup;
+    /**
+     * Whether scene is cropped smaller than its main canvas size.
+     * Not currently used (Planned for Chunky 2.5.0; See PR #786)
+     */
+    private final String cropping;
 
     private SceneListItem(JsonObject scene, File sceneFile) {
       int width = scene.get("width").intValue(400);
@@ -228,6 +236,20 @@ public class SceneChooserController implements Initializable {
 
       this.dimensions = dimensions;
       sppCount = scene.get("spp").intValue(0);
+
+      cropping = null;
+      // Not currently used (Planned for Chunky 2.5.0; See PR #786)
+      //    JsonValue crop = scene.get("crop");
+      //    if (crop.isArray()) {
+      //      JsonArray cropArray = crop.asArray();
+      //      if (cropArray.size() >= 4) {
+      //        int w = cropArray.get(2).asInt(width);
+      //        int h = cropArray.get(3).asInt(height);
+      //        if (w != width || h != height || cropArray.get(0).asInt(0) != 0 || cropArray.get(1).asInt(0) != 0) {
+      //          cropping = " [" + w + "x" + h + "]";
+      //        } else cropping = null;
+      //      } else cropping = null;
+      //    } else cropping = null;
 
       long milliseconds = scene.get("renderTime").longValue(0);
       if (sppCount==0 || milliseconds<500) {
--- a/chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
+++ b/chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
@@ -180,7 +180,8 @@ public class SceneChooserController implements Initializable {
     List<File> fileList = SceneHelper.getAvailableSceneFiles(sceneDir);
     // TODO: Sort by recently modified instead of file name.
     Collections.sort(fileList, Comparator
-            .comparing((File o) -> o.toString().toLowerCase()) // ignore case on sort by file name.
+            .comparing((File o) -> -(Long)o.lastModified()) // negative of last modified -> most recent at top
+            .thenComparing((File o) -> o.toString().toLowerCase()) // Then by file name, ignoring case.
             .thenComparing(o -> o));
     for (File sceneFile : fileList) {
 
--- a/chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
+++ b/chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
@@ -1,5 +1,5 @@
-/*
- * Copyright (c) 2016 Jesper Öqvist <jesper@llbit.se>
+/* Copyright (c) 2016-2021 Jesper Öqvist <jesper@llbit.se>
+ * Copyright (c) 2016-2021 Chunky contributors
  *
  * This file is part of Chunky.
  *
@@ -31,6 +31,7 @@ import javafx.scene.control.TableView;
 import javafx.scene.control.Tooltip;
 import javafx.stage.FileChooser;
 import javafx.stage.Stage;
+import org.apache.commons.math3.util.FastMath;
 import se.llbit.chunky.main.SceneHelper;
 import se.llbit.chunky.renderer.scene.Scene;
 import se.llbit.fxutil.Dialogs;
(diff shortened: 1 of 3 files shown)


StanleyMines:
Most of this is from commit 195a872f in #786 (minor changes have been added since then):
https://github.com/chunky-dev/chunky/pull/786/commits/195a872f3c62aae314880b6a04c7ddfa4fe9335f

chunky/src/res/se/llbit/chunky/ui/SceneChooser.fxml
```diff
@@ -9,28 +26,28 @@
...
+    <children>
+        <Label text="Select scene to load:"/>
+        <TableView fx:id="sceneTbl" maxHeight="1.7976931348623157E308" prefHeight="293.0" VBox.vgrow="ALWAYS">
+            <columns>
+                <TableColumn fx:id="nameCol" prefWidth="200.0" text="Name"/>
+                <TableColumn fx:id="chunkCountCol" prefWidth="100.0" text="Chunks"/>
+                <TableColumn fx:id="sizeCol" prefWidth="150.0" text="Size"/>
+                <TableColumn fx:id="sppCol" prefWidth="100.0" text="Current SPP"/>
+                <TableColumn fx:id="renderTimeCol" prefWidth="100.0" text="Render time"/>
+            </columns>
```
leMaik, line 39: It would be cool to have "last modified" as a column here so that the user can manually sort by that, too.

StanleyMines: It would.... but that would make the window a lot wider or have a horizontal scroll bar, and I was trying to make it too too much wider, since the window is already pretty wide. (I already increased it from 600px wide to nearly 700, and with last modified, that'd probably be like 900 wide to show the whole thing.) (If they wanted to sort by "last modified" after sorting by something else, they could close the window and open it again, though a column would make it more obvious.)

If you still think it's worth it, I can go test that out and see what it looks like.

StanleyMines:
@leMaik 
> It would be cool to have "last modified" as a column here so that the user can manually sort by that, too.

In my previous comment I mentioned that this would make the load box much wider. Heres a comparison:


### Before this PR:
previous load scene box

### Currently in this PR:
pr load scene box

### Adding Date Modified Column:
proposed load scene box

ShirleyNekoDev:
IMO: Would be cool to have a menu to select what to show (example here: Windows explorer) with additional columns resulting in a horizontal bar if you don't want to increase the window size.
image

Otherwise, the horizontal space of the _Size_, _Current SPP_ and _Chunks_ column could be reduced.

MadDog443:
This is kind of wishful thinking but, what if in scene selection it showed a preview image of the scene?

StanleyMines:
> IMO: Would be cool to have a menu to select what to show (example here: Windows explorer) with additional columns resulting in a horizontal bar if you don't want to increase the window size.
> image
> 
> Otherwise, the horizontal space of the _Size_, _Current SPP_ and _Chunks_ column could be reduced.

Chunks could be reduced a little, Size will use that space up very soon once I finish subarea render things (probably), and Current SPP cant get much smaller while still showing the text. 

To be able to select what columns are visible like that.... Idk how to do that, and if we were to add that, that'd be for another PR. Feel free to open a feature request issue though if you want.

StanleyMines:
> This is kind of wishful thinking but, what if in scene selection it showed a preview image of the scene?

That seems... fairly difficult to do. I'm not too sure how that would be implemented, but thats definitely also outside the scope of this PR.

leMaik:
I agree, let's do the column selection and preview stuff later. :wink:

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
```diff
@@ -173,7 +178,11 @@ public void setStage(Stage stage) {
   private void populateSceneTable(File sceneDir) {
     List<SceneListItem> scenes = new ArrayList<>();
     List<File> fileList = SceneHelper.getAvailableSceneFiles(sceneDir);
-    Collections.sort(fileList);
+    // TODO: Sort by recently modified instead of file name.
```
leMaik, line 181: Doesn't this PR do just that? Also, the user should be able to change the sorting order by clicking on the table column head… it's weird that we force the sorting.

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
```diff
@@ -173,7 +178,11 @@ public void setStage(Stage stage) {
...
+    Collections.sort(fileList, Comparator
+            .comparing((File o) -> -(Long)o.lastModified()) // negative of last modified -> most recent at top
+            .thenComparing((File o) -> o.toString().toLowerCase()) // Then by file name, ignoring case.
+            .thenComparing(o -> o));
```
leMaik, line 185: This line does nothing?

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
```diff
@@ -211,28 +220,63 @@ public void setController(ChunkyFxController controller) {
      * Whether this scene description file is a backup file and the original .json is missing.
      */
     private final boolean isBackup;
+    /**
+     * Whether scene is cropped smaller than its main canvas size.
+     * Not currently used (Planned for Chunky 2.5.0; See PR #786)
+     */
+    private final String cropping;
```
leMaik, line 227: :smile: I was wondering where this came from

leMaik (changes requested):

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
IceSelkie: oop, yeah. I added that in at the beginning before I wrote the rest. I'll have to remove that.

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
IceSelkie: If a scene is duplicated by copy and paste, they could have the same modify time, but not the same file name, and thus a thenComparing will ensure that it is always sorted in the same way (similar to how minecraft does it), otherwise they might get sorted into different orders on different reads.

leMaik: Yes, but the third compare still doesn't do anything, does it?

ShirleyNekoDev: The third comparison only happens as fallback when lowercase pathnames are equal (depending on OS/file system) and checks `File::compareTo` via `Comparable` interface:
https://docs.oracle.com/javase/8/docs/api/java/io/File.html#compareTo-java.io.File-
_Compares two abstract pathnames lexicographically. The ordering defined by this method depends upon the underlying system._

But because this case is rarely relevant, we could remove the second comparison entirely and simply use the last one for both.

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
```diff
@@ -173,7 +178,11 @@ public void setStage(Stage stage) {
   private void populateSceneTable(File sceneDir) {
     List<SceneListItem> scenes = new ArrayList<>();
     List<File> fileList = SceneHelper.getAvailableSceneFiles(sceneDir);
-    Collections.sort(fileList);
+    // TODO: Sort by recently modified instead of file name.
+    Collections.sort(fileList, Comparator
+            .comparing((File o) -> -(Long)o.lastModified()) // negative of last modified -> most recent at top
```
ShirleyNekoDev, line 183:
```suggestion
            .comparingLong(File::lastModified)
            .reversed() // most recent by file last modified at top
```

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
```diff
@@ -173,7 +178,11 @@ public void setStage(Stage stage) {
...
+            .thenComparing((File o) -> o.toString().toLowerCase()) // Then by file name, ignoring case.
```
ShirleyNekoDev, line 184:
```suggestion
            .thenComparing(Comparator.naturalOrder()) // Then by File::compareTo the pathnames lexicographically
```

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
ShirleyNekoDev: I've added my suggestion for a (imo) easier understandable comparator (without the third comparison). :)

ShirleyNekoDev (changes requested):

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
IceSelkie: We want to sort the latest modified file to the top of the list, I believe this flips the order. If you use method reference instead of method call, you can't negate the values.

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
IceSelkie: Lexicographical order sorts all uppercase characters prior to all lower case characters, and that can be confusing if someone is looking for something by a certain name, and it happens to have different capitalization and thus not show up where they wanted, which is why I specifically chose to sort by lowercase strings.

But I suppose that is a little less important now that it is being sorted by last modified time... I would prefer to keep it this way, unless LeMaik wants otherwise.

ShirleyNekoDev: Good call, I haven't checked if my simpler code results in the same sorting. I'll check it right now!

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
ShirleyNekoDev: Comparison of sortings: https://gist.github.com/ShirleyNekoDev/1c3f58a9caa9e1943bac2b7cb46948bd
If you prefer your current sort oder I won't object ;)

ShirleyNekoDev (approved):

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
ShirleyNekoDev: Reversed applies a flip on the previous comparator (which is lastModified), therefore the last recent files are still sorted at the top :)

chunky/src/java/se/llbit/chunky/ui/SceneChooserController.java
ShirleyNekoDev: Uppercase before lowercase is indeed not ideal. On the other hand, names with suffix numbers are ordered somewhat more coherent (which is not that important because a matching last modified timestamp happens rather infrequently).

IceSelkie:
Merged @ThatRedox 's Async Scene Loading PR into the new scene picker. This also removed the third sort comparing line.

PR #1025, opened 2021-08-08
Move render controls UI

ShirleyNekoDev:
- move render controls into left panel
- give the start and pause buttons text
- add spacing to progress texts

pros: dedicated area for render controls (they even take up about the width of the sidebar) using less vertical space
cons: more "semantic distance" from the progress report in the layout

left side: current Chunky
right side: this PR
image

behaviour under minimal size:
image


leMaik:
Thank you, @ShirleyNekoDev :clap:

PR #1029, opened 2021-08-18
Faster BVH AABB intersect by removing pre-check.

ThatRedox:
```
Entity test world
  master: ~1.38 million sps
  branch: ~1.85 million sps
```


leMaik:
Would this be even faster when we replace FastMath with Math? Not sure how smart the intrinsics get.

ThatRedox:
Original pr had a slight issue
image
This now fixes it but is slightly slower. Still significantly faster than master however:
```
master: 1.225m sps
   fix: 1.553m sps
```

ThatRedox:
`Math` vs `FastMath` made no difference. I think both are compiled to be branchless.

abhayrawat15226:
iu

abhayrawat15226:
jh

PR #1031, opened 2021-09-03
Add CubicChunks support

NotStirred:
The commit history should be very clean, I didn't revert any changes so you should be able to go commit by commit to review

I'm going to do some more testing, and add javadoc to everything before this should be merged. This entire pr basically worked first try and that concerns me

Changes to existing classes:
- `Region` is now an interface, `MCRegion` is the vanilla implementation
- Create a `region` package for my own sanity
- All region objects are now created by the `World` implementation, previously several unrelated places
- Add a few min and max Y parameters on methods for chunk data

`ImposterCubicRegion` exists because CubicChunks regions are `256x256x256` blocks, while the vanilla ones are `1024*X*1024`, so internally the imposter takes the data from cubes for the requested height range and combines it into a chunk. This may be hundreds of regions vertically. It does that for each of the (currently) 4 region columns within its position

Also I may do some additional fixes for some oddities and bugs in the vanilla region stuff, let me know if that would be better in a separate PR after this one has been merged

PR #1032, opened 2021-09-08
Simple map view optimisations

NotStirred:
The main speedup for legacy worlds is the static initialisation of tags. Instead of creating a new tag object for every single block, it just indexes a pre-initialised 4096 length array.

In `SurfaceLayer#<init>` I *actually use the heightmap data* calculated the line before calling it, instead of recalculating it in the constructor

Ideally don't merge this PR at the same time as #1031 as I have to implement these optimisations there too. I'd prefer merging #1031 first, to keep things together that logically should be, but it doesn't really matter

### DIFF
--- a/chunky/src/java/se/llbit/chunky/block/legacy/LegacyBlocks.java
+++ b/chunky/src/java/se/llbit/chunky/block/legacy/LegacyBlocks.java
@@ -9,12 +9,27 @@ import se.llbit.nbt.Tag;
 
 public class LegacyBlocks {
 
+  /** statically initialised array of every possible tag */
+  private static final Tag[] legacyTags = new Tag[(1 << 8) * (1 << 4)];
+
+  static {
+    for (int id = 0; id < 1 << 8; id++) { //id is 8 bits
+      for (int data = 0; data < 1 << 4; data++) { //data is 4 bits
+        legacyTags[id * (1 << 4) + data] = getTag(id, data);
+      }
+    }
+  }
+
   public static Tag getTag(int offset, byte[] blocks, byte[] blockData) {
     int id = blocks[offset] & 0xFF;
     int data = 0xFF & blockData[offset / 2];
     data >>= (offset % 2) * 4;
     data &= 0xF;
 
+    return legacyTags[id * (1 << 4) + data];
+  }
+
+  private static Tag getTag(int id, int data) {
     CompoundTag tag = new CompoundTag();
     switch (id) {
       case 0:   return nameTag(tag, "air");
--- a/chunky/src/java/se/llbit/chunky/map/SurfaceLayer.java
+++ b/chunky/src/java/se/llbit/chunky/map/SurfaceLayer.java
@@ -50,14 +50,14 @@ public class SurfaceLayer extends BitmapLayer {
    * @param dim current dimension
    * @param chunkData data for the chunk
    */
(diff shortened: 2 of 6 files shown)


NotStirred:
Average time to load one chunk (seconds)
`pre : 0.0032`
`post: 0.0007`

from an average of several thousand chunks on the same 1.12 world

chunky/src/java/se/llbit/chunky/block/legacy/LegacyBlocks.java
```diff
@@ -9,12 +9,27 @@
 
 public class LegacyBlocks {
 
+  /** statically initialised array of every possible tag */
```
leMaik, line 12: I'm pretty sure that there are tags that contain additional data in 1.12 worlds, e.g. for bed colors.

NotStirred: bed colours work fine with this, just tested it

NotStirred: My understanding of the 1.12 saving is that everything is `blockID`(8bits) `meta`(4bits), and we have a function that maps from that to post flattening tags, which is `LegacyBlocks#getTag(int, int)` in this PR

NotStirred: If any blocks are broken now, they would've also been broken before this PR

leMaik: You're right… additional data is done by using block entities, e.g. for beds: https://github.com/chunky-dev/chunky/blob/master/chunky/src/java/se/llbit/chunky/block/legacy/blocks/LegacyBed.java#L21, so we can actually use this map to speed up world loading.

How much memory does this map consume? If it's too much, we could build it before loading chunks _once_ so that it could be garbage-collected after loading.

leMaik: Or, we could use some weak map :thinking:

NotStirred: Which map are you referring to?

chunky/src/java/se/llbit/chunky/map/SurfaceLayer.java
```diff
@@ -50,14 +50,14 @@
...
-  public SurfaceLayer(int dim, ChunkData chunkData, BlockPalette palette, int yMax) {
+  public SurfaceLayer(int dim, ChunkData chunkData, BlockPalette palette, int[] heightmapData) {
     bitmap = new int[Chunk.X_MAX * Chunk.Z_MAX];
     topo = new int[Chunk.X_MAX * Chunk.Z_MAX];
     for (int x = 0; x < Chunk.X_MAX; ++x) {
       for (int z = 0; z < Chunk.Z_MAX; ++z) {
 
         // Find the topmost non-empty block.
-        int y = Math.min(chunkData.maxY() - 1, yMax);
+        int y = Math.min(chunkData.maxY() - 1, heightmapData[z*16 + x]);
```
leMaik, line 60:
```suggestion
        int y = Math.min(chunkData.maxY() - 1, heightmapData[z * Chunk.X_MAX + x]);
```

leMaik (approved):

PR #1044, opened 2021-10-01
Fix parallel projection autofocus

### LINKED ISSUE #946: Parallel projection autofocus does not "focus"
The autofocus uses the camera position do find the subject distance between camera and the block which is in the center of the camera projection. Because parallel projection "offsets the height as horizontal movement" (at least that is the result, because parallel projection), the vertical distance part of the subject distance should probably be excluded from the focus calculations or something like that?

Auto-focussed render for standard projection with adjusted DoF:
image

Auto-focussed render for parallel projection with adjusted DoF:
image

Expected focussed parallel projection (by subtracting the camera y position from the subject distance):
image

ShirleyNekoDev:
fixes #946

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/projection/ParallelProjector.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/projection/ParallelProjector.java
@@ -24,11 +24,11 @@ import se.llbit.math.Vector3;
  * Casts parallel rays from different origin points on a plane
  */
 public class ParallelProjector implements Projector {
-  protected final double worldWidth;
+  protected final double worldDiagonalSize;
   protected final double fov;
 
-  public ParallelProjector(double worldWidth, double fov) {
-    this.worldWidth = worldWidth;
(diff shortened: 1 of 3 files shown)


ShirleyNekoDev:
This fix produces the correct results, but it would be better to fix the math behind this issue.

The projector of the camera uses the octree diagonal to set the initial ray location (https://github.com/chunky-dev/chunky/blob/71939db128ff94f4e31ca6afa82cfe6c7d6f7d62/chunky/src/java/se/llbit/chunky/renderer/scene/Camera.java#L214-L220). The autofocus then uses the distance between the camera and the scene intersection (https://github.com/chunky-dev/chunky/blob/71939db128ff94f4e31ca6afa82cfe6c7d6f7d62/chunky/src/java/se/llbit/chunky/renderer/scene/Camera.java#L674-L685) to focus.

The problem is that this distances is incorrect for parallel projection because the FoV of the parallel projector appears like a zoom of the camera towards the scene.

PR #1052, opened 2021-10-05
Allow negative numbers as command line arguments in options

### LINKED ISSUE #1030: Can't pass negative numbers as arguments for headless mode
Trying to run commands such as:

java -jar ChunkyLauncher.jar -set camera.orientation.pitch -1 default 

fail because of the negative 1

ShirleyNekoDev:
Allows negative numbers for _value_ in `-set <NAME> <VALUE> [<SCENE>]`
Fixes #1030


leMaik (approved):
Looks good. Maybe we can use [insert CLI library here] one day, though?

Issue #977, opened 2021-05-25
Download a player skin by UUID or player name


leMaik:
It is already possible to load player skins from a file and with #975 skins of players in the world also get downloaded automatically. The only missing part is adding skins to existing or manually added players.

We could add a "Download skin…" button next to the "Select skin…" button. This should open a popup where one can either enter a player name or a UUID, which would then be resolved and the skin downloaded.

PR #1057, opened 2021-10-09
Download player skins by UUID or Username

ThisIsBrady:
This adds the ability to download skins from the Entities tab. Pressing the "Download Skin" button opens a small popup which allows the user to enter either a UUID or Username. If a skin can't be found using either a username or uuid lookup, a small warning will appear in the popup. The model of the entity will be updated based on the parameters of the downloaded skin.

As a side-effect, this also adds the ability to lookup a uuid using a username. Results from the lookup are stored in a cache file.

Implements #977.

New button in the Entities tab:
java_2021-10-08_20-43-40

Popup window:
java_2021-10-08_20-28-32

Warning message:
java_2021-10-08_20-45-15


leMaik:
Nice popup window! I think I'll refactor that and use it for the scene rename dialog.

Issue #1046, opened 2021-10-03
UI clear up and removal of right panel


jackjt8:
![general](https://user-images.githubusercontent.com/3740917/135766377-a498af2b-e009-4f26-a218-9b9b763374c1.png)

`Scene: <scenename>`, `Scene save`, and `Scene load` should be relocated under the `Scene` tab. The `<scenename>` should be reflected in the Chunky Window name.

`Save current frame` and `Copy current frame` could be relocated in the `Render Preview`s `Right click menu`.

---

`Map View` tab --> Map context menu / Menu bar (I'm not sure about the exact locations for these).

---

![chunks_tab](https://user-images.githubusercontent.com/3740917/135766439-8131cd74-6556-4946-9296-6970bb3a1bc9.png)

`Clear Selection` is redundant.

`Export view to PNG`  should be listed along side `Map View` tab content, probably within the map context menu.

`Export chunks to ZIP` and `Delete selected chunks` should be in the menu bar.

---

![options_tab](https://user-images.githubusercontent.com/3740917/135766480-f1a14be4-5869-4126-a32c-afe2cd7c696b.png)

`Edit resource packs`, `Disable default textures`, and `single color textures` could be placed either with Render Controls, or due to the impact on the map view in the menu bar.

`Open Scenes Directory` - Redundant.

`Change Scenes Directory` - Menu bar.

---

About --> Menu bar + cleanup.

PR #1058, opened 2021-10-09
Improve UI

### LINKED ISSUE #1173: Add a way to browse for a scene to load
Currently, Chunky allows you to change the world directory and browse for a world to load from within the _Change World_ menu. The _Load Scene_ menu does not have such functionality. The _Change Scene Directory_ option exists, but it is hidden in the rarely accessed _Options_ tab. _Change Scene Directory_ should be added to the _Load Scene_ menu, and "Browse for another scene" should be added as an option to that menu as well, to mirror the _Change World_ menu. It would also be useful in the event someone downloads a scene to render, but does not know that is should go in the `scenes` folder. Then they can simply browse for the scene.

leMaik:
Related to #1046 
Closes #1173 

Only the _Map View_ and _Options_ tabs are remaining.

image
image
image
image
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/RenderContext.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/RenderContext.java
@@ -19,157 +19,83 @@ package se.llbit.chunky.renderer;
 
 import se.llbit.chunky.main.Chunky;
 import se.llbit.chunky.main.ChunkyOptions;
-import se.llbit.chunky.renderer.scene.Scene;
 
 import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.FileOutputStream;
-import java.io.InputStream;
-import java.io.OutputStream;
 
 /**
  * Rendering context - keeps track of the Chunky configuration used for rendering.
  *
  * @author Jesper Öqvist <jesper@llbit.se>
  */
-public class RenderContext {
+public class RenderContext implements SceneIOProvider {
 
-  protected final Chunky chunky;
-  protected final ChunkyOptions config;
-  public RenderWorkerPool.Factory renderPoolFactory = RenderWorkerPool::new;
-  private File sceneDirectory;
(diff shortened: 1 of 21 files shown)


jackjt8:
I like these changes

Peregrine05:
Maybe rename the `Help` tab to "Camera Controls" or simply move its contents into the `Camera` tab, so there are not two `Help` tabs/menus.

jackjt8:
New UI looks really good and most things are self explanatory.

`Map View` right click menu item `Export chunks as zip file...` should read `Export selection as zip file...` given all other selection dependant options use this terminology.
image


`Scene` > `Open scene directory` should be greyed out if no scene is loaded.

image


`File` > `Save`, `Save as..`, `Save  copy...` work as expected.

`File` Shortcuts are... going to have a learning curve. If you don't click on the left pane you end up moving your render... 

image

> I do believe the restoring the `Lock` button that was around from the 1.X days is worthwhile. That way people have disable keyboard/mouse inputs to the preview to avoid accidental movement.

`Resources Packs & Textures`

I wonder if we need a pop out editor if we have such a small tab? Surely we could embed the list of resource packs to make it quicker and easier to see what packs are loaded?

image

We should also improve how information is displayed here. Showing the full path to the resource pack is not user-friendly / shouldn't be the main focus. We should display the file/folder name, resource pack version (pack.mcmeta), and pack description (pack.mcmeta) as a priority in my opinion. I may create a separate issue to track these changes as they may be out of scope for this PR and would delay things.

---

In any case this is a massive improvement.

leMaik:
@jackjt8 Thanks for the feedback! :+1: 

I improved the chunk export label. The _open scene directory_ button is now disabled if the scene directory doesn't exist. The shortcuts do nothing if the render view is focused, so I guess that's fine for now.

As discussed in chat, the resource packs dialog is out-of-scope for this PR, but @ShirleyNekoDev has some great work going on, so it might be vastly improved in the future.

PR #1071, opened 2021-11-18
Rename getN to getNormal.

ThatRedox:
This was unintentionally changed in #901. Change `Ray.getN()` back to `Ray.getNormal()` to fix compatibility with plugins such as denoiser. Also re-add `Ray.orientNormal(...)` which saves on a `Vector3` allocation which should improve rendering performance. Clean up the old intersection logic from `TorchModel` which was not being used.


leMaik:
Thank you!

PR #1078, opened 2021-12-01
Basic support of the new block data location in 1.18

NotStirred:
Does not support biome palettes.

Across the snapshots the sections array has been in `.Level.Sections` -> `.Level.sections` -> `.sections`
do you want me to add support for `.Level.sections` as well? iirc it's in a few of the snapshots though I can't really remember

Notable changes for review:
- Added `getTagFromNames` function, which gets the first valid tag from the given names (helps with mojang moving around identical structures)
- Changed `Map<String, Tag> data` to `Tag data` for easier use of `getTagFromNames`


NotStirred:
very nice :+1:

PR #1082, opened 2021-12-05
Implement water shading using noise

### LINKED ISSUE #628: Making the Water configurable
Would you be able to add a ui element to chunky that would enable you to change the height of the waves? Or how random they could be?

leMaik had asked me to create an issue on GitHub for this. As a reminder. ;)

Thanks,
Pylons

### LINKED ISSUE #880: Water waves create a grid effect
This is the same topic I was discussing on Discord, but I thought I'd leave a GitHub issue here as well. When viewing from a distance, the waves in water generate a grid pattern, which could be solved using some sort of perlin noise (as leMaik suggested) or something more complicated.

image

aTom3333:
Closes #880, probably closes #628 as well?
This PR adds a new option to use procedural water shading and a few parameters to control it. The new procedural water is a simple fractal noise based on simplex noise. I tried to do something more fancy but in the end it just looked bad so I decided to stick with the simplicity.

Example:
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/model/CauldronModel.java
+++ b/chunky/src/java/se/llbit/chunky/model/CauldronModel.java
@@ -413,7 +413,7 @@ public class CauldronModel {
     Quad water = waterLevels[level];
     if (water != null && water.intersect(ray)) {
       if (!scene.stillWaterEnabled()) {
-        scene.getWaterShading().doWaterShading(ray);
+        scene.getWaterShading().doWaterShading(ray, scene.getAnimationTime());
       } else {
         ray.setNormal(water.n);
       }
--- a/chunky/src/java/se/llbit/chunky/model/CauldronModel.java
+++ b/chunky/src/java/se/llbit/chunky/model/CauldronModel.java
@@ -18,6 +18,7 @@ package se.llbit.chunky.model;
 
 import se.llbit.chunky.block.Lava;
 import se.llbit.chunky.block.Water;
+import se.llbit.chunky.renderer.scene.Scene;
 import se.llbit.chunky.resources.Texture;
 import se.llbit.math.Quad;
 import se.llbit.math.Ray;
@@ -392,7 +393,7 @@ public class CauldronModel {
     return hit;
   }
 
-  public static boolean intersectWithWater(Ray ray, boolean stillWater, int level) {
+  public static boolean intersectWithWater(Ray ray, Scene scene, int level) {
     boolean hit = false;
     ray.t = Double.POSITIVE_INFINITY;
     for (int i = 0; i < quads.length; ++i) {
@@ -411,9 +412,8 @@ public class CauldronModel {
     // TODO since this water is the same block, refraction is not taken into account – still better than no water
     Quad water = waterLevels[level];
     if (water != null && water.intersect(ray)) {
-      if (!stillWater) {
-        // TODO Hook into the more generic water shading?
-        WaterModel.doWaterDisplacement(ray);
+      if (!scene.stillWaterEnabled()) {
+        scene.getWaterShading().doWaterShading(ray);
       } else {
         ray.setNormal(water.n);
       }
--- a/chunky/src/java/se/llbit/chunky/model/CauldronModel.java
+++ b/chunky/src/java/se/llbit/chunky/model/CauldronModel.java
@@ -412,6 +412,7 @@ public class CauldronModel {
     Quad water = waterLevels[level];
     if (water != null && water.intersect(ray)) {
       if (!stillWater) {
+        // TODO Hook into the more generic water shading?
         WaterModel.doWaterDisplacement(ray);
       } else {
         ray.setNormal(water.n);
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -236,6 +236,7 @@ public class Scene implements JsonSerializable, Refreshable {
   protected double waterPlaneHeight = World.SEA_LEVEL;
   protected boolean waterPlaneOffsetEnabled = true;
   protected boolean waterPlaneChunkClip = true;
+  protected WaterShading waterShading = new WaterShading();
 
   /**
    * Enables fast fog algorithm
@@ -499,6 +500,7 @@ public class Scene implements JsonSerializable, Refreshable {
     waterPlaneHeight = other.waterPlaneHeight;
     waterPlaneOffsetEnabled = other.waterPlaneOffsetEnabled;
     waterPlaneChunkClip = other.waterPlaneChunkClip;
+    waterShading.set(other.waterShading);
 
     spp = other.spp;
     renderTime = other.renderTime;
@@ -2602,6 +2604,12 @@ public class Scene implements JsonSerializable, Refreshable {
       colorObj.add("blue", waterColor.z);
       json.add("waterColor", colorObj);
     }
+    json.add("useProceduralWater", !waterShading.useFixedTexture);
+    if(!waterShading.useFixedTexture) {
+      json.add("proceduralWaterIterations", waterShading.iterations);
+      json.add("proceduralWaterFrequency", waterShading.baseFrequency);
+      json.add("proceduralWaterAmplitude", waterShading.baseAmplitude);
+    }
     JsonObject fogColorObj = new JsonObject();
     fogColorObj.add("red", fogColor.x);
     fogColorObj.add("green", fogColor.y);
@@ -2895,6 +2903,12 @@ public class Scene implements JsonSerializable, Refreshable {
       waterColor.x = colorObj.get("red").doubleValue(waterColor.x);
       waterColor.y = colorObj.get("green").doubleValue(waterColor.y);
       waterColor.z = colorObj.get("blue").doubleValue(waterColor.z);
+    }
+    waterShading.useFixedTexture = !json.get("useProceduralWater").boolValue(false);
+    if(!waterShading.useFixedTexture) {
+      waterShading.iterations = json.get("proceduralWaterIterations").intValue(4);
+      waterShading.baseFrequency = json.get("proceduralWaterFrequency").doubleValue(0.1);
+      waterShading.baseAmplitude = json.get("proceduralWaterAmplitude").doubleValue(0.2);
     }
     JsonObject fogColorObj = json.get("fogColor").object();
     fogColor.x = fogColorObj.get("red").doubleValue(fogColor.x);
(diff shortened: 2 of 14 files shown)


chunky/src/java/se/llbit/chunky/model/CauldronModel.java
```diff
@@ -412,6 +412,7 @@ public static boolean intersectWithWater(Ray ray, boolean stillWater, int level)
     Quad water = waterLevels[level];
     if (water != null && water.intersect(ray)) {
       if (!stillWater) {
+        // TODO Hook into the more generic water shading?
```
leMaik, line 415: Not sure if overkill, but you _could_ pass the `scene` down into this method to get access to `scene.getWaterShading()`.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2602,6 +2604,12 @@ public void setUseCustomWaterColor(boolean value) {
       colorObj.add("blue", waterColor.z);
       json.add("waterColor", colorObj);
     }
+    json.add("useProceduralWater", !waterShading.useFixedTexture);
+    if(!waterShading.useFixedTexture) {
+      json.add("proceduralWaterIterations", waterShading.iterations);
+      json.add("proceduralWaterFrequency", waterShading.baseFrequency);
+      json.add("proceduralWaterAmplitude", waterShading.baseAmplitude);
+    }
```
leMaik, line 2612: Can we put this into a new `proceduralWater` object?

chunky/src/java/se/llbit/chunky/renderer/scene/WaterShading.java
```diff
@@ -0,0 +1,75 @@
...
+ * along with Chunky.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package se.llbit.chunky.renderer.scene;
+
+import se.llbit.chunky.model.WaterModel;
+import se.llbit.math.Ray;
+import se.llbit.math.SimplexNoise;
+import se.llbit.math.Vector3;
+
+public class WaterShading {
```
leMaik, line 24: Maybe we can extract an interface from this to allow plugins to hook into water shading? And then have two implementations (`SimplexWaterShader implements WaterShader` and `LegacyWaterShader implements WaterShader`)?

chunky/src/java/se/llbit/math/SimplexNoise.java
```diff
@@ -0,0 +1,217 @@
...
+//
+// This library is public domain software, released by the author
+// into the public domain in February 2011. You may do anything
+// you like with it. You may even remove all attributions,
+// but of course I'd appreciate it if you kept my name somewhere.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// General Public License for more details.
```
leMaik, line 25: Is this public domain, gpl or dual-licensed? We should probably add the notice to the readme and about dialog, too.

leMaik (changes requested):

chunky/src/java/se/llbit/math/SimplexNoise.java
aTom3333: I've taken the code from here https://github.com/keijiro/sketches2016/blob/master/Simplex2/SimplexNoise.java which is itself a java reimplementation of the code from here https://github.com/stegu/perlin-noise/blob/master/src/sdnoise1234.c. You can see that the original code no longer has the GPL mention, see https://github.com/stegu/perlin-noise/issues/4
So the original code is definitely public domain, but I am unsure about the reimplementation. I assume that the author intended to release under the same condition but this is a bit unclear... Maybe I should try to ask the author?
As for adding them in the readme+dialog, no problem

chunky/src/java/se/llbit/chunky/renderer/scene/WaterShading.java
aTom3333: And doing the usual registry of implementations, dropdown instead of checkbox and the work I suppose? Why not

chunky/src/java/se/llbit/math/SimplexNoise.java
leMaik: Clearly, the author of the Java reimplementation wanted to un-license it: https://github.com/keijiro/sketches2016#license

chunky/src/java/se/llbit/chunky/renderer/scene/WaterShading.java
leMaik: Maybe one day, but that's not blocking this PR from being merged. :wink: For now, I'd prefer interface+inheritance instead of the `useFixedTexture` magic boolean.

chunky/src/java/se/llbit/chunky/model/CauldronModel.java
aTom3333: Forgot that intersect already has the scene, thought this would mean changing the signature of the intersect function of Block (and every derived class) but it only requires minimal change so I'll do it

ThatRedox:
Why not use something like [this implementation](https://github.com/KdotJPG/OpenSimplex2/blob/master/java/OpenSimplex2F.java#L135) which as a 3D implementation of open simplex noise in order to make animations work?

aTom3333:
I wanted an implementation that gives the partial derivative of the noise as well as this is how the normal can be computed. If the implementation doesn't provide the partial derivatives, it can still be approximated using the definition of the derivative dN(x, y, z)/dx = (N(x+h, y, z) - N(x, y, z)) / h for some small h, so this is not an absolute requirement but having an implementation that gives the partial derivative means we only need to evaluate the noise function once instead of 3 times.
To be honest I didn't even think of animation when implementing that so I was looking for a 2D implementation in the first place.
Ideally, we'd want to find a 3D implementation that gives the partial derivatives as well, but maybe I'm too fixated on that and having multiple evaluation is not an issue.
I think we can do one of the following:
 * Keep the 2D noise
 * Find a 3D implementation with partial derivatives (ideal)
 * Implement it ourselves based on https://github.com/stegu/perlin-noise/blob/master/src/sdnoise1234.c#L336
 * Use any 3D implementation and compute the derivative by evaluating the function at 3 points
I'd like to know what you all think about this

aTom3333:
Found a java implementation by the same author based on the same original C implementation https://github.com/keijiro/sketches2017/blob/master/Simplex3D/SimplexNoise.java
I'll make the changes this evening.
I don't know what the benefits of having the time variable be treated different as in the implementation you linked are

chunky/src/java/se/llbit/chunky/ui/CreditsController.java
```diff
@@ -115,6 +117,11 @@ public void initialize(URL location, ResourceBundle resources) {
     fastutilLicense.setBorder(Border.EMPTY);
     fastutilLicense.setOnAction(e -> launchAndReset(fastutilLicense, "http://www.apache.org/licenses/LICENSE-2.0"));
 
+    simplexnoise.setBorder(Border.EMPTY);
+    simplexnoise.setOnAction(e -> launchAndReset(fastutil, "https://github.com/keijiro/sketches2016/blob/master/Simplex2/SimplexNoise.java"));
```
leMaik, line 121:
```suggestion
    simplexnoise.setOnAction(e -> launchAndReset(simplexnoise, "https://github.com/keijiro/sketches2016/blob/master/Simplex2/SimplexNoise.java"));
```

chunky/src/java/se/llbit/chunky/ui/CreditsController.java
```diff
@@ -115,6 +117,11 @@ public void initialize(URL location, ResourceBundle resources) {
...
+    simplexnoiseLicense.setBorder(Border.EMPTY);
+    simplexnoiseLicense.setOnAction(e -> launchAndReset(fastutilLicense, "http://unlicense.org/"));
```
leMaik, line 123:
```suggestion
    simplexnoiseLicense.setOnAction(e -> launchAndReset(simplexnoiseLicense, "http://unlicense.org/"));
```

chunky/src/java/se/llbit/chunky/ui/render/WaterTab.java
```diff
@@ -156,8 +181,72 @@ public void initialize(URL location, ResourceBundle resources) {
...
+
+    proceduralWaterAnimationSpeed.setName("Animation speed");
+    proceduralWaterAnimationSpeed.setTooltip("Animation speed of the water. "
+            + " Only relevant when rendering animation by varying animation time.");
+    proceduralWaterAnimationSpeed.setRange(0, 10);
+    proceduralWaterAnimationSpeed.onValueChange(speed -> {
+      WaterShader shader = scene.getWaterShading();
+      if(shader instanceof SimplexWaterShader) {
+        ((SimplexWaterShader) shader).animationSpeed = speed;
+        scene.refresh();
```
leMaik, line 246:
```suggestion
```

chunky/src/java/se/llbit/chunky/ui/render/WaterTab.java
```diff
@@ -156,8 +181,72 @@ public void initialize(URL location, ResourceBundle resources) {
...
+    });
+
+    proceduralWaterAmplitude.setName("Amplitude");
+    proceduralWaterAmplitude.setTooltip("The amplitude of the noise");
+    proceduralWaterAmplitude.setRange(0, 1);
+    proceduralWaterAmplitude.onValueChange(amp -> {
+      WaterShader shader = scene.getWaterShading();
+      if(shader instanceof SimplexWaterShader) {
+        ((SimplexWaterShader) shader).baseAmplitude = amp;
+        scene.refresh();
```
leMaik, line 233:
```suggestion
```

chunky/src/java/se/llbit/math/SimplexNoise.java
```diff
@@ -0,0 +1,302 @@
...
+// but of course I'd appreciate it if you kept my name somewhere.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+//
+
+package se.llbit.math;
+
+public final class SimplexNoise
```
leMaik, line 29:
```suggestion
/**
 * A Simplex noise generator with true analytic derivative in 1D to 4D, based on sdnoise1234.
 *
 * @author Stefan Gustavson
 * @author Keijiro Takahashi
 */
public final class SimplexNoise
```

leMaik (approved):

PR #1102, opened 2021-12-19
Add -reload-chunks option to headless mode

Doelia:
Haven't found a method to load changes mades on my world.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/main/CommandLineOptions.java
+++ b/chunky/src/java/se/llbit/chunky/main/CommandLineOptions.java
@@ -258,6 +258,10 @@ public class CommandLineOptions {
       options.force = true;
     });
 
+    registerOption("-reload-chunks", new Range(0), arguments -> {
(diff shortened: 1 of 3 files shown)


leMaik:
Thank you, nice first PR! :tada:

chunky/src/java/se/llbit/chunky/main/CommandLineOptions.java
```diff
@@ -258,6 +258,10 @@ public CommandLineOptions(String[] args) {
       options.force = true;
     });
 
+    registerOption("-reload-chunks", new Range(0), arguments -> {
```
leMaik, line 261: We should probably add this option to the [USAGE string](https://github.com/chunky-dev/chunky/blob/master/chunky/src/java/se/llbit/chunky/main/CommandLineOptions.java#L66), so that people know it is there.

leMaik (approved):

PR #1103, opened 2021-12-19
Headless mode : Allow json parameter in -set option

Doelia:
Let the following command :
```
chunky -set chunkList "[[1,-2]]" scene
```

Before this commit, scene.json will result in (which is invalid) :
```
{
...
  "chunkList": "[[1,-2]]",
...
}
```

After this commit, scene.json will result in :
```
{
...
  "chunkList": [
    [
      1,
      -2
    ]
  ],
...
}
```


NotStirred:
Feels a bit weird to check with `startsWith`. Could try try to parse as an array if it throws, default to the previous impl?
I'm sure there is a nicer way, can't think of one at the moment

ShirleyNekoDev:
I'd say this is fine for now and could get merged - maybe a follow-up PR later for a different implementation if needed.

leMaik:
Thanks, @Doelia :tada: And sorry it took so long to get merged.

PR #1109, opened 2021-12-29
Improve emitter sampling.

ThatRedox:
* Emitter sampling now attempts to sample the faces of blocks (including custom block models). 
* Increment emitter grid version. New emitter grid stores emitter block ID and position.
* Fixed some of the math behind emitter sampling. 
* Added `ONE_BLOCK` which samples all the faces of a random block within the emitter grid cell.
* Added `CachedObjectProvider` which caches some small objects like `Vector3` to be reused which improves performance and lightens the load on the GC a lot.

`ALL` 256 spp, 22 minutes 44 sec
Natural Test World-all-256-22min44sec

`ONE_BLOCK` 256 spp, 3 min 41 sec
Natural Test World-one_block-256-3min41sec

`ONE` 256 spp, 1 min 46 sec
Natural Test World-one-256-1min46sec

`NONE` 256 spp, 1 min 19 sec
Natural Test World-none-256-1min19sec


chunky/src/java/se/llbit/chunky/model/BlockModel.java
```diff
@@ -3,9 +3,19 @@
...
+
+import java.util.List;
+import java.util.Random;
 
 @PluginApi
 public interface BlockModel {
 
   boolean intersect(Ray ray, Scene scene);
+
+  int numFaces();
```
ShirleyNekoDev, line 16:
```suggestion
  int faceCount();
```
I've seen more "*Count()" methods in Java projects (might be easier to read because the thing we count is named first?).

chunky/src/java/se/llbit/chunky/renderer/CachedObjectProvider.java
```diff
@@ -0,0 +1,78 @@
+package se.llbit.chunky.renderer;
+
+import se.llbit.math.Ray;
+import se.llbit.math.Vector3;
+
+import java.util.ArrayList;
+import java.util.function.Supplier;
+
```
ShirleyNekoDev, line 8: Could you add a quick JavaDocs to describe this class and when to use it?
Btw. this is an object pool right? Maybe add pool somewhere in the name, because this describes the design pattern (https://en.wikipedia.org/wiki/Object_pool_pattern).

chunky/src/java/se/llbit/chunky/renderer/EmitterSamplingStrategy.java
```diff
@@ -1,5 +1,5 @@
 package se.llbit.chunky.renderer;
 
 public enum EmitterSamplingStrategy {
-  NONE, ONE, ALL
+  NONE, ONE, ONE_BLOCK, ALL
```
ShirleyNekoDev, line 4: Please add documentation what the expectation is for each strategy ;)

ThatRedox:
Loading emittergrid is currently broken. Will undraft once that is fixed.

ThatRedox:
Emitter grid was being loaded before octree. New emitter grid requires block data so emitter grid loading was moved to after loading the octree.

ThatRedox:
Benchmarks proved not having a rendering object pool was fastest.
| Test          | SPS ± 95% CI   |
|---------------|----------------|
| No Pool       | 767,625 ± 0.43% |
| Simple Pool   | 747,095 ± 0.45% |
| Complex Pool  | 734,456 ± 0.46% |
| Bypassed Pool | 718,548 ± 0.36% |

ThatRedox:
Campfires and candle flames now work.
image

chunky/src/java/se/llbit/chunky/entity/Campfire.java
```diff
@@ -347,9 +349,14 @@ private static int getOrientationIndex(String facing) {
 
   @Override
   public Grid.EmitterPosition[] getEmitterPosition() {
+      if (block == null) {
+          Log.warn("Attempted to build emitter grid from unassociated campfire entity.");
+          return new Grid.EmitterPosition[0];
```
leMaik, line 354: This means that it won't work after loading an existing scene. Should we store the associated block position to fix this?

ThatRedox: No it should still work. The emitter grid file stores all the `EmitterPosition`s and can get all the information needed from there. This shouldn't ever be called once a scene is loaded.

leMaik:
Awesome! :partying_face:

PR #1113, opened 2021-12-31
Implement changing the number of render threads without restarting Chunky.

ThatRedox:
* `RenderManager` now has an optional method to `setThreadCount`
* Implement changing the number of threads in `DefaultRenderManager` and `RenderWorkerPool`
  * Render workers are now stored in an `ArrayList`, workers are sequentially assigned id's and seeds.
  * Replace `RenderWorkerPool.threads` with a `getThreadCount` method.
  * Update interrupt logic and work waiting logic to support killing zombie workers.
* Allow for overwriting the common thread pool to change the number of threads.
* Update advanced tab to support changing the number of threads.


chunky/src/java/se/llbit/chunky/main/Chunky.java
```diff
@@ -337,6 +338,15 @@ public static ForkJoinPool getCommonThreads() {
     return commonThreads;
   }
 
+  /**
+   * Replace the common thread pool with a new thread pool with a specified number of threads.
+   * Note: This is done on a best-effort basis.
+   */
+  public static void setCommonThreadsCount(int threads) {
+    getCommonThreads();
```
leMaik, line 346:
```suggestion
```

ThatRedox:
Should probably do a quick test before merging, its been a while.

leMaik:
image

PR #1114, opened 2021-12-31
Show a warning in the launcher when not using a 64-bit JVM

### LINKED ISSUE #1100: Warn User on Java 32 Bit Use
The base concept is really simple, the user should be told that they are using a 32 bit version of Java. As to how this could be done I have a few ideas. 
* Print a message to the console, the obvious draw back though is it that the user would have to enable the console
* Have a banner/alert in the launcher that does the same thing as the console message, but in the launcher.
* Limit the ram to 4G

leMaik:
Closes #1100

### DIFF
--- a/launcher/src/se/llbit/chunky/launcher/ChunkyDeployer.java
+++ b/launcher/src/se/llbit/chunky/launcher/ChunkyDeployer.java
@@ -500,7 +500,7 @@ public final class ChunkyDeployer {
       cmd.add(ChunkyLauncher.class.getName());
       cmd.add("--checkJvm");
       ProcessBuilder builder = new ProcessBuilder(cmd);
-      return builder.start().waitFor() != 0;
+      return builder.start().waitFor() == 0;
     } catch (Exception e) {
     }
     return true;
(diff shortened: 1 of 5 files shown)


launcher/src/se/llbit/chunky/launcher/ChunkyDeployer.java
```diff
@@ -489,4 +490,19 @@ public static boolean canLaunch(VersionInfo version, ChunkyLauncherController la
...
+  public static boolean is64BitJvm(LauncherSettings settings) {
+    try {
+      ArrayList<String> cmd = new ArrayList<>();
+      cmd.add(JreUtil.javaCommand(settings.javaDir));
+      cmd.add("-cp");
+      cmd.add(ManagementFactory.getRuntimeMXBean().getClassPath());
+      cmd.add(ChunkyLauncher.class.getName());
+      cmd.add("--checkJvm");
+      ProcessBuilder builder = new ProcessBuilder(cmd);
+      return builder.start().waitFor() != 0;
```
leMaik, line 503:
```suggestion
      return builder.start().waitFor() == 0;
```

Duh!

PR #1119, opened 2022-01-02
Move block properties from Material to Block.

leMaik:
`solid` and `waterlogged` are block properties (Minecraft logic) and have nothing to do with the material (affects rendering). 

`opaque` is also more of a optimization for blocks that fill out the entire voxel, that could also be moved.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/block/legacy/blocks/LegacyFire.java
+++ b/chunky/src/java/se/llbit/chunky/block/legacy/blocks/LegacyFire.java
@@ -1,5 +1,6 @@
 package se.llbit.chunky.block.legacy.blocks;
 
+import se.llbit.chunky.block.Block;
 import se.llbit.chunky.block.BlockFace;
 import se.llbit.chunky.block.FinalizationState;
 import se.llbit.chunky.block.legacy.LegacyBlockUtils;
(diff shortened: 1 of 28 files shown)


chunky/src/java/se/llbit/math/NodeBasedOctree.java
```diff
@@ -143,7 +144,7 @@ public void set(Octree.Node data, int x, int y, int z) {
   }
 
   @Override
-  public Material getMaterial(int x, int y, int z, BlockPalette palette) {
+  public Block getMaterial(int x, int y, int z, BlockPalette palette) {
```
ShirleyNekoDev, line 147:
```suggestion
  public Block getBlock(int x, int y, int z, BlockPalette palette) {
```

chunky/src/java/se/llbit/chunky/block/legacy/blocks/LegacyFire.java
```diff
@@ -66,7 +67,7 @@ private static boolean isFlammable(Material material) {
         name.endsWith("_carpet");
   }
 
-  private static boolean isTopBurnable(Material material) {
+  private static boolean isTopBurnable(Block material) {
     // solid blocks like dirt and stone can burn on top but not on the side (i.e. are not flammable)
     return material.solid || isFlammable(material);
   }
```
aTom3333, line 73:
```suggestion
  private static boolean isTopBurnable(Block block) {
    // solid blocks like dirt and stone can burn on top but not on the side (i.e. are not flammable)
    return block.solid || isFlammable(block);
  }
```

chunky/src/java/se/llbit/chunky/renderer/scene/OctreeFinalizer.java
```diff
@@ -79,12 +78,12 @@ private static void hideBlocks(Octree worldTree, BlockPalette palette, int x,
   private static void processBlock(Octree worldTree, Octree waterTree, BlockPalette palette, int x,
       int cy, int z, Vector3i origin) {
     int y = cy - origin.y;
-    Material mat = worldTree.getMaterial(x, y, z, palette);
-    Material wmat = waterTree.getMaterial(x, y, z, palette);
+    Block mat = worldTree.getBlock(x, y, z, palette);
+    Block wmat = waterTree.getBlock(x, y, z, palette);
```
aTom3333, line 82:
```suggestion
    Block block = worldTree.getBlock(x, y, z, palette);
    Block wblock = waterTree.getBlock(x, y, z, palette);
```
?

aTom3333 (approved):

PR #1120, opened 2022-01-02
Make translucent blocks non-solid by default.

leMaik:
Follow-up for #1116 (e.g. lever was still solid)


ThatRedox (changes requested):
Monster spawners and double slabs don't connect to fences when they should (tested in 1.12)

ThatRedox:
Blocks that connect to fences when they should not:
* Shulker box
* Glowstone
* Sea lantern

PR #1128, opened 2022-01-09
Make scene chooser last modified have time.

ThatRedox:


leMaik:
Default sorting still doesn't work for me :thinking:

PR #1140, opened 2022-01-19
Add Nullable Annotation

NotStirred:
also moved `NotNull` into new `annotation` package

### DIFF
--- a/chunky/src/java/se/llbit/chunky/chunk/ChunkData.java
+++ b/chunky/src/java/se/llbit/chunky/chunk/ChunkData.java
@@ -1,7 +1,6 @@
 package se.llbit.chunky.chunk;
 
 import se.llbit.nbt.CompoundTag;
-import se.llbit.util.NotNull;
 
 import java.util.Collection;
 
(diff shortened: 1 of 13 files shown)


chunky/src/java/se/llbit/chunky/chunk/ChunkData.java
```diff
@@ -1,7 +1,6 @@
 package se.llbit.chunky.chunk;
 
 import se.llbit.nbt.CompoundTag;
-import se.llbit.util.NotNull;
```
leMaik, line 4: Is `@NotNull` used in this class?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -105,6 +105,8 @@
 import java.io.*;
 import java.util.*;
 import java.util.concurrent.*;
+
+import se.llbit.util.annotation.NotNull;
```
leMaik, line 109: Is this import actually used?

leMaik (approved):

leMaik:
Can't wait for the biomes PR to use this :smile:

PR #1148, opened 2022-01-24
Legacy map view optimisations

### LINKED ISSUE #1074: Pre flattening worlds should use a custom palette to improve octree loading performance
Currently they use the hashmap palette, with a `BlockPalette#put` per block, which is ludicrously slow

NotStirred:
Closes #1074
The main thing here is that the first 4096 blocks in any block palette are now legacy blocks
Those blocks can also be used by non-legacy chunks if their tag matches (thanks redox)

Palettes are created once per `RegionParser` thread now, instead of once per chunk

I also tweaked the load factor of the palette hashmap to somewhat avoid expensive collisions, given ~33000 blocks in the palette, the overhead is ~0.5MiB per palette

### DIFF
--- a/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
+++ b/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
@@ -17,12 +17,10 @@
 package se.llbit.chunky.chunk;
 
 import se.llbit.chunky.block.*;
+import se.llbit.chunky.block.legacy.LegacyBlocks;
 import se.llbit.chunky.plugin.PluginApi;
 import se.llbit.math.Octree;
-import se.llbit.nbt.CompoundTag;
-import se.llbit.nbt.IntTag;
-import se.llbit.nbt.StringTag;
-import se.llbit.nbt.Tag;
+import se.llbit.nbt.*;
 
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
@@ -67,6 +65,9 @@ public class BlockPalette {
     this.blockMap = initialMap;
     this.palette = initialList;
     this.materialProperties = getDefaultMaterialProperties();
+
+    putLegacyTags(); //must be done first as legacy blocks are index-dependant
+
     CompoundTag airTag = new CompoundTag();
     airTag.add("Name", new StringTag("minecraft:air"));
     CompoundTag stoneTag = new CompoundTag();
@@ -127,6 +128,30 @@ public class BlockPalette {
       return id;
     } finally {
       lock.unlock();
+    }
+  }
+
+  private void putLegacyTags() {
+    byte[] block = new byte[1];
+    byte[] data = new byte[1];
+    for (int i = 0; i < 256; i++) {
+      for (int j = 0; j < 16; j++) {
+        block[0] = (byte) i;
+        data[0] = (byte) j;
+        putLegacy(LegacyBlocks.getTag(0, block, data));
+      }
+    }
+  }
+
+  private void putLegacy(Tag tag) {
+    BlockSpec spec = new BlockSpec(tag);
+    blockMap.computeIfAbsent(spec, k -> palette.size());
+    try {
+      Block block = spec.toBlock();
+      applyMaterial(block);
+      palette.add(block);
+    } catch (Throwable t) {
+      palette.add(null);
     }
   }
 
--- a/chunky/src/java/se/llbit/chunky/world/Chunk.java
+++ b/chunky/src/java/se/llbit/chunky/world/Chunk.java
@@ -437,8 +437,7 @@ public class Chunk {
               int blockY = sectionMinBlockY + y;
               for (int z = 0; z < Z_MAX; z++) {
                 for (int x = 0; x < X_MAX; x++) {
-                  chunkData.setBlockAt(x, blockY, z, blockPalette.put(
-                      LegacyBlocks.getTag(offset, blocksBytes, blockDataBytes)));
+                  chunkData.setBlockAt(x, blockY, z, LegacyBlocks.legacyIdx(offset, blocksBytes, blockDataBytes));
                   offset += 1;
                 }
               }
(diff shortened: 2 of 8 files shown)


NotStirred:
erm, this test (`DeadlockTest`) does not fail on my machine, did it time out?

Nothing I've changed here should mess with that at all

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
```diff
@@ -67,6 +65,9 @@ public BlockPalette(Map<BlockSpec, Integer> initialMap, List<Block> initialList)
     this.blockMap = initialMap;
     this.palette = initialList;
     this.materialProperties = getDefaultMaterialProperties();
+
+    putLegacyTags(); //must be done first as legacy blocks are index-dependant
```
leMaik, line 69: The `palette` defaults to `initialList`, which might not be empty. In this case, this can't be guaranteed.

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
```diff
@@ -130,6 +131,30 @@ public int put(BlockSpec spec) {
...
+
+  private void putLegacy(Tag tag) {
+    BlockSpec spec = new BlockSpec(tag);
+    blockMap.computeIfAbsent(spec, k -> palette.size());
+    try {
+      Block block = spec.toBlock();
+      applyMaterial(block);
+      palette.add(block);
+    } catch (Throwable t) {
+      palette.add(null);
```
leMaik, line 154: This should never happen; if it does, we should log an error and put Air into that index instead of `null` (which I guess will crash and burn)?

chunky/src/java/se/llbit/chunky/world/Chunk.java
```diff
@@ -161,14 +161,14 @@ public synchronized boolean loadChunk(ChunkData chunkData, int yMin, int yMax) {
...
-    loadSurface(data, chunkData, yMin, yMax);
+    loadSurface(data, blockPalette, biomePalette, chunkData, yMin, yMax);
     biomesTimestamp = dataTimestamp;
 
     world.chunkUpdated(position);
     return true;
   }
 
-  private void loadSurface(@NotNull Tag data, ChunkData chunkData, int yMin, int yMax) {
+  private void loadSurface(@NotNull Tag data, BlockPalette blockPalette, BiomePalette biomePalette, ChunkData chunkData, int yMin, int yMax) {
```
leMaik, line 171: Should we add a `ChunkDataContainer` that contains the chunk data, biome palette and block palette?

chunky/src/java/se/llbit/chunky/world/region/RegionParser.java
```diff
@@ -53,6 +57,11 @@ public RegionParser(WorldMapLoader loader, RegionQueue queue, MapView mapView) {
   }
 
   @Override public void run() {
+    //setting load factor to decrease expensive hash collisions between BlockSpec objects
```
leMaik, line 60: Maybe we could take a look at [jo-nbt](https://github.com/llbit/jo-nbt) to see if there are any possible optimizations? The `equals` and `hashCode` methods there look good to me, though.

leMaik (changes requested):

leMaik:
Looks good, I also restarted the action. :+1:

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
NotStirred: Not true, there are sanity checks in blocks that throw a LOT (checking that a furnace isn't facing down for example)
This attempts to create every possible legacy block state, so there are bound to be such invalid states

NotStirred:
The failing build makes no sense, that test is for the renderer, and doesn't interact with loading in any way

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
leMaik: Could you comment that for future readers?

leMaik:
Somehow it fails consistently, though. Do the tests fail locally, too? I don't even know if the test actually tests for the bug it used to test for anymore.

NotStirred:
No they do not fail locally

ThatRedox:
image
Profiling the deadlock test, it seems like a significant amount of time is spent creating the new buffered scene in `DefaultRenderManager` and initializing the block palette. On master the time spent creating the scene is much, much smaller.

NotStirred:
I see, the test is creating a LOT of scene objects. I'll push a simple solution, one moment

NotStirred:
turns out having null blocks in the palette would modify the save format, which is not something I would like to do
I'll see if I can think of a nice solution

NotStirred:
I suppose I could add an unknown block there actually

ThatRedox:
This branch actually fails to load this world.
[1148 test.zip](https://github.com/chunky-dev/chunky/files/7965426/1148.test.zip)
```
Exception in thread "Scene Manager" java.lang.NullPointerException
	at se.llbit.chunky.chunk.BlockPalette.applyMaterial(BlockPalette.java:244)
	at java.util.ArrayList.forEach(ArrayList.java:1259)
	at se.llbit.chunky.chunk.BlockPalette.applyMaterials(BlockPalette.java:255)
	at se.llbit.chunky.renderer.scene.Scene.loadOctree(Scene.java:2218)
	at se.llbit.chunky.renderer.scene.Scene.loadScene(Scene.java:613)
	at se.llbit.chunky.renderer.scene.SynchronousSceneManager.loadScene(SynchronousSceneManager.java:150)
	at se.llbit.chunky.renderer.scene.AsynchronousSceneManager.lambda$loadScene$0(AsynchronousSceneManager.java:104)
	at se.llbit.chunky.renderer.scene.AsynchronousSceneManager.run(AsynchronousSceneManager.java:80)
```

NotStirred:
yes that is what this is referring to https://github.com/chunky-dev/chunky/pull/1148#issuecomment-1023051668

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
leMaik: At least we shouldn't catch `Throwable` but maybe introduce some `IllegalBlockStateException` or catch `RuntimeException` here for now.

PR #1159, opened 2022-01-31
Sort worlds by last modified time.

ThatRedox:
image


leMaik:
Looks great! Not sure about the default sort order though. Maybe we should just remember the previously selected order instead. Same for scene loading. I'll do that tomorrow or so :)

PR #1164, opened 2022-02-07
Use the correct scene directory throughout the project

NotStirred:
There were several places where we assumed the scene directory was the default location, which is
not the case when using the `-scene-dir` argument.

This PR removes `resolveSceneDirectory()` entirely, as that assumed the default location

In my testing, this works correctly both with no `-scene-dir` argument, and a specified one.
I have not fiddled with this part of chunky before, so some testing from someone more experienced would be appreciated

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
@@ -139,6 +139,7 @@ public class AsynchronousSceneManager extends Thread implements SceneManager {
       }
     });
   }
+
   @PluginApi
   @Deprecated
   @Override public void saveScene() {
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
@@ -111,6 +111,7 @@ public class AsynchronousSceneManager extends Thread implements SceneManager {
       }
     });
   }
+
   @PluginApi
   @Deprecated
   @Override public void loadScene(String name) {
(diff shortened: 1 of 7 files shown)


leMaik (approved):

ThatRedox:
Changing the scene manager API will break (a lot of) things. Maybe keep the old methods as deprecated for now?

NotStirred:
would you want me to add the back the methods with their old signatures, unused, and marked deprecated? Or do you have another idea in mind

ThatRedox:
Yeah adding back the methods with their old signatures and mark them deprecated. We can remove the methods in 2.6 snapshots.

chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
```diff
@@ -98,6 +99,20 @@ public void setOnChunksLoaded(Runnable onChunksLoaded) {
...
+        sceneManager.loadScene(sceneDirectory, name);
+      } catch (IOException e) {
+        Log.warn("Could not load scene.\nReason: " + e.getMessage());
+      } catch (InterruptedException e) {
+        Log.warn("Scene loading was interrupted.");
+      }
+    });
+  }
+  @PluginApi
+  @Deprecated
```
leMaik, line 115:
```suggestion

  @PluginApi
  @Deprecated
```

chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
```diff
@@ -113,6 +129,18 @@ public void setOnChunksLoaded(Runnable onChunksLoaded) {
...
+    enqueueTask(() -> {
+      try {
+        sceneManager.saveScene(sceneDirectory);
+      } catch (InterruptedException e) {
+        Log.warn("Scene saving was interrupted.");
+      }
+    });
+  }
+  @PluginApi
+  @Deprecated
```
leMaik, line 143:
```suggestion

  @PluginApi
  @Deprecated
```

PR #1171, opened 2022-02-17
UI package refactoring

ShirleyNekoDev:
- grouped UI elements (classes and their FXMLs) into matching packages
- fixed number validation not showing invalid state (and rewritten validator to check more cases)
- dropped children and content tags from FXML (they are redundant)


ShirleyNekoDev:
~ rebased onto master

chunky/src/res/se/llbit/chunky/ui/dialogs/ConfirmResetPopup.fxml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
leMaik, line 1: Couldn't this popup just be a confirm dialog, instead of having its own implementation?

leMaik (approved):

Issue #1177, opened 2022-02-24
Chunky changes the postprocessing filter while rendering only if the current SPP is an even number


Peregrine05:
While rendering, if the postprocessing filter is changed, Chunky will only update the display when the current SPP is an even number. If it is an odd number, then Chunky must render one more SPP to visibly change the postprocessing filter. Sometimes Chunky will not change the postprocessing filter even if the current SPP is an even number. Chunky must render one more SPP to change the filter.

ShirleyNekoDev:
Probable cause is that chunky swaps between a front and a back buffer for the preview, but post-processing is only applied on the back buffer.
A better interface for the chunky render canvas which encapsulates the sample buffer, alpha buffer and preview buffers would help to prevent such issues in the future.

PR #1178, opened 2022-02-24
Refactor sample buffer access in post-processors

### LINKED ISSUE #1176: Exporting as PFM applies postprocessing even if set to None
When exporting the render to PFM, Chunky always exports the render with the Gamma correction postprocessing filter applied. Setting the filter to None or any other setting still makes Chunky export the PFM with Gamma correction. Using ShirleyNeko's [ImageMagick plugin](https://github.com/ShirleyNekoDev/Chunky-MagickExportPlugin) will allow exporting of PFM files without postprocessing.

ShirleyNekoDev:
~~Fixes post-processing side-effects on PFM export~~


ShirleyNekoDev:
This modifies a @PluginApi, maybe wait for a refactoring of the canvas buffers as mentioned in #1177 before changing it?

PR #1192, opened 2022-03-11
Add a new setting to choose the aperture shape for custom bokeh

aTom3333:
I added a new button in camera tab that let's someone choose a mask image that defines the shape of the aperture. The mask image must be square (checked in the code) and in grayscale (not checked in the code, only one channel is used, if the mask image has colors, nothing bad will happen).
Custom aperture shape is interesting for bokeh effect.

Default aperture (equivalent to having a circle):
bokeh-default-10000

Hexagonal aperture:
hexagon-aperture
bokeh-hexagon-10000

Star aperture:
star-aperture
bokeh-star-10000

Gaussian aperture:
gaussian-aperture
bokeh-gaussian-10000

In the mask image, white means the light can go through and black means the light can't. We can use gray, in which case the amount of light that can go through is proportional to brightness of the gray


aTom3333:
Another one as bonus because I have nothing better to do (and also because I wanted to check that an aperture with the center obstructed would work.

Ring aperture:
ring-aperture
bokeh-ring-10000

leMaik:
Nice! Should we add a few built-in shapes? We could simply load them from resources bundled in the jar... This would require a way to load shapes without files being involved, though.

jackjt8:
A few built-in shapes would be nice if it can be figured out.

aTom3333:
Any idea of how to do that ui-wise? Even now I'm not satisfied with the UI, there is no indication that a mask has been added

leMaik:
@aTom3333 Dropdown with star, circle, …, custom?

leMaik:
It's amazing how you keep delivering some of the best features of 2.5 :tada: :clap:

PR #1197, opened 2022-03-12
Fix empty 16x16x16 sections of identical blocks.

### LINKED ISSUE #1191: Multiple empty chunks on the renders
Multiple chunk-sized holes are visible on the renders. Probably because those chunks are only made up of stone. This is surely an optimization of Chunky but these empty chunks are still visible on the rendering which of course is unintended.
Example

Here is the Minecraft region with the bug (1.18.1):
[r.0.0.zip](https://github.com/chunky-dev/chunky/files/8226819/r.0.0.zip)
Here is the camera position and orientation:
Camera position+orientation

ThatRedox:
Closes #1191 
image

Looks like sections that are optimized by not including a block data array and rather just having a single block palette with a non-air block were ignored.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/world/Chunk.java
+++ b/chunky/src/java/se/llbit/chunky/world/Chunk.java
@@ -278,6 +278,23 @@ public class Chunk {
                 }
               }
             }
+          } else {
+            // Single block palette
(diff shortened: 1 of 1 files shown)


leMaik:
Thank you for figuring this out! :pray:

PR #1215, opened 2022-04-10
Add `High Quality` sun sampling mode. Fix issue where enabling draw sun but not enabling sunlight would draw a sun with no contribution to the scene.

### LINKED ISSUE #1184: Add a way to disable NEE for the sun.
While sunlit scenes without sun NEE this would take much longer to converge, would this fix the absence of caustics in non-still water?
It would also be a nice feature for those who want to experiment with Chunky and path tracing.

ThatRedox:
Closes #1184.
This adds a `Sun Luminosity` setting which changes the true brightness of the sun when intersected with organically and a new sun sampling mode which does both sun samples and allows for sun intersections for diffuse materials. This new sun sampling mode allows for visual effects like caustics (with much render time):
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -1,5 +1,5 @@
-/* Copyright (c) 2012-2021 Jesper Öqvist <jesper@llbit.se>
- * Copyright (c) 2012-2021 Chunky contributors
+/* Copyright (c) 2012-2022 Jesper Öqvist <jesper@llbit.se>
+ * Copyright (c) 2012-2022 Chunky contributors
  *
  * This file is part of Chunky.
  *
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -2898,6 +2898,22 @@ public class Scene implements JsonSerializable, Refreshable {
     emittersEnabled = json.get("emittersEnabled").boolValue(emittersEnabled);
     emitterIntensity = json.get("emitterIntensity").doubleValue(emitterIntensity);
 
+    if (json.get("sunSamplingStrategy").isUnknown()) {
+      boolean sunSampling = json.get("sunEnabled").boolValue(false);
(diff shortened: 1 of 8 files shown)


leMaik:
Nice! :)

> Fix issue where enabling draw sun but not enabling sunlight would draw a sun with no contribution to the scene.

I always assumed this would be a feature so that one can use sky light and emitters only but still have a visual sun?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2896,7 +2897,13 @@ public synchronized void importFromJson(JsonObject json) {
     saveSnapshots = json.get("saveSnapshots").boolValue(saveSnapshots);
     emittersEnabled = json.get("emittersEnabled").boolValue(emittersEnabled);
     emitterIntensity = json.get("emitterIntensity").doubleValue(emitterIntensity);
-    sunEnabled = json.get("sunEnabled").boolValue(sunEnabled);
+
+    if (json.get("sunEnabled").boolValue(false)) {
+      sunSamplingStrategy = SunSamplingStrategy.Fast;
+    } else {
+      sunSamplingStrategy = SunSamplingStrategy.valueOf(json.get("sunSamplingStrategy").asString(SunSamplingStrategy.Fast.getId()));
```
leMaik, line 2904: Does this work correctly if `sunEnabled` is true and `sunSamplingStrategy` is not present, ie. existing scenes with sun?

leMaik (approved):

jackjt8:
> Nice! :)
> 
> > Fix issue where enabling draw sun but not enabling sunlight would draw a sun with no contribution to the scene.
> 
> I always assumed this would be a feature so that one can use sky light and emitters only but still have a visual sun?

Correct. It is often used to replace the sun in a skymap.

jackjt8:
Compared to `2.5.0-SNAPSHOT.124.ge9d7e123` (obtained today)

Master/current nightly - 441,874 SPS | 46 s

This PR:
- SSS:OFF/NON_LUMINOUS ~~632,919 SPS | 32 s
- SSS:FAST 438,496 SPS | 46 s (approx. matching master)
- SSS:HIGH_QUALITY 497,111 SPS | 41 s

---

I do not see a performance regression from this change.

PR #1225, opened 2022-04-16
Add trivial implementation of 3d biomes and 3d blur (off by default)

NotStirred:
In the future it would be great to add a better implementation for performance

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -254,7 +254,6 @@ public class Scene implements JsonSerializable, Refreshable {
 
   protected boolean biomeColors = true;
   protected boolean biomeBlending = true;
-  protected boolean use3dBiomes = false;
   protected boolean transparentSky = false;
   protected boolean renderActors = true;
   protected Collection<ChunkPosition> chunks = new ArrayList<>();
@@ -367,6 +366,11 @@ public class Scene implements JsonSerializable, Refreshable {
   private String bvhImplementation = PersistentSettings.getBvhMethod();
 
   /**
+   * The BiomeStructure implementation to use
+   */
+  private String biomeStructureImplementation = PersistentSettings.getBiomeStructureImplementation();
+
+  /**
    * Additional data that is associated with a scene, this can be used by plugins
    */
   private JsonObject additionalData = new JsonObject();
@@ -392,11 +396,6 @@ public class Scene implements JsonSerializable, Refreshable {
     worldOctree = new Octree(octreeImplementation, 1);
     waterOctree = new Octree(octreeImplementation, 1);
     emitterGrid = null;
-
-    //TODO: use selected from in GUI by default
-    grassTexture = new Trivial2dBiomeStructureImpl();
-    foliageTexture = new Trivial2dBiomeStructureImpl();
-    waterTexture = new Trivial2dBiomeStructureImpl();
   }
 
   /**
@@ -886,6 +885,8 @@ public class Scene implements JsonSerializable, Refreshable {
     Set<ChunkPosition> loadedChunks = new HashSet<>();
     int numChunks = 0;
 
+    BiomeStructure.Factory biomeStructureFactory = BiomeStructure.get(this.biomeStructureImplementation);
+
     try (TaskTracker.Task task = taskTracker.task("(1/6) Loading regions")) {
       task.update(2, 1);
 
@@ -903,6 +904,11 @@ public class Scene implements JsonSerializable, Refreshable {
       palette = new BlockPalette();
       worldOctree = new Octree(octreeImplementation, requiredDepth);
       waterOctree = new Octree(octreeImplementation, requiredDepth);
+
+      grassTexture = biomeStructureFactory.create();
+      foliageTexture = biomeStructureFactory.create();
+      waterTexture = biomeStructureFactory.create();
+
       if(emitterSamplingStrategy != EmitterSamplingStrategy.NONE)
         emitterGrid = new Grid(gridSize);
 
@@ -961,6 +967,7 @@ public class Scene implements JsonSerializable, Refreshable {
     Set<ChunkPosition> legacyChunks = new HashSet<>();
 
     Position2IntStructure biomePaletteIdxStructure;
+    boolean use3dBiomes = biomeStructureFactory.is3d();
     if (use3dBiomes) {
       biomePaletteIdxStructure = new Position3d2IntPackedArray();
     } else {
@@ -1358,16 +1365,6 @@ public class Scene implements JsonSerializable, Refreshable {
     actors.trimToSize();
     palette.unsynchronize();
 
-    if(use3dBiomes) {
-      grassTexture = new Trivial3dBiomeStructureImpl();
-      foliageTexture = new Trivial3dBiomeStructureImpl();
-      waterTexture = new Trivial3dBiomeStructureImpl();
-    } else {
-      grassTexture = new Trivial2dBiomeStructureImpl();
-      foliageTexture = new Trivial2dBiomeStructureImpl();
-      waterTexture = new Trivial2dBiomeStructureImpl();
-    }
-
     try (TaskTracker.Task task = taskTracker.task("(4/6) Finalizing octree")) {
 
       worldOctree.startFinalization();
@@ -1713,14 +1710,6 @@ public class Scene implements JsonSerializable, Refreshable {
     }
   }
 
-  public void setUse3dBiomes(boolean value) {
-    if (value != use3dBiomes) {
-      use3dBiomes = value;
-
-      refresh();
-    }
-  }
-
   /**
    * Center the camera over the loaded chunks
    */
@@ -1815,10 +1804,6 @@ public class Scene implements JsonSerializable, Refreshable {
 
   public boolean biomeBlendingEnabled() {
     return biomeBlending;
-  }
-
-  public boolean using3dBiomes() {
-    return use3dBiomes;
   }
 
   /**
@@ -3429,6 +3414,15 @@ public class Scene implements JsonSerializable, Refreshable {
   public void setBvhImplementation(String bvhImplementation) {
     this.bvhImplementation = bvhImplementation;
   }
+
+  public String getBiomeStructureImplementation() {
+    return biomeStructureImplementation;
+  }
+
+  public void setBiomeStructureImplementation(String biomeStructureImplementation) {
+    this.biomeStructureImplementation = biomeStructureImplementation;
+  }
+
 
   @PluginApi
   public Octree getWorldOctree() {
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -1521,10 +1521,7 @@ public class Scene implements JsonSerializable, Refreshable {
               for (int z = 0; z < 16; ++z) {
                 int wz = cp.z * 16 + z;
 
-                Integer id = biomePaletteIdxStructure.get(wx, 0, wz);
-                if (id == null) {
-                  continue;
-                }
+                int id = biomePaletteIdxStructure.get(wx, 0, wz);
                 Biome biome = biomePalette.get(id);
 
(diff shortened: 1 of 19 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -956,7 +954,13 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
 
     Set<ChunkPosition> nonEmptyChunks = new HashSet<>();
     Set<ChunkPosition> legacyChunks = new HashSet<>();
-    Heightmap biomePaletteIdxMap = new Heightmap();
+
+    Position2IntStructure biomePaletteIdxStructure;
+    if (using3dBiomes) {
+      biomePaletteIdxStructure = new Position3d2IntPackedArray();
+    } else {
+      biomePaletteIdxStructure = new Position2d2IntPackedArray();
```
leMaik, line 962: The most complicated name for a bitmap so far :rofl:

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2146,8 +2208,9 @@ private synchronized void saveOctree(RenderContext context, TaskTracker taskTrac
 
       boolean saved = false;
       try (DataOutputStream out = new DataOutputStream(new FastBufferedOutputStream(new GZIPOutputStream(context.getSceneFileOutputStream(fileName))))) {
-        OctreeFileFormat.store(out, worldOctree, waterOctree, palette,
-            grassTexture, foliageTexture, waterTexture);
+        //TODO: READD THIS
+//        OctreeFileFormat.store(out, worldOctree, waterOctree, palette,
+//            grassTexture, foliageTexture, waterTexture);
```
leMaik, line 2213: :bell: Reminder?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2216,9 +2279,10 @@ private synchronized boolean loadOctree(RenderContext context, TaskTracker taskT
         worldOctree = data.worldTree;
         worldOctree.setTimestamp(fileTimestamp);
         waterOctree = data.waterTree;
-        grassTexture = data.grassColors;
-        foliageTexture = data.foliageColors;
-        waterTexture = data.waterColors;
+        //TODO: READD THIS
+//        grassTexture = data.grassColors;
+//        foliageTexture = data.foliageColors;
+//        waterTexture = data.waterColors;
```
leMaik, line 2285: :bell: Reminder?

chunky/src/java/se/llbit/chunky/ui/render/tabs/GeneralTab.java
```diff
@@ -273,6 +275,10 @@ public GeneralTab() throws IOException {
     biomeColors.selectedProperty().addListener((observable, oldValue, newValue) -> {
       scene.setBiomeColorsEnabled(newValue);
     });
+    using3dBiomes.setTooltip(new Tooltip("Attempt to load biomes as 3d (Added in 1.18+)."));
+    using3dBiomes.selectedProperty().addListener((observable, oldValue, newValue) -> {
+      scene.setUsing3dBiomes(newValue);
```
leMaik, line 280: Should this trigger the reload chunks dialog?

chunky/src/java/se/llbit/chunky/ui/render/tabs/GeneralTab.java
```diff
@@ -273,6 +275,10 @@ public GeneralTab() throws IOException {
     biomeColors.selectedProperty().addListener((observable, oldValue, newValue) -> {
       scene.setBiomeColorsEnabled(newValue);
     });
+    using3dBiomes.setTooltip(new Tooltip("Attempt to load biomes as 3d (Added in 1.18+)."));
```
leMaik, line 278:
```suggestion
    using3dBiomes.setTooltip(new Tooltip("Attempt to load 3d biomes (added in 1.15)."));
```

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -256,6 +249,7 @@ public class Scene implements JsonSerializable, Refreshable {
   /** Controls how much the fog color is blended over the sky/skymap. */
   protected double skyFogDensity = 1;
 
+  protected boolean using3dBiomes = false;
```
leMaik, line 252:
```suggestion
  protected boolean use3dBiomes = false;
```

Also, remember to save and load that value from the scene description :bulb:

leMaik (changes requested):

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
NotStirred: This will actually be somewhat annoying, as the octree format will have to change
Perhaps an octree internal "2d/3d texture format" would be good, as then we can have implementation defined loading without adjusting the octree format again. Plugins could then register them on load and be asked to load for their registered formats

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1034,12 +1038,14 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
...
-          for (int cx = 0; cx < 16; ++cx) {
-            int wx = cx + wx0;
-            int biomePaletteIdx = biomeData.getBiome(cx, 0, cz); // TODO add vertical biomes support (1.15+)
-            biomePaletteIdxMap.set(biomePaletteIdx, wx, wz);
+        for (int y = chunkData.minY(); y < chunkData.maxY(); y++) {
+          for (int cz = 0; cz < 16; ++cz) {
+            int wz = cz + wz0;
+            for (int cx = 0; cx < 16; ++cx) {
+              int wx = cx + wx0;
+              int biomePaletteIdx = biomeData.getBiome(cx, y, cz); // TODO add vertical biomes support (1.15+)
```
aTom3333, line 1046: This TODO is no longer relevant as this is what you've just done or is there still something to do?

NotStirred: good point - yeah this can be removed. I'll mark this pr as draft for now as there are a lot of silly things around

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1370,51 +1372,111 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
...
+                    int wx = cp.x * 16 + sx;
+                    for (int sz = z - 1; sz <= z + 1; ++sz) {
+                      int wz = cp.z * 16 + sz;
+                      for (int sy = y - 1; sy < y + 1; sy++) {
+                        int wy = sectionY * 16 + sy;
+                        ChunkPosition ccp = ChunkPosition.get(wx >> 4, wz >> 4);
+                        if (nonEmptyChunks.contains(ccp)) {
+                          nsum += 1;
+                          Integer id = biomePaletteIdxStructure.get(wx, wy, wz);
+                          if (id == null) {
```
aTom3333, line 1395: If I'm not mistaken, get returns an int so the boxen and test for null is not needed

chunky/src/java/se/llbit/math/structures/Position2ReferenceStructure.java
```diff
@@ -0,0 +1,11 @@
+package se.llbit.math.structures;
+
+public interface Position2ReferenceStructure<T> {
```
aTom3333, line 3: I think a specialized Position2ColorStructure interface would lead more leeway to optimize the grass/foliage/water color in the future. With this generic interface, I don't how I could implement something that stores the colors in a flatten float[] (instead of in a float[][])

NotStirred: It should work fine with this interface right? Can just copy into the falttened one. If we had varargs generics it might be better but it certainly works

aTom3333: Yeah sorry, I was a bit confused about the fact that it was a generic interface.
In my head, because the interface was generic, the implementation had to be generic as well and I couldn't see how an implementation that is generic could take advantage of the flattening (because T could be something else than an array) but what I want is just a non generic implementation that implements Position2ReferenceStructure<float[]>.

chunky/src/java/se/llbit/chunky/ui/render/tabs/GeneralTab.java
NotStirred: atm even `biomeColors` doesn't - should it? I can make them all follow the same behaviour

leMaik: What does biomeColors not do?

NotStirred: replied to the wrong thing apparently - meant `Should this trigger the reload chunks dialog?`

leMaik: It probably should, but switching biome colors even requires reloading the chunks. Is there a dialog for that, too? Probably from the emitter sampling?

NotStirred:
Still need to do https://github.com/chunky-dev/chunky/pull/1225#discussion_r851571870. Do you want me to add it to the json serialisation for the scene? It shouldn't affect anything but the `Chunks need to be reloaded dialog` if it gets out of sync (at least afaict atm)

NotStirred:
Currently the only problem with this PR I see is that the biome palette indices are loaded into a structure which isn't yet implementable by anything else. Ideally we'd just have the structures be generic and it'd just work but java generics can't handle primitives :(

chunky/src/java/se/llbit/chunky/ui/render/tabs/AdvancedTab.java
```diff
@@ -190,6 +192,24 @@ public PictureExportFormat fromString(String string) {
...
+      biomeStructureTooltipBuilder.append(entry.getKey());
+      biomeStructureTooltipBuilder.append(": ");
+      biomeStructureTooltipBuilder.append("Description"); //entry.getValue().getDescription());
+      biomeStructureTooltipBuilder.append('\n');
+    }
+    biomeStructureTooltipBuilder.append("Requires reloading chunks to take effect.");
+    biomeStructureImplementation.getItems().addAll(biomeStructureNames);
+    biomeStructureImplementation.getSelectionModel().selectedItemProperty()
+      .addListener((observable, oldvalue, newvalue) -> {
+        scene.setBiomeStructureImplementation(newvalue);
```
leMaik, line 208: A prompt (as for the emitter grid implementation) would be nice in a follow-up PR :+1:

chunky/src/java/se/llbit/chunky/ui/render/tabs/GeneralTab.java
```diff
@@ -140,6 +133,11 @@ public GeneralTab() throws IOException {
       loadOtherEntities.setSelected(preferences.shouldLoadClass(null));
     }
     biomeColors.setSelected(scene.biomeColorsEnabled());
+    if(!scene.biomeColorsEnabled()) {
+      biomeBlending.setDisable(true);
+    }
```
leMaik, line 138: Does this get updated if biome colors are enabled later?

```suggestion
    biomeBlending.setDisable(!scene.biomeColorsEnabled());
```

leMaik (approved):

chunky/src/java/se/llbit/chunky/ui/render/tabs/GeneralTab.java
NotStirred: Yes, this is just the initial value, see https://github.com/chunky-dev/chunky/pull/1225/files/01941b8ce6f2a018dd32920d7cd5b548c27a688d#diff-508504c789f61f4751ea96ead9517f30b1a68bbb68a7fecade06614ff5d37575R275

NotStirred: Though I agree with your logic :+1:

PR #1229, opened 2022-04-16
Update Chunky copyright

Peregrine05:
Update the copyright shown in the About tab.


Peregrine05:
Actually, I forgot that I was running the latest snapshot. This is already fixed in Chunky 2.4.2. Should this be closed?

PR #1232, opened 2022-04-17
Update material defaults

### LINKED ISSUE #1033: 21w37a
- [x] New biomes (dripstone cave, grove, lofty peaks, lush caves, meadow, snowcapped peaks, snowy slopes, stony peaks)
- [ ] Enchanting tables now emit a low amount of light (light level 7)

https://www.minecraft.net/en-us/article/minecraft-snapshot-21w37a

Peregrine05:
Resolves #1033 

Water specular now 0.255 from 0.12;
Torch and wall torch emittance now 5.0 from 50.0;
Soul torch and Soul wall torch emittance now 3.5 from 35.0;
Added Redstone wall torch emittance of 1.0;
Added Enchanting table emittance of 0.6;

### DIFF
--- a/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
+++ b/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
@@ -1,4 +1,4 @@
-/* Copyright (c) 2019-2021 Chunky contributors
+/* Copyright (c) 2019-2022 Chunky contributors
  *
  * This file is part of Chunky.
  *
@@ -236,7 +236,7 @@ public class BlockPalette {
     materialProperties.put(
         "minecraft:water",
         block -> {
-          block.specular = 0.12f;
+          block.specular = 0.255f;
           block.ior = 1.333f;
           block.refractive = true;
         });
@@ -338,11 +338,14 @@ public class BlockPalette {
     materialProperties.put("minecraft:redstone_torch", block -> {
       block.emittance = 1.0f;
     });
+    materialProperties.put("minecraft:redstone_wall_torch", block -> {
+      block.emittance = 1.0f;
+    });
     materialProperties.put("minecraft:torch", block -> {
-      block.emittance = 50.0f;
+      block.emittance = 5.0f;
     });
     materialProperties.put("minecraft:wall_torch", block -> {
-      block.emittance = 50.0f;
+      block.emittance = 5.0f;
     });
     materialProperties.put("minecraft:fire", block -> {
       block.emittance = 1.0f;
@@ -440,21 +443,24 @@ public class BlockPalette {
       block.emittance = 0.6f;
     });
     materialProperties.put("minecraft:soul_fire_torch", block -> { // MC 20w06a-20w16a
-      block.emittance = 35.0f;
+      block.emittance = 3.5f;
     });
     materialProperties.put("minecraft:soul_torch", block -> { // MC >= 20w17a
-      block.emittance = 35.0f;
+      block.emittance = 3.5f;
     });
     materialProperties.put("minecraft:soul_fire_wall_torch", block -> { // MC 20w06a-20w16a
-      block.emittance = 35.0f;
+      block.emittance = 3.5f;
     });
     materialProperties.put("minecraft:soul_wall_torch", block -> { // MC >= 20w17a
-      block.emittance = 35.0f;
+      block.emittance = 3.5f;
     });
     materialProperties.put("minecraft:soul_fire", block -> {
       block.emittance = 0.6f;
     });
     materialProperties.put("minecraft:crying_obsidian", block -> {
+      block.emittance = 0.6f;
+    });
+    materialProperties.put("minecraft:enchanting_table", block -> {
       block.emittance = 0.6f;
     });
     materialProperties.put("minecraft:respawn_anchor", block -> {
--- a/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
+++ b/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
@@ -342,10 +342,10 @@ public class BlockPalette {
       block.emittance = 1.0f;
     });
     materialProperties.put("minecraft:torch", block -> {
-      block.emittance = 5.0f;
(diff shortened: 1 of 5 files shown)


chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
```diff
@@ -338,11 +338,14 @@ public static Map<String, Consumer<Block>> getDefaultMaterialProperties() {
     materialProperties.put("minecraft:redstone_torch", block -> {
       block.emittance = 1.0f;
     });
+    materialProperties.put("minecraft:redstone_wall_torch", block -> {
+      block.emittance = 1.0f;
+    });
```
leMaik, line 343: Now that I see it… shouldn't the torches take their on/off state into account? Is this a bug you'd like to fix?

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
```diff
@@ -236,7 +236,7 @@ public static Map<String, Consumer<Block>> getDefaultMaterialProperties() {
     materialProperties.put(
         "minecraft:water",
         block -> {
-          block.specular = 0.12f;
+          block.specular = 0.255f;
```
leMaik, line 239: Looks _way_ better, I thought we already changed this :+1:

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
```diff
@@ -338,11 +338,14 @@ public static Map<String, Consumer<Block>> getDefaultMaterialProperties() {
     materialProperties.put("minecraft:redstone_torch", block -> {
       block.emittance = 1.0f;
     });
+    materialProperties.put("minecraft:redstone_wall_torch", block -> {
+      block.emittance = 1.0f;
+    });
     materialProperties.put("minecraft:torch", block -> {
-      block.emittance = 50.0f;
+      block.emittance = 5.0f;
```
leMaik, line 345: :+1: for reducing the high brightness.

The emittance values are "normalized" currently, such that a Minecraft emittance value of 15 is 1.0 in Chunky for a 16x16x16 (area = 1536) units block (plus some rounding to get "nice" values for no reason).The torch is 2x2x10, i.e. has an area of 88 units. By that calculation, the torch should have an emittance of 17.45 in Chunky.

If you only take the high-emittance (i.e. yellow) part of the torch into account, that area is roughly 20, so an emittance value of 50 would actually be too low (should be 76.8).

Same considerations for the other torches. Since these changes affect every render created with future Chunky versions (unless the materials are tweaked manually), we should be careful about them.

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
```diff
@@ -440,23 +443,26 @@ public static Map<String, Consumer<Block>> getDefaultMaterialProperties() {
...
+      block.emittance = 3.5f;
     });
     materialProperties.put("minecraft:soul_fire", block -> {
       block.emittance = 0.6f;
     });
     materialProperties.put("minecraft:crying_obsidian", block -> {
       block.emittance = 0.6f;
     });
+    materialProperties.put("minecraft:enchanting_table", block -> {
+      block.emittance = 0.6f;
```
leMaik, line 464: Crying obsidian has a light level of _10_ and emittance _0.6_. Enchanting table has a light level of 7, which would be 0.46 (round as you see fit) in Chunky.

leMaik (changes requested):

Peregrine05:
How does the emittance "multiplier"? of 13 factor in?

leMaik:
@Peregrine05 Every emittance value gets multiplied by that global multiplier. The "calculation" I mentioned previously is mostly to make the emitters "compatible" relative to each other.

(repeating this from Discord for the sake of completeness)

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
Peregrine05: We noticed that torches outshone outshone other emitters, even though the emittance values are normalized, and even though the torch emittance value is techincally too low. How can that be?

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
Peregrine05: Maybe I can try fixing that in a future PR.

Peregrine05:
If 1 is the emittance value for a full-block emitter whose in-game light level is 15, then why is there a default emittance multiplier of 13? If the calculated value is 1, then why multiply it by 13? Is it because the emitters would be too dim otherwise, seeing as the calculation is simply to normalize the emittance values based on in-game light level and block surface area?

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
Peregrine05: Actually, torches have an in-game light level of 14, so the calculation turns out to be 71.68.

leMaik: Undo for now and wait for community feedback?

Peregrine05: Yes, maybe a post on Reddit could be made, with some comparisons.

leMaik:
@Peregrine05 Exactly. The calculation above is to make the emittance values good relative to each other, the 13 is to make everything bright enough.

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
Peregrine05: Torches and sould torches have been reset to their default values.

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
Peregrine05: Enchanting table emittance now set to 0.5

leMaik (approved):

PR #1240, opened 2022-04-22
Fix crash if loading chunk data times out.

leMaik:


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -989,10 +989,11 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
         loadedChunks.add(cp);
 
         try {
-          nextChunkDataTask.get(50, TimeUnit.MILLISECONDS);
+          nextChunkDataTask.get(1, TimeUnit.SECONDS);
```
ThatRedox, line 992: See if we can just remove the timeout?

PR #1241, opened 2022-04-22
Fix wrong scene path after saving it.

### LINKED ISSUE #1093: Renaming and saving a scene does not create new directory
2.3.0, 2.4.0, 2.4.1 - Works
2.5.0-11-g7b67ad1d - Fails...

Expected behaviour - In previous versions: Creating a scene with `scenename_a` and saving produces a directory within `scenes` called `scenename_a` with scene files as `scenename_a.<something>`. Upon renaming the scene with say `scenename_b` and saving a new directory called `scenename_b` and all scene files created during a save action are created in the new directory with the new files names. Chunky can open these scenes.

What's going wrong - In Chunky 2.5.0-11-g7b67ad1d, and maybe previous development versions, The act of renaming a scene and saving it does not create a new directory for `scenename_b`. Instead all files for `scenename_b` are stored within the directory for  `scenename_a`. This breaks loading of these new scenes.

leMaik:
Closes #1093


leMaik:
_Finally_ :partying_face:

PR #1243, opened 2022-04-23
Fix biomeData NPE on SimpleChunkData clear

ShirleyNekoDev:
- fix biomeData NPE on SimpleChunkData clear
  (Had this happen in some chunks of my worlds, which caused them to not load.)
- also moved member initialization from constructor to declaration (removed trivial constructor)


leMaik:
thank you :)
regarding `Arrays.fill` performance: it seems to be a loop in JDK 17; but maybe the jvm will be extra smart about it one day.

PR #1249, opened 2022-04-29
22w17a - Update Mangrove Roots Model

### LINKED ISSUE #1248: 22w17a - Mangrove Roots model incorrect
Mangrove roots had previously had an incorrect model, which caused the pixels of the inner planes of the Mangrove roots texture not to line up with the pixels of the outer planes. This was fixed in Java snapshot 22w17a; issue [MC-250075](https://bugs.mojang.com/browse/MC-250075). This fix has not been reflected in Chunky, so the pixels of the inner planes of the mangrove roots texture do not line up with the pixels on the outer planes.

Minecraft 22w17a:

Chunky Latest snapshot:
Mangrove roots incorrect

Peregrine05:
Fixes #1248 

The pixels of the inner planes of the mangrove roots texture now line up with the pixels on the outer planes.

The same fix for Minecraft, "This is a very easy fix - all of the "0.8" values in the model should be changed to "0" and every "15.2" to "16"," ([MC-250075](https://bugs.mojang.com/browse/MC-250075)), also applied to Chunky.

Minecraft 22w17a:

Chunky latest snapshot:
Mangrove roots incorrect

Fixed:
Mangrove roots correct test 1

### DIFF
--- a/chunky/src/java/se/llbit/chunky/model/MangroveRootsModel.java
+++ b/chunky/src/java/se/llbit/chunky/model/MangroveRootsModel.java
@@ -15,27 +15,27 @@ public class MangroveRootsModel extends QuadModel {
 
   private static final Quad[] quads = new Quad[]{
     new Quad(
-      new Vector3(0.8 / 16.0, 16 / 16.0, 8 / 16.0),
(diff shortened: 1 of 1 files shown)


leMaik:
Thank you! :clap:

PR #1256, opened 2022-05-08
Add per-region selection instead of a list of chunks

NotStirred:
In theory this is much faster for large selections, though due to the weirdness of the map view there is no performance gain
(`chunkUpdated` must be fired for each chunk with its selection changed, `regionUpdated` is only correct if the scale is showing regions tiles)

This PR will give a nice way of getting per-region chunk selection for scene loading without having to collect the chunk positions into a map before loading them.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/ChunkMap.java
+++ b/chunky/src/java/se/llbit/chunky/ui/ChunkMap.java
@@ -48,6 +48,7 @@ import se.llbit.chunky.world.Icon;
 import se.llbit.chunky.world.PlayerEntityData;
 import se.llbit.chunky.world.World;
 import se.llbit.chunky.world.listeners.ChunkUpdateListener;
+import se.llbit.chunky.world.region.MCRegion;
 import se.llbit.log.Log;
 import se.llbit.math.QuickMath;
 import se.llbit.math.Ray;
@@ -56,6 +57,7 @@ import se.llbit.math.Vector3;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Collection;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


leMaik:
> chunkUpdated must be fired for each chunk with its selection changed

How about chunkSelectionUpdated or something? What do the listeners do?

NotStirred:
I didn't create the listeners sadly, both `chunkUpdated` and `chunkSelectionUpdated` need to be fired when the selection changes due to how the map view works.

This also means that for every single chunk you select, chunky checks if the region file is updated, amongst a bunch of other stuff that's completely unnecessary for a selection change

Once I get around to the map view rewrite stuff, I'll untangle the mess but for now it's the same behaviour as before

NotStirred:
Only issue with this PR currently is that occasionally a chunk will be deselected but its region isn't removed from the region map. Doesn't affect behaviour, but would be nice to clean it up

NotStirred:
Additional performance improvements in 803dde6. See example below:

**Before**:

https://user-images.githubusercontent.com/16853282/169709938-f87a1f45-8aea-4e1e-8031-3f2ea4a36f3f.mp4

**After**:

https://user-images.githubusercontent.com/16853282/169709940-53b8fcb7-ff18-4504-9f7c-cf515f49bc72.mp4

chunky/src/java/se/llbit/chunky/ui/ChunkMap.java
```diff
@@ -261,8 +289,8 @@ protected synchronized void selectWithinRect() {
       // If ctrlModifier to deselect, then do deselect
       // If no ctrlModifier, select chunks
       // but if they are all already selected, then deselect.
-      if (ctrlModifier || !chunkSelection.selectChunks(mapLoader.getWorld(), x0, z0, x1, z1))
-        chunkSelection.deselectChunks(x0, z0, x1, z1);
+      if (ctrlModifier || !chunkSelection.setChunks(mapLoader.getWorld(), x0, z0, x1, z1, true)) //TODO: what the..?
```
leMaik, line 292: ?

chunky/src/java/se/llbit/chunky/world/ChunkSelectionTracker.java
```diff
@@ -32,10 +32,100 @@
...
+
+    if (selectedChunksForRegion.nextSetBit(0) == -1) {
+      //all bits are 0, we don't need to track this region anymore
+      selectedChunksByRegion.remove(regionPos.getLong());
+    }
+
+    if(selectionChanged) {
+      notifyRegionChunksUpdated(regionPos, changedChunks);
+//      notifyChunksUpdated(changedChunks);
+//      notifyRegionChunksUpdated(regionPos);
```
leMaik, line 90: Remove this then?

leMaik (changes requested):

chunky/src/java/se/llbit/chunky/world/ChunkSelectionTracker.java
NotStirred: oh - good point, this was some debugging stuff

PR #1257, opened 2022-05-08
Load chunks per-region in Scene#loadChunks

NotStirred:
Merge after #1256

Rather than loading chunks in the order they appear in the chunk list, now load chunks region by region, which should lead to much better region loading performance, also the task scheduling has been cleaned up.
Maybe some minor improvements cache-wise for the octree due to loading chunks close to each other too

Still need to create some `ChunkData` cache, as each chunk currently allocates one

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
@@ -23,13 +23,15 @@ import se.llbit.chunky.renderer.RenderManager;
 import se.llbit.chunky.renderer.SceneIOProvider;
 import se.llbit.chunky.renderer.SceneProvider;
 import se.llbit.chunky.world.ChunkPosition;
+import se.llbit.chunky.world.RegionPosition;
 import se.llbit.chunky.world.World;
 import se.llbit.log.Log;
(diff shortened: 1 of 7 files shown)


NotStirred:
Should now be ready for review!

NotStirred:
wait I really bricked it and didn't realise

ShirleyNekoDev (approved):
Haven't tested these changes locally, but code looks good and if it is faster it's even better :)

ShirleyNekoDev:
Required PR is merged, therefore I assume that this one is ready once the conflicts are resolved? 🔥

leMaik:
I rebased this on top of `master`, please let me know if it still works as intended, @NotStirred :pray: 

Could we somehow use this to do more work in parallel? :thinking:  Building the octree has to be single-threaded, but parsing the chunks and building up the block palette could be parallel in theory (the block palette would need to be some concurrent map though).

NotStirred:
The next obvious step is to parse into chunkdata on many threads, though `Scene#loadChunks` is so messy that we might want to refactor that first...

leMaik:
Allright. Happy to merge this PR after 3 years :see_no_evil: 
If you want to refactor `Scene#loadChunks` first, feel free to split that into an intermediate PR if you want.

PR #1253, opened 2022-05-01
Fix broken change scenes directory button

ShirleyNekoDev:
fixes #1252
- simple fix in commit 1 (FXML loader could not find the file)
- refactor in commit 2: removes the scene picker dialog and opens the directory browser directly instead

PR #1261, opened 2022-05-15
Move the scenes directory buttons to the scene chooser dialog.

leMaik:
As [promised](https://github.com/chunky-dev/chunky/pull/1253#issuecomment-1120205530)... This will help removing the options panel.

Bildschirmfoto von 2022-05-15 16 02 36


leMaik:
Related to #1173 (browsing for a scene file is still missing, though)

PR #1263, opened 2022-05-16
Feature/per region octree

NotStirred:
merge after #1256 and #1257

Mainly looking for comments on overly verbose or confusing stuff, or better approaches to what I'm doing. There is a lot missing from this PR currently (multithreading, per-region finalisation, scene abstraction)

This makes CubicChunks support quite a pita, I plan do some scene refactoring to allow for some abstraction in the chunk loading steps

This PR is relatively complex, and some of the logic is a little hard to follow. I intend to comment it more to explain some of it.

It introduces the concept of a `sub-octree` which is a (at maximum) 512^3 size octree. All of the sub-octrees for the region are loaded, and then they replace an entire node in the global octree

I hope to add finalisation to this also, and then finalise the seams once all regions are loaded.


NotStirred:
Still to do for this pr:
- Per-region octree finalisation, where the inside of each octree is finalised and the seams are finalised after merging
- Option to disable the per-region octree mode to reduce the size of the octree

PR #1265, opened 2022-05-20
Fix palette error due to expecting block name to contian "minecraft:"

NotStirred:
also add a general catch for the region parser to prevent such errors from going unnoticed / without a stacktrace

### DIFF
--- a/chunky/src/java/se/llbit/chunky/block/MinecraftBlockProvider.java
+++ b/chunky/src/java/se/llbit/chunky/block/MinecraftBlockProvider.java
@@ -997,7 +997,6 @@ public class MinecraftBlockProvider implements BlockProvider {
     // drop the minecraft: prefix
     String[] split = namespacedName.split(":", 2); // split into maximum 2 parts
 
-    //TODO: modded block support? non-minecraft prefix is ignored
     if(split.length != 2 || !split[0].equals("minecraft")) {
       return null;
     }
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/chunky/block/MinecraftBlockProvider.java
```diff
@@ -994,7 +994,14 @@ public Collection<String> getSupportedBlocks() {
...
   @Override
   public Block getBlockByTag(String namespacedName, Tag tag) {
-    String name = namespacedName.substring(10); // drop the minecraft: prefix
+    // drop the minecraft: prefix
+    String[] split = namespacedName.split(":", 2); // split into maximum 2 parts
+
+    //TODO: modded block support? non-minecraft prefix is ignored
+    if(split.length != 2 || !split[0].equals("minecraft")) {
+      return null;
+    }
```
leMaik, line 1003: Actually, this already _is_ modded block support. Chunky will try all registered `BlockProvider`s and use the first one that returns a non-null value.

```suggestion
    if(split.length != 2 || !split[0].equals("minecraft")) {
      return null;
    }
```

chunky/src/java/se/llbit/chunky/world/region/RegionParser.java
```diff
@@ -54,26 +54,30 @@ public RegionParser(WorldMapLoader loader, RegionQueue queue, MapView mapView) {
...
+            if (map.shouldPreload(chunk)) {
+              if (chunk.loadChunk(chunkData, mapView.getYMin(), mapView.getYMax())) {
+                chunkData.get().clear();
+              }
             }
           }
         }
       }
+    } catch (Throwable t) {
+      Log.warn("Region Parser Error", t);
```
leMaik, line 80: Okay, so we keep crashing the region parser? Will Chunky still work after this warning?

leMaik (approved):

PR #1270, opened 2022-05-21
Update Gradle

ShirleyNekoDev:
Upgrade Gradle 7.4 -> 8.0.2
see https://gradle.org/releases/ for changelogs


ShirleyNekoDev:
Something something packages not found 🤔

leMaik:
We could merge this, but the build needs to be fixed :man_shrugging:

NotStirred:
Relevant issue https://github.com/openjfx/javafx-gradle-plugin/issues/140

ShirleyNekoDev:
This should now be ready :tada:
Once we move on to Java 17 or newer (without breaking existing users' setups), we can also update the class file targetCompatability and sourceCompatibility, as well as the JavaFX library.

PR #1276, opened 2022-05-22
Draft: Add support for PBR maps (specular, normal, metalness, roughness, emittance)

leMaik:


jackjt8:
pbr_4

Working as intended. Looks really cool

chunky/src/java/se/llbit/chunky/resources/pbr/LabPbrSpecularMap.java
```diff
@@ -0,0 +1,143 @@
...
+  private float[] roughnessMap;
+  private boolean hasMetalness;
+
+  public LabPbrSpecularMap(BitmapImage texture) {
+    width = texture.width;
+    height = texture.height;
+
+    emissionMap = new byte[texture.width * texture.height];
+    boolean hasEmission = false;
+    for (int y = 0; y < texture.height; ++y) {
```
Peregrine05, line 27: We loop over the texture three times here. Can this be refactored so that we only loop once?

Peregrine05 (changes requested):

chunky/src/java/se/llbit/chunky/resources/pbr/OldPbrSpecularMap.java
```diff
@@ -0,0 +1,106 @@
...
+  private byte[] reflectanceMap;
+  private byte[] roughnessMap;
+
+  public OldPbrSpecularMap(BitmapImage texture) {
+    width = texture.width;
+    height = texture.height;
+    emissionMap = new byte[texture.width * texture.height];
+
+    boolean hasEmission = false;
+    for (int y = 0; y < texture.height; ++y) {
```
Peregrine05, line 25: The same applies in this file.

PR #1287, opened 2022-05-28
New resource pack selection dialog and improved resource pack loading

ShirleyNekoDev:
image


Peregrine05:
Please re-add loading resource packs from directories other than the Minecraft directory.

ShirleyNekoDev:
@Peregrine05 added `browse for resource packs dialog` ("+" button) and replaced unicode glyphs with custom icons
image

Peregrine05:
Thank you.

leMaik:
The UI looks amazing! :clap: Just from looking at it, I have two questions though:

1. What does "Clear all" mean and what's the difference to "Deselect all"?
2. Should we change the "+" button and give it an explicit label? I first looked for a "Browse…" button or something like that.

ShirleyNekoDev:
> 1. What does "Clear all" mean and what's the difference to "Deselect all"?

Select All = `Ctrl + A` (selection for packs as in "select these packs and use the arrows to move them between the columns")
Would "Remove All" or "Clear List" be a better name than "Clear All"?

> 2. Should we change the "+" button and give it an explicit label? I first looked for a "Browse…" button or something like that.

Would be an option, but I'm not sure if this change is necessary. Below the right list is enough space for it, if we want to have "Browse...".

ShirleyNekoDev:
Removed selection buttons, added context menu, moved browse button (now with label):
image
(remind me later to add context menu to scene and world selection dialog too)

ShirleyNekoDev:
Always has been :)

Issue #1247, opened 2022-04-28
Water World Mode - Incorrect Block State


MadDog443:
On Water World Mode there's been this issue that has persisted for a long time which honestly, I think has a decent effect on renders in terms of it standing out. On the Edge of chunks water has the wrong block state so if you use Water World Mode, it gives you this really annoying line into the abyss.
![image](https://user-images.githubusercontent.com/52639083/165677426-f645e6c2-dbf5-4b6e-8426-9cd9fa9b85fb.png)

MadDog443:
The only other issue I could see happening is not being able to set a **sea floor height** or **block palette w/ noise map** (kind of like procedural water) to avoid the water world looking off in the render due to it reflecting light differently. _This second comment is mainly a feature suggestion to improve the Water World Feature so feel free to ignore it._
As seen here:
![image](https://user-images.githubusercontent.com/52639083/165735329-af6dbdd6-867a-496a-ba66-71de35b50491.png)

Peregrine05:
A workaround at the moment is to disable `Hide water plane in loaded chunks`. That extends the water plane over the incorrect water model and hides the gap.

MadDog443:
> A workaround at the moment is to disable `Hide water plane in loaded chunks`. That extends the water plane over the incorrect water model and hides the gap.

Yeah, that would not work for me very well.... too many things visible below water level.

MadDog443:
A very belated cheers to this fix!

PR #1293, opened 2022-06-06
Refactor octree finalization, make water level independent of block solid property, fix water level below submerged blocks.

### LINKED ISSUE #1236: Water level is incorrect in 2.5.0 snapshots
In Minecraft, water that is next to farmland renders as shown in the picture below.

But in Chunky, water does not have the correct water level when next to farmland in the same way water in Minecraft does.
image

leMaik:
Fixes #1236 

TODO:

- [ ] Check lava
- [ ] Cleanup (maybe get rid of the water methods in `FinalizationState`?)
- [ ] Check if legacy block finalization still works
- [ ] Look into #1247 by figuring out whether we are at the edge of the selected chunks in `FinalizationState`

### DIFF
--- a/chunky/src/java/se/llbit/chunky/block/BlockFace.java
+++ b/chunky/src/java/se/llbit/chunky/block/BlockFace.java
@@ -1,40 +1,64 @@
 package se.llbit.chunky.block;
 
 public enum BlockFace {
+  SELF("self", 0, 0, 0),
   /**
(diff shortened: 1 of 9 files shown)


NotStirred:
Haven't yet looked carefully through the changes in logic, but I like the idea. Anything that moves code out of `Scene` is great

leMaik:
Not ready for merge yet, but I'll finish this in two weeks.

PR #1310, opened 2022-06-12
Hide unknown blocks

### LINKED ISSUE #1302: Option to hide unknown blocks/question marks
Currently unknown blocks will be displayed as question marks within the world.

An option to forcibly hide these blocks would be useful so that people can still render worlds.

leMaik:
Closes #1302 

This PR hides unknown blocks at intersection time. We could also hide them while creating the octree (ie. in `BlockSpec.toBlock()` but that would require access to the scene, which we don't have there yet.


chunky/src/java/se/llbit/chunky/ui/render/tabs/AdvancedTab.java
```diff
@@ -279,6 +285,7 @@ public void update(Scene scene) {
     gridSize.set(scene.getGridSize());
     preventNormalEmitterWithSampling.setSelected(scene.isPreventNormalEmitterWithSampling());
     animationTime.set(scene.getAnimationTime());
+    hideUnknownBlocks.setSelected((scene.getHideUnknownBlocks()));
```
leMaik, line 288:
```suggestion
    hideUnknownBlocks.setSelected(scene.getHideUnknownBlocks());
```

PR #1328, opened 2022-07-01
Minor code cleanups in post processing and image exports

ShirleyNekoDev:
- add missing copyrights
- reuse buffer in PFM export (IO speedup)
- code cleanup of unused parameters (export with multiple threads was never implemented)
- add some more code documentation


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1,4 +1,8 @@
+<<<<<<< HEAD
```
leMaik, line 1: old merge conflict stuff?

leMaik (approved):

PR #1329, opened 2022-07-02
Use the compacted WorldTexture for 2d biomes by default again

### LINKED ISSUE #1327: Excessive memory usage since 3d biome support was added
It seems like the commit "Add 3d biomes and 3d blur (off by default) (#1225)" makes Chunky use drastically more memory when loading a scene. Loading scenes with more than 200k chunks has been no problem for me before, but now I run out of memory when loading scenes with as little as 20k chunks.

leMaik:
This should fix #1327


jackjt8:
```4k chunks
2.4.2		983.9 MB
2.5.0-128	8527.4 MB
2.5.0-133	7705.5 MB
PR1329		7879.4 MB```

Doesn't look like this works.

PR #1330, opened 2022-07-03
Allocate less memory while loading chunks

leMaik:


jackjt8:
`2.5.0-DEV.134.g21ccf11` (ie master)
2 5 0-DEV 134 g21ccf11


`PR.1330`
PR 1330

Looks like this PR hits the GC less often during loading. It seems that we might get some more GC during rendering but performance in this case is unchanged.

image

In addition to the goal of reducing the memory allocation. Load time and Used memory is reduced by a notable amount. Hopefully someone else can confirm these findings for me.

PR #1341, opened 2022-07-06
Fix Export ZIP and Export PNG tooltips

Peregrine05:
Fixed an issue where the Export ZIP had an incorrect tooltip and the Export PNG had no tooltip.


jackjt8:
Needs to be applied to `chunky-2.4.x` in addition to `master`.

Peregrine05:
It is being applied only to 2.4.x. The tooltip does not exist in master.

PR #1344, opened 2022-07-08
Fix incorrect biomes in some worlds

### LINKED ISSUE #1343: Biome colors are incorrect in the latest snapshot, and change depending on biomes within chunk selection
Depending on the biomes within the chunk selection, the colors of all of the foliage change to that of one of the biomes in the chunk selection. The color changes depending on the biomes within the selection.

Biome_incorrect_1

Biome_incorrect_2

NotStirred:
Closes #1343 

We were previously assuming the biomeIdxStructure was always compatible with 3d coordinates, which isn't true

I'll make a PR to clean up the biome palette stuff soon (@leMaik suggested something which I've yet to implement), but wanted to get it fixed straight away


leMaik:
Nice, thank you for the quick fix! :)

PR #1347, opened 2022-07-09
Update singleColorBtn name.

Peregrine05:
The name was inaccurate, as using this option does not actually require a restart.

Added a tooltip to the control.


leMaik:
Didn't I rename all of them already? :eyes:

PR #1350, opened 2022-07-09
Remove outdated "Move Camera to Player" button in CameraTab

ShirleyNekoDev:
The underlying feature does not work as expected* and can also be accessed in EntitiesTab via "Camera to Entity".

* Code assumes that only one player exists in the scene, otherwise it uses the last loaded one - since we are able to load all players in a world, this is not what the user wants to do in most cases.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/render/tabs/CameraTab.java
+++ b/chunky/src/java/se/llbit/chunky/ui/render/tabs/CameraTab.java
@@ -70,7 +70,6 @@ public class CameraTab extends ScrollPane implements RenderControlsTab, Initiali
   @FXML private DoubleTextField yawField;
   @FXML private DoubleTextField pitchField;
   @FXML private DoubleTextField rollField;
-  @FXML private Button cameraToPlayer;
   @FXML private Button centerCamera;
   @FXML private ChoiceBox<ProjectionMode> projectionMode;
   @FXML private DoubleAdjuster fov;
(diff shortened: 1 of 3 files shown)


chunky/src/java/se/llbit/chunky/ui/render/tabs/CameraTab.java
```diff
@@ -243,13 +242,6 @@ private void updateShift() {
     rollField.setTooltip(new Tooltip("Camera roll."));
     rollField.addEventFilter(KeyEvent.KEY_PRESSED, directionHandler);
 
-    cameraToPlayer.setTooltip(new Tooltip("Move camera to the player position."));
-    cameraToPlayer.setOnAction(e -> {
-      scene.moveCameraToPlayer();
-      updateCameraPosition();
-    });
-
-    centerCamera.setTooltip(new Tooltip("Center camera above loaded chunks."));
```
leMaik, line 252: Is this tooltip supposed to be removed?

chunky/src/java/se/llbit/chunky/ui/render/tabs/CameraTab.java
```diff
@@ -243,13 +242,6 @@ private void updateShift() {
...
     centerCamera.setOnAction(e -> {
```
NotStirred, line 245:
```suggestion
    centerCamera.setTooltip(new Tooltip("Center camera above loaded chunks."));
    centerCamera.setOnAction(e -> {
```

Issue #712, opened 2020-10-31
Chunky 1.4.5 changing scene settings not applying unless enter is hit


DannyN2222:
Render controls (notably camera position / orientation) don't apply unless you press enter in each text box,
expected behaviour would be to use the value in the box regardless of weather the user pressed enter at the end or not an alternative would be to revert the value when the text box has lost focus, also noticted similar issues with the lighting sliders when you change them and try a reset and rerender the settings arnt applied, another option would be to colour the box red if the settings havent applied and green for changed and applied settings

jackjt8:
This is also true for Chunky 2.X.

@IceSelkie Could you implement box colouring for edited but not applied settings?

PR #1351, opened 2022-07-09
Add more labels to camera controls

ShirleyNekoDev:
- Have labels for each input field:
image
  Previous:
image

- Moved the "center camera" (above loaded chunks) button to the camera coordinate input.
- Also rewritten some of the tooltips and removed self-explanatory ones.

This PR is based on top of #1350

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/elements/TextFieldLabelWrapper.java
+++ b/chunky/src/java/se/llbit/chunky/ui/elements/TextFieldLabelWrapper.java
@@ -1,0 +1,110 @@
+/* Copyright (c) 2022 Chunky contributors
+ *
+ * This file is part of Chunky.
+ *
+ * Chunky is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


NotStirred:
suggestion: (look at the `Position` (old), vs `Orientation` (new)
image

ShirleyNekoDev:
done 👍 
image

NotStirred (changes requested):
Looks like this suffers from something similar to what was fixed in #1205, the field isn't updated on focus loss

ShirleyNekoDev:
> Looks like this suffers from something similar to what was fixed in #1205, the field change isn't applied on focus loss

This was not the case before, therefore I would prefer to fix that in a separate PR if needed.

NotStirred:
Ah, I thought it was introduced by this PR. Yep we should do it in a different PR I'll make an issue now if there isn't one already

NotStirred (approved):

NotStirred:
yep, it's https://github.com/chunky-dev/chunky/issues/712

PR #1358, opened 2022-07-10
Fix issue where 2d biome index structure was taking much more memory than intended.

ThatRedox:


NotStirred:
same issue is in `Position2d2ReferencePackedArrayStructure`

does it make more sense to have `packedSectionPos` take block coordinates, and pack them into a section pos? not sure

```
private long packedSectionPos(int x, int z) {
  return ((z >> 4) & 0xFFFFFFFFL) | ((x >> 4) & 0xFFFFFFFFL) << 32;
}
```

ThatRedox:
Did not notice it was there too. 👍 
I think its fine to just have `packedSectionPos` take in chunk coordinates. Its basically just doing the same thing as `ChunkPosition` anyways.

NotStirred (approved):

chunky/src/java/se/llbit/math/structures/Position2IntStructure.java
```diff
@@ -1,7 +1,17 @@
 package se.llbit.math.structures;
 
 public interface Position2IntStructure {
+  /**
+   * @param x World X.
+   * @param y World Y.
+   * @param z World z.
```
NotStirred, line 7:
```suggestion
   * @param z World Z.
```

chunky/src/java/se/llbit/math/structures/Position2IntStructure.java
```diff
@@ -1,7 +1,17 @@
...
+   */
   void set(int x, int y, int z, int data);
 
+  /**
+   * @param x World X.
+   * @param y World Y.
+   * @param z World z.
```
NotStirred, line 14:
```suggestion
   * @param z World Z.
```

leMaik (approved):

Issue #1345, opened 2022-07-08
java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because the return value of "se.llbit.chunky.world.Chunk.getVersion()" is null


Peregrine05:
```
Scene manager has crashed due to an uncaught exception. Chunky will not work properly until you restart it. If you think this is a bug, please report it to the developers.
java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because the return value of "se.llbit.chunky.world.Chunk.getVersion()" is null
	at se.llbit.chunky.renderer.scene.Scene.loadChunks(Scene.java:1325)
	at se.llbit.chunky.renderer.scene.SynchronousSceneManager.loadChunks(SynchronousSceneManager.java:265)
	at se.llbit.chunky.renderer.scene.AsynchronousSceneManager.lambda$loadChunks$9(AsynchronousSceneManager.java:213)
	at se.llbit.chunky.renderer.scene.AsynchronousSceneManager.run(AsynchronousSceneManager.java:91)

```

PR #1361, opened 2022-07-12
Only update the map selection if the user loads the world associated with a scene

### LINKED ISSUE #1359: Loading a scene but not switching to its world will cause chunks to be selected in the current one
If you load a scene without switching to the world associated with it, its selection is still applied. If the selection overlaps with empty chunks in the current world, you have an invalid selection

This is the original cause for #1345, as the invalid selection causes empty chunks to load, which have a `null` version

NotStirred:
Closes #1359 and proper fix for #1345

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
+++ b/chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
@@ -361,19 +361,23 @@ public class ChunkyFxController
         refreshSettings();
         guiUpdateLatch.countDown();
         World newWorld = scene.getWorld();
-        if (newWorld != EmptyWorld.INSTANCE
-          && mapLoader.getWorld() != EmptyWorld.INSTANCE
-          && !mapLoader.getWorld().getWorldDirectory().equals(newWorld.getWorldDirectory())) {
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
```diff
@@ -374,6 +373,7 @@ public void exportMapView() {
           Dialogs.stayOnTop(loadWorldConfirm);
           if (loadWorldConfirm.showAndWait().orElse(ButtonType.CANCEL) == ButtonType.YES) {
             mapLoader.loadWorld(newWorld.getWorldDirectory());
+            getChunkSelection().setSelection(chunky.getSceneManager().getScene().getChunks());
```
leMaik, line 376: If the world is already loaded, we shhould also update the selection. Otherwise, do we already clear the selection?

NotStirred: Good point, and no we don't clear the selection currently

leMaik (approved):

leMaik:
Does #1379 also need to be backported to 2.4.x?

NotStirred:
No, #1379 was a regression in #1256

PR #1362, opened 2022-07-12
Throw exception when failing to load a chunk, allowing it to be safely skipped

NotStirred:
The user is also warned about chunks being skipped in both map view loading, and in octree loading.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/world/Chunk.java
+++ b/chunky/src/java/se/llbit/chunky/world/Chunk.java
@@ -23,6 +23,7 @@ import se.llbit.chunky.block.Water;
 import se.llbit.chunky.block.legacy.LegacyBlocks;
 import se.llbit.chunky.chunk.BlockPalette;
 import se.llbit.chunky.chunk.ChunkData;
+import se.llbit.chunky.chunk.ChunkLoadingException;
 import se.llbit.chunky.chunk.EmptyChunkData;
 import se.llbit.chunky.chunk.biome.BiomeDataFactory;
 import se.llbit.chunky.map.*;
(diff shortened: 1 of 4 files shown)


chunky/src/java/se/llbit/chunky/world/Chunk.java
```diff
@@ -421,8 +430,9 @@ public synchronized void renderTopography() {
    * @param palette Block palette
    * @param minY The requested minimum Y to be loaded into the chunkData object. The chunk implementation does NOT have to respect it
    * @param maxY The requested maximum Y to be loaded into the chunkData object. The chunk implementation does NOT have to respect it
+   * @throws ChunkLoadingException If there is an issue loading the chunk, and it should be aborted
    */
-  public synchronized void getChunkData(@NotNull Mutable<ChunkData> reuseChunkData, BlockPalette palette, BiomePalette biomePalette, int minY, int maxY) {
+  public synchronized void getChunkData(@NotNull Mutable<ChunkData> reuseChunkData, BlockPalette palette, BiomePalette biomePalette, int minY, int maxY) throws ChunkLoadingException {
```
leMaik, line 435: Should this be an `Exception` instead of a `RuntimeException`, since we are able to declare it everywhere?

leMaik (approved):
Maybe change the exception to `extends Exception`, rest is nice :+1: 
As long as new and legacy worlds still load.

chunky/src/java/se/llbit/chunky/world/Chunk.java
NotStirred: The reason it doesn't extend `Exception` is because non-`RuntimeException`s can't be thrown from `Executor#submit(Runnable)`, but they can be from `Executor#submit(Callable<T>)` where T is the return type.

Which is just bizarre. So we either extend `RuntimeException`, or we use a `Callable<Void>` and return `null`. I just thought it was a little janky

PR #1363, opened 2022-07-13
Updated the preset names and icons

Peregrine05:
The preset names used somewhat non-standard names, e.g., *East-South* instead of *South-East*. Direction names were standardized.
The preset icons showed the camera as being in the incorrect position relative to the preset name (assuming North is upward on the icon.) Icons were rearranged to properly represent the camera direction.

**Old:**
image

**New:**
image


chunky/src/java/se/llbit/chunky/renderer/scene/CameraPreset.java
```diff
@@ -26,7 +26,7 @@
  * @author Jesper Öqvist <jesper@llbit.se>
  */
 public enum CameraPreset {
-  ISO_WEST_NORTH("Isometric West-North", Icon.isoWN.fxImage(), -Math.PI / 4, -Math.PI / 4) {
+  ISO_WEST_NORTH("Isometric North-West", Icon.isoNW.fxImage(), -Math.PI / 4, -Math.PI / 4) {
```
NotStirred, line 29:
```suggestion
  ISO_NORTH_WEST("Isometric North-West", Icon.isoNW.fxImage(), -Math.PI / 4, -Math.PI / 4) {
```

chunky/src/java/se/llbit/chunky/renderer/scene/CameraPreset.java
```diff
@@ -40,7 +40,7 @@ public enum CameraPreset {
       camera.setShift(0, 0);
     }
   },
-  ISO_EAST_SOUTH("Isometric East-South", Icon.isoES.fxImage(), -5 * Math.PI / 4, -Math.PI / 4) {
+  ISO_EAST_SOUTH("Isometric South-East", Icon.isoSE.fxImage(), -5 * Math.PI / 4, -Math.PI / 4) {
```
NotStirred, line 43:
```suggestion
  ISO_SOUTH_EAST("Isometric South-East", Icon.isoSE.fxImage(), -5 * Math.PI / 4, -Math.PI / 4) {
```

chunky/src/java/se/llbit/chunky/renderer/scene/CameraPreset.java
Peregrine05: Good idea

Peregrine05:
They are not. The presets have to do with where the camera is facing, so it makes sense if they are on the opposite side of where they are facing.

leMaik:
It doesn't imho. The icons are very confusing now, the red dot clearly represents the camera

Peregrine05:
Do I need to make a diagram for you?

leMaik:
I got it, see the edit. Maybe we could change the icons to make clear whether the dot is the camera or the viewport. For iso icons, showing the camera location makes more sense to me.

NotStirred:
image
perhaps something like this?

PR #1364, opened 2022-07-13
Actually update the procedural water defaults

Peregrine05:


NotStirred:
This does actually change the defaults, I've tested a few edge cases I thought of and all seems fine

NotStirred (approved):

PR #1371, opened 2022-07-18
Add entity position fields to allow user to position the entity

### LINKED ISSUE #1325: Add fields to set the position of entities
Today I decided to make a render where our entire clan is in the maid costume. I stumbled upon a problem that has always annoyed me: models just can't be moved in XYZ. I would like this feature to be added in the future
Screenshot_5471
.

NotStirred:
Closes #1325

I thought it fit better next to scale and head scale, as it's the position of the entity overall, not the position of any part

image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/render/tabs/EntitiesTab.java
+++ b/chunky/src/java/se/llbit/chunky/ui/render/tabs/EntitiesTab.java
@@ -658,14 +658,19 @@ public class EntitiesTab extends ScrollPane implements RenderControlsTab, Initia
   private void withPose(Entity entity, String part, Consumer<JsonArray> consumer) {
     if (entity instanceof Poseable && part != null && !part.isEmpty()) {
       Poseable poseable = (Poseable) entity;
-      JsonArray pose = poseable.getPose().get(part).array();
+      JsonObject poseObject = poseable.getPose();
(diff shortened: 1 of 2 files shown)


Peregrine05:
So far, so good, except that when the selected entity is a lectern, the location controls apply only to the lectern, while all other controls apply to the book on the lectern.
I don't know if this is related, but when a lecten does not have a book, then the lectern does not get an entry in the Entities list.

NotStirred:
I see the issue, lecterns are annoying

Peregrine05:
The entity location controls now apply to not only the lectern, but also the book on top. Lecterns without books are still not listed in the entities list, however.
image

NotStirred:
we should probably discuss what the behaviour should be
a lectern itself isn't poseable, it just has a position, and so shouldn't really be in the entity list. (like campfires, and other similar "entities")

Should we just have the book and remove the lectern?

NotStirred:
Removed lectern changes, will push another PR fixing it

NotStirred:
other PR is #1376

Peregrine05 (approved):
From testing, everything is working as expected, except for the lectern bug, which will be fixed separately.

ShirleyNekoDev (approved):

leMaik (approved):

PR #1376, opened 2022-07-20
Separate Lectern and Book to match enchantment tables and its book

NotStirred:
Previously a lectern redirected all pose information to its child book, which is different to the enchantment table. Likely because both lectern and book are entities

Main changes in this PR:
 - blocks and entities can return multiple entities on `toEntity` and `fromJson` respectively
 - lecterns are no longer poseable entities
 - lecterns now return both the lectern and its book as entities

As a result, lecterns don't show up in the entities list (as they shouldn't), and lectern books show up in the entities list

### DIFF
--- a/chunky/src/java/se/llbit/chunky/entity/Entity.java
+++ b/chunky/src/java/se/llbit/chunky/entity/Entity.java
@@ -71,7 +71,7 @@ abstract public class Entity {
    * @param json json data.
    * @return unmarshalled entity, or {@code null} if it was not a valid entity.
    */
-  public static Entity fromJson(JsonObject json) {
+  public static Collection<Entity> fromJson(JsonObject json) {
     String kind = json.get("kind").stringValue("");
     switch (kind) {
       case "painting":
--- a/chunky/src/java/se/llbit/chunky/entity/Entity.java
+++ b/chunky/src/java/se/llbit/chunky/entity/Entity.java
@@ -21,6 +21,7 @@ import se.llbit.chunky.chunk.BlockPalette;
 import se.llbit.chunky.model.minecraft.DecoratedPotModel;
 import se.llbit.json.JsonObject;
 import se.llbit.json.JsonValue;
+import se.llbit.log.Log;
 import se.llbit.math.Grid;
 import se.llbit.math.Octree;
 import se.llbit.math.Vector3;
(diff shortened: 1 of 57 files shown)


leMaik:
Lecterns are entities only because they are bigger than a block, i.e. they would be block entities. But then there's the book, which must be an entity (an actor) so that it can be posed.

Having a block entity _and_ an entity at the same time would help, but I guess returning collections helps make this future-proof?

NotStirred:
Yeah that was my thinking, if you'd like me to implement it differently let me know

leMaik:
@NotStirred I rebased this on top of `master`, maybe you could verify that this still does what you intended. :rocket: 

As a follow-up PR, we could add a button to create a book entity from a lectern, right now adding a lectern creates an _empty_ lectern and not one with a book.

leMaik (approved):

NotStirred:
Yeah that's probably better :+1:

leMaik:
Rebased, renamed methods and added more javadoc.

Having `Entity::fromJson` return multiple entities doesn't make sense now that I think about it. Since we create an entity per entity we already store them separately in the  scene file, so loading an entity from there will always only produce a single entity per entity.

NotStirred:
Rebased on master@https://github.com/chunky-dev/chunky/commit/156e2bba2526fe4cf218243704b6f3e6d0281d1a

chunky/src/java/se/llbit/chunky/entity/Entity.java
```diff
@@ -68,72 +70,52 @@ public void loadDataFromOctree(Octree octree, BlockPalette palette, Vector3i ori
...
    * Unmarshalls an entity object from JSON data.
    *
+   * <p>This method only returns a {@link Collection} to support legacy scenes.</p>
+   *
    * @param json json data.
-   * @return unmarshalled entity, or {@code null} if it was not a valid entity.
+   * @return The entities, or an empty collection if no entity was found.
    */
-  public static Collection<Entity> fromJson(JsonObject json) {
+  public static Collection<Entity> entitiesFromJson(JsonObject json) {
```
NotStirred, line 78: This method had to rename due to the return type not matching the "statically overridden" `*Entity.fromJson`

leMaik: I like this approach because it doesn't change all entities just for some legacy support. 👍

PR #1377, opened 2022-07-20
Add viewport lines to camera preset icons

NotStirred:
as requested in https://github.com/chunky-dev/chunky/pull/1363#issuecomment-1183506713


NotStirred:
image

Peregrine05 (approved):
This looks good. The only "problem" that I still see is that the "chunk" is viewed diagonally, which does not really make sense if the camera is aiming toward a diagonal direction (i.e., North-West), but I doubt that many people will care.

Peregrine05:
Well, considering that the projection is parallel, there will be no perspective in the preview, so that is approximately how it will look.

NotStirred:
Sadly we have to pick between a descriptive image, and subtly incorrect lines
or less helpful image with correct lines

Peregrine05:
I would choose the more helpful but "incorrect perspective" image. Overall, it makes more sense.

leMaik (approved):

PR #1382, opened 2022-07-20
Make Chunky open the parent folder of the selected resource pack

### LINKED ISSUE #1381: Using `Open in system file browser` tries to open the resource pack itself in the file explorer
Right-clicking on a resource pack in the and then clicking `Open in system file browser` attempts to open the resource pack itself, instead of the parent folder of the resource pack. For example, if the selected resource pack is a ZIP file, then Chunky attempts to open the ZIP file of the resource pack. If the selected resource pack is defined by the existence of a pack.mcmeta file, then Chunky attempts to open the folder containing the pack.mcmeta file.

This causes issues, especially if the system file explorer does not support the resource pack file format. For example, Windows does not support opening of JAR files, even though JAR files can be loaded as a resource pack. The system file browsers of other operating systems may be incapable of opening even ZIP archives.

The `Open in system file browser` control should always attempt to open the parent folder of the selected resource pack, to avoid any problems.

However, opening the pre-loaded Minecraft client.jar in the system file browser does open the parent folder of that client.jar.

Peregrine05:
Previously, Chunky would attempt to open the resource pack itself in the file explorer, which would fail if the file format of the resource pack were not supported by the system file explorer. This fix makes Chunky attempt to open the parent folder of the selected resource pack. Fixes #1381

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/controller/ResourcePackChooserController.java
+++ b/chunky/src/java/se/llbit/chunky/ui/controller/ResourcePackChooserController.java
@@ -215,7 +215,7 @@ public class ResourcePackChooserController implements Initializable {
               Desktop.getDesktop().open(
                 getItem().isDefaultPack()
                   ? getItem().file.getParentFile()
-                  : getItem().file
+                  : getItem().file.getParentFile()
               );
             } catch (IOException ex) {
               Log.warn("Failed to open resource pack file in system file browser.", ex);


ThatRedox:
Is it possible to open the parent folder and select it. Similar to how open in explorer works in IntelliJ?

NotStirred:
Java 9 has `Desktop#browseFileDirectory` for this

ThatRedox:
Yeah, I think the best way might be: 
1. Check that method (I think I would prefer reflection over JDK version check)
2. If the texture pack is a folder, open the folder
3. Open the containing folder

ShirleyNekoDev:
@ThatRedox can you push your implementation? I closed my PR.

NotStirred:
No, if java 9 is available it will have additional functionality

Peregrine05:
What is the expected functionality of the latest changes? If it is to select the resource pack itself when its parent directory is opened, then it is failing to do so. I am running Java 17, and downloaded the latest changes to this PR to test them.

Peregrine05:
However, it the expected functionality is to open the resourcepack itself when it is in folder form, and open the parent folder when it is in ZIP or JAR form, then it is working fine.

ShirleyNekoDev:
If you are running Windows 11, this is normal - the feature to select the file in a file browser is not supported there, even with Java 17 or 18. Other platforms may support it.

ShirleyNekoDev:
Yeah, if supported, this is supposed to happen, otherwise the fallback just opens the parent folder :)

ThatRedox (approved):

PR #1385, opened 2022-07-21
Sort materials list

### LINKED ISSUE #1375: Alphabetize the Materials list
Currently, the list is not ordered. Alphabetizing it will make materials easier to find, especially in conjunction with the `Filter` control, such as when "minecraft:glass" is searched. Then the discovered glass materials will be ordered alphabetically, and less scanning through the list will need to be done to find the desired material.

ShirleyNekoDev:
closes #1375


NotStirred (approved):

Peregrine05 (approved):
I have tested this and it appears to be working as expected. Using the `Filter` control returns results in alphabetical order as well.

PR #1400, opened 2022-07-29
Fix always rounding to 3 decimal places

### LINKED ISSUE #1268: Fog limited accuracy
The Fog Slider only allows 0.0, 0.01 < x < 1.0 (max 2dp).

This provides limited accuracy. If possible we should at least allow 0.001 as the lower bounds of the slider.

The Fog Field has a 3dp limit (ie 0.001). Numbers like 0.0005 are rounded to 0.0 (not sure if it's just UI or internal).

NotStirred:
This is a quick fix to a weird issue where we often update the text field with the internally stored value. Closes #1268

Currently on any focus loss, or enter press to an input field we set the text to a locale formatted string version of the current value in chunky.

So the current logic is:
1. User sets text, internal value is updated
2. Set text field to locale formatted version of the internal value
3. Text field has changed (by ourselves in the previous step), set internal value to re-parsed text value

kind of insane.


leMaik:
We should fix this properly at some point.

PR #1408, opened 2022-08-03
Catch `ZipError` when loading resource packs.

ThatRedox:
From [discord](https://discord.com/channels/541221265512464394/571769711402942467/1004066872918024192):
```
Exception in thread "main" java.util.zip.ZipError: zip END header not found
	[...]
Chunky exited abnormally with exit value 1
```

This PR catches the `ZipError` and rethrows it as an `IOException`. In my testing this only happens in Java 8. This appears safe to catch despite being an `Error` (see [stackoverflow post](https://stackoverflow.com/a/51715939)).


leMaik:
Wow, that was fast! 🚀

PR #1420, opened 2022-08-07
Use the new number entry for the map coordinates

NotStirred:
I also made it match the size of the buttons below because that looks nice

old:
image

new:
image


NotStirred:
^ Some extra formatting changes for the right panel, first is the default size, second is it expanded
image
image

PR #1421, opened 2022-08-07
Fix horizon offset being dropped when render starts.

### LINKED ISSUE #1295: Horizon offset does not take effect unless changed after render start
Currently, changing the horizon offset appears to change in the render preview, but changes do not apply in the actual render unless the setting is changed while the path tracing renderer is in progress. Upon render reset, the horizon offset appears to revert to its default value in the render preview, even though the slider and input field display the new value.

leMaik:
Fixes #1295


Peregrine05 (approved):
Testing this works as intended.

PR #1422, opened 2022-08-07
Fix icon size of the delete selected chunks icon for larget resource packs.

### LINKED ISSUE #1294: Scale down control icons that use textures from resource packs.
One example is the `Delete selected chunks` button. It uses the tnt_side texture from the currently-loaded resource pack as its icon, so if that texture is much larger than the default, it will cause the control item to be much larger than necessary.

leMaik:
Fixes #1294. It's the only icon that uses a texture from the resource pack. Since we're going to remove this functionality from Chunky (at move it into a plugin), I don't think it's necessary to change the icon at this point.


Peregrine05 (approved):
Testing this works as intended. With a 64x64 resolution resource pack, the icon remained at 16x16, after restarting Chunky.

PR #1423, opened 2022-08-07
Fix preview render grid not being aligned to chunks.

leMaik:
Just realized that the grid is supposed to be aligned to the chunks, like this:
image


Peregrine05 (approved):
Testing appears to work as intended.
image

PR #1424, opened 2022-08-07
Fix sky mode settings not being shown after loading a scene.

### LINKED ISSUE #1296: Sky mode settings do not indicate their saved values when a saved scene is loaded.
When a saved scene is loaded, the Sky mode settings that were saved do not change to indicate their saved values. Instead, they display the default values, even though the render preview displays the result of the saved values.

leMaik:
Fixes #1296


Peregrine05 (approved):
Testing this works as intended.

Issue #1005, opened 2021-07-02
More granular control over scene synchronization.


alexhliu:
Introduce some sort of soft refresh that resyncs the scenes without resetting the render. Add some save only scene refresh (could be the same as soft refresh) for options that require a reload to apply (#955).

Issue #955, opened 2021-05-19
Octree and BVH implementation settings don't save unless the scene is refreshed somehow.


alexhliu:
If the Octree and/or BVH implementation settings are changed, and then the scene is imminently saved, the saved scene does not affect the new implementation selections.

There must be a scene refresh of some sort (ie resetting the render) in order for the save to properly reflect the new selections.

PR #1426, opened 2022-08-07
Add ability to switch octree implementations.

### LINKED ISSUE #1416: Add option to switch octree impls in an existing scene.
I believe octree has a switch impl method, or we can serialize the octree to a temporary file and deserialize it with the alternate implementation.

ThatRedox:
When complete closes #1416

Complete:
- Switch octrees in an existing scene
- Progress bar while switching octrees

To do:
- [x] Check it saves the new octree setting
- [x] Block rendering when switching octrees (can cause NPE)

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/DefaultRenderManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/DefaultRenderManager.java
@@ -246,7 +246,7 @@ public class DefaultRenderManager extends Thread implements RenderManager {
         // Wait for a scene state change (e.g. user editing the scene)
         ResetReason reason = sceneProvider.awaitSceneStateChange();
 
-        if (reason == ResetReason.SETTINGS_CHANGED_SOFT) {
+        if (!reason.shouldRerender()) {
           // Settings sync reset. No need to re-render.
           continue;
         }
--- a/chunky/src/java/se/llbit/chunky/renderer/DefaultRenderManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/DefaultRenderManager.java
@@ -246,6 +246,11 @@ public class DefaultRenderManager extends Thread implements RenderManager {
         // Wait for a scene state change (e.g. user editing the scene)
         ResetReason reason = sceneProvider.awaitSceneStateChange();
 
+        if (reason == ResetReason.SETTINGS_CHANGED_SOFT) {
+          // Settings sync reset. No need to re-render.
+          continue;
+        }
+
         // Copy the new scene state to the bufferedScene
         synchronized (bufferedScene) {
           sceneProvider.withSceneProtected(scene -> {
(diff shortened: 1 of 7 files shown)


ThatRedox:
Octree is now set to an empty octree when switching.

jackjt8:
If using an octree provided by the Octree Plugin this does not work but no error is thrown. 

image

---

Otherwise, switching between built in Octrees works as expected.

jackjt8:
Unfortunately, it does not write the updated octree or update the .json so you are stuck with the old octree.

ThatRedox:
Added soft refresh to propagate octree (and bvh) changes without forcing a rerender. Closes #1005. Closes #955.

ThatRedox:
> If using an octree provided by the Octree Plugin this does not work but no error is thrown.
> 
> image
> 
> Otherwise, switching between built in Octrees works as expected.

The progress overflow should be fixed. I can't seem to reproduce the failing with no error thrown (perhaps it was working but the progress bar just overflowed).

In general, I have found that `NODE` appears to be completely broken for me (#1434) and `DICTIONARY` is missing blocks when loading from a file so they may not be the best to test with.

jackjt8:
> Added soft refresh to propagate octree (and bvh) changes without forcing a rerender. Closes #1005. Closes #955.

Can confirm that this is working as intended.

> The progress overflow should be fixed. I can't seem to reproduce the failing with no error thrown (perhaps it was working but the progress bar just overflowed).

At the very least the fix is working.

> ... `DICTIONARY` is missing blocks when loading from a file ([issue](https://github.com/aTom3333/chunky-octree-plugin/issues/2)) so they may not be the best to test with.

I will investigate this issue further.

jackjt8:
This PR is ready last time I tested it. It would just not work correctly with the octree plugin.

aTom3333:
Apart from the `DICTIONARY` octree issue, was there something else preventing this to work with the octree plugin?

jackjt8:
> Apart from the `DICTIONARY` octree issue, was there something else preventing this to work with the octree plugin?

Actually no. I just tested the PR further after all the fixes and technically there is no issue with the Plugin and this PR.

chunky/src/java/se/llbit/chunky/renderer/DefaultRenderManager.java
```diff
@@ -246,6 +246,11 @@ public void run() {
         // Wait for a scene state change (e.g. user editing the scene)
         ResetReason reason = sceneProvider.awaitSceneStateChange();
 
+        if (reason == ResetReason.SETTINGS_CHANGED_SOFT) {
+          // Settings sync reset. No need to re-render.
+          continue;
+        }
```
leMaik, line 252: How about introducing a new `requiresRerender` to the `ResetReason` and then

```suggestion
        if (!reason.isRerenderRequired()) {
          // Settings sync reset. No need to re-render.
          continue;
        }
```

leMaik (approved):
lgtm. We could refactor the soft-reset render reason, but it's not a blocker.

leMaik (approved):

PR #1433, opened 2022-08-13
Use the available physical memory as upper memory limit for the slider in the launcher.

leMaik:
This should help with questions like this: https://old.reddit.com/r/chunky/comments/wjezxv/anyway_to_allow_chunky_to_use_more_then_16gb_of/

### DIFF
--- a/launcher/src/se/llbit/chunky/launcher/ui/ChunkyLauncherController.java
+++ b/launcher/src/se/llbit/chunky/launcher/ui/ChunkyLauncherController.java
@@ -315,7 +315,7 @@ public final class ChunkyLauncherController implements Initializable, UpdateList
         .getOperatingSystemMXBean()).getTotalPhysicalMemorySize() / 1024 / 1024);
     } catch (Exception e) {
       // fallback for JDKs that don't have com.sun.management.OperatingSystemMXBean
-      return 1 << 14;
+      return 1 << 14; // 16 GiB
     }
   }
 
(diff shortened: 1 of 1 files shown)


leMaik:
On Discord, it was suggested to change the slider to keep a bit of memory. We could either limit it to some percentage or to X% of _free_ ram at the time of starting. Just to make sure that people don't run OOM by sliding it all the way to the right.

launcher/src/se/llbit/chunky/launcher/ui/ChunkyLauncherController.java
```diff
@@ -308,6 +309,16 @@ public void updateLauncher(Consumer<String> errorCallback, Consumer<LauncherInfo
     updateChecker.start();
   }
 
+  private int getUpperMemoryLimitMb() {
+    try {
+      return (int) (((com.sun.management.OperatingSystemMXBean) ManagementFactory
+        .getOperatingSystemMXBean()).getTotalPhysicalMemorySize() / 1024 / 1024);
+    } catch (Exception e) {
+      // fallback for JDKs that don't have com.sun.management.OperatingSystemMXBean
+      return 1 << 14;
```
ShirleyNekoDev, line 318:
```suggestion
      return 1 << 14; // 16 GiB
```

ShirleyNekoDev:
> On Discord, it was suggested to change the slider to keep a bit of memory. We could either limit it to some percentage or to X% of _free_ ram at the time of starting. Just to make sure that people don't run OOM by sliding it all the way to the right.

I'd assume that this should not be a problem for modern systems - When Java requests all system memory, the OS might assign it as virtual memory and page it out as needed, otherwise the JVM should terminate either way if allocation from system fails.
(This feature is probably mostly relevant for power users, which want to manage their system resources without restrictions.)

ShirleyNekoDev (approved):

PR #1435, opened 2022-08-14
Fix NPE when loading a world without water with NODE octree.

### LINKED ISSUE #1434: Cannot load chunks with NODE octree.
Node octree crashes with a null pointer exception. This happens in both 2.4.3 and 2.5.0-SNAPSHOT

Load chunks with node octree selected:
image

ThatRedox:
Closes #1434


leMaik:
@chunky-dev/qa Could anyone check if this works with an existing node-based octree? :pray:

jackjt8:
Small correction here-

In previous versions you could not even create a scene using `Node` if the selected chunks did not contain any water. Not sure when this started.

This PR allows new scenes to be made using `NODE`.

PR #1436, opened 2022-08-21
Fix loading and saving scenes without loaded chunks.

leMaik:


Peregrine05 (approved):
A scene with no chunks loaded can be saved and loaded as usual.

PR #1442, opened 2022-09-10
Add Loading scenes list title to inform user

NotStirred:
Also fix being able to open multiple scene chooser windows at once.

I'm not entirely sure we're closing them correctly, `close()` actually just calls `hide()` internally. We might be slowly stacking up popup windows internally, not sure

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
+++ b/chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
@@ -87,6 +87,7 @@ import se.llbit.log.Log;
 import se.llbit.math.Vector3;
 import se.llbit.util.ProgressListener;
 import se.llbit.util.TaskTracker;
+import se.llbit.util.annotation.Nullable;
 
 /**
  * Controller for the main Chunky window.
@@ -100,6 +101,8 @@ public class ChunkyFxController
   private ChunkMap map;
   private MapView mapView;
   protected ChunkSelectionTracker chunkSelection = new ChunkSelectionTracker();
+
+  @Nullable private SceneChooser sceneChooser;
 
   @FXML private Canvas mapCanvas;
   @FXML private Canvas mapOverlay;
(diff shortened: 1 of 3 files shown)


jackjt8:
Unfortunately I don't see the title. From a cold boot of my OS and opening Chunky:

After clicking on `File > Load...` Chunky stops responding. After ~10 seconds or so the `Load Chunky Scene` window appears and scenes are loaded asynchronously.

After this occurs, all other times `Load...` is pressed the `Load Chunky Scene` window opens within a second until the system is rebooted*. (* - It's possible that this caching stops working but I've not found anything to trigger it).

Ideally we would need to figure out what causes the UI to stall after pressing `Load...` and before the `Load Chunky Scene` window opens.

ThatRedox:
I'm not able to see the title either, although I have fewer scenes and it only takes ~2 seconds for the window to appear.

It appears opening multiple scene chooser windows *is* fixed.

--- 
> I'm not entirely sure we're closing them correctly, `close()` actually just calls `hide()` internally. We might be slowly stacking up popup windows internally, not sure

That is correct. See the [`Stage` javadocs](https://docs.oracle.com/javase/8/javafx/api/javafx/stage/Stage.html#close--).

---

> After this occurs, all other times `Load...` is pressed the `Load Chunky Scene` window opens within a second until the system is rebooted*. (* - It's possible that this caching stops working but I've not found anything to trigger it).

The scene list is cached. See [`FileTimeCache` class](https://github.com/chunky-dev/chunky/blob/c3d801aa365718fe063439e6dd7d0b8bd6bcaee6/chunky/src/java/se/llbit/chunky/ui/controller/SceneChooserController.java#L362) and the [`sceneListCache` HashMap](https://github.com/chunky-dev/chunky/blob/c3d801aa365718fe063439e6dd7d0b8bd6bcaee6/chunky/src/java/se/llbit/chunky/ui/controller/SceneChooserController.java#L68).

PR #1448, opened 2022-10-02
Add scene refresh event plugin API.

ThatRedox:
For plugins like the denoiser plugin, it would be useful to keep track of scene refresh events (to invalidate auxiliary buffers for example). Currently this is only possible (without hacks) if the plugin registers a renderer and that renderer is being used.


leMaik (approved):
Seems like a great idea for the denoiser and might also come in handy for other plugins.

PR #1449, opened 2022-10-02
Remove the chunk position cache.

ThatRedox:
Previously `ChunkPosition` objects were compared by identity and were cached to make this possible. This removes that cache and implements equality and hashing logic to `ChunkPosition` so they can be created with `new`. This allows the JVM to do more optimizations (such as allocating on the stack).

Tests run on 1 region of Greenfield. On first run, the loading times are comparable: 4.190s (this) vs 4.206s (master). On subsequent runs the loading times is up to 12% faster: 2.098s (this) vs. 2.406s (master).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,6 +38,7 @@ subprojects {
         options.release = 8
         options.encoding = 'UTF-8'
         options.debug = true
+        options.deprecation = true
     }
 
     javafx {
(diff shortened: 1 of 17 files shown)


leMaik (approved):
Trivia: This was _way_ slower before we refactored the biome color lookup to not use `ChunkPosition`.

Nice that the less complex code (without the cache) is still faster.

PR #1450, opened 2022-10-02
Resize `Settings Export` Box. Add `Select All` button.

### LINKED ISSUE #1439: Automatically resize the `Settings Export` dialog box.
In newer versions of Chunky especially, the number of options available in Chunky has become great enough that the size of the window may extend above and below the top and bottom of the screen, respectively. Limiting the default height of the dialog box and adding a scroll feature to display all options is a better solution than an extremely long window.

image

ThatRedox:
Closes #1439 . New settings export box:
image


leMaik:
I like this already. :+1: Two more suggestions:

- Add a button to save the JSON to a file (and load a file in the import dialog)
- Add a list of display names for the attributes (maybe using a hashmap), so eg. _yMin_ becomes _Minimum Y level_
- Misc and Advances are confusing (to me, at least) :thinking: Can we add a label that explains what they include?

PR #1451, opened 2022-10-02
Add context menu transformers for map and render view.

ThatRedox:
These transformers can be used to add more context menu items to the map and render displays. (ie. buttons for map editor plugin or show albedo/normal/beauty/denoised for denoiser).

### DIFF
--- a/chunky/src/java/se/llbit/chunky/plugin/ContextMenuTransformer.java
+++ b/chunky/src/java/se/llbit/chunky/plugin/ContextMenuTransformer.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2022 Chunky contributors
+ *
+ * This file is part of Chunky.
+ *
+ * Chunky is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


chunky/src/java/se/llbit/chunky/plugin/ContextMenuTransformer.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package se.llbit.chunky.plugin;
+
+import javafx.scene.control.ContextMenu;
+
+import java.util.function.Consumer;
+
+@FunctionalInterface
+public interface ContextMenuTransformer extends Consumer<ContextMenu> {
```
leMaik, line 26: These could be even more powerful if they were _actual_ transformers, ie. `Function<ContextMenu, ContextMenu>` because then they could even replace the menu entirely. :thinking: Not sure how much power we want to give to plugins though.

leMaik (approved):

PR #1455, opened 2022-10-05
Hide the Water world mode settings collapsible panel if Water world mode is disabled

raphaelts3:
Closes https://github.com/chunky-dev/chunky/issues/1454

Hey @Peregrine05, this is my first PR here, let me know if something is not right and can be improved, but it feels pretty straightforward.

PR #1456, opened 2022-10-05
Automatically hide and collapse `Water` tab panels

### LINKED ISSUE #1454: Hide the `Water world mode settings` collapsible panel if `Water world mode` is disabled
The `Procedural water settings` collapsible panel already makes use of this feature, and applying the same to this panel will ~~reduce the size of the `Water` tab.~~

Peregrine05:
Fixes #1454.

- Hide and collapse `Water world mode settings` panel if `Water world mode` is disabled.

- Hide and collapse `Procedural water settings` panel if `Procedural water` is disabled.

- Set the animated property of aforementioned panels and the `Sky mode settings` panel to false.

**Collapsed and hidden when disabled:**
image

**Expanded and shown when enabled:**
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/render/tabs/WaterTab.java
+++ b/chunky/src/java/se/llbit/chunky/ui/render/tabs/WaterTab.java
@@ -171,6 +171,7 @@ public class WaterTab extends ScrollPane implements RenderControlsTab, Initializ
 
     waterWorldModeDetailsPane.visibleProperty().set(waterPlaneEnabled.isSelected());
     waterWorldModeDetailsPane.expandedProperty().set(waterPlaneEnabled.isSelected());
+    waterWorldModeDetailsPane.managedProperty().set(waterPlaneEnabled.isSelected());
 
     waterPlaneEnabled.setTooltip(
       new Tooltip("If enabled, an infinite ocean fills the scene. This ignores air from loaded chunks."));
@@ -178,6 +179,7 @@ public class WaterTab extends ScrollPane implements RenderControlsTab, Initializ
       scene.setWaterPlaneEnabled(newValue);
       waterWorldModeDetailsPane.setVisible(newValue);
       waterWorldModeDetailsPane.setExpanded(newValue);
+      waterWorldModeDetailsPane.setManaged(newValue);
     });
 
     waterPlaneHeight.setName("Water height");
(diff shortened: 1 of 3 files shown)


leMaik:
Can we make it so that the panels don't take up vertical space if they are hidden?

Peregrine05:
image

ThatRedox (approved):
Looks good! Any particular reason why animation is disabled?

Peregrine05:
It unnecessarily adds time to opening / closing the panels. Some panels had them enabled (e.g., `Sky mode settings`), while others already had them disabled (e.g., `Load entities`), so I simply set all of them to disabled for simplicity. They can be reenabled if everyone thinks it looks better.

chunky/src/java/se/llbit/chunky/ui/render/tabs/WaterTab.java
```diff
@@ -168,11 +169,18 @@ public void initialize(URL location, ResourceBundle resources) {
       }
     });
 
+    waterWorldModeDetailsPane.visibleProperty().set(waterPlaneEnabled.isSelected());
+    waterWorldModeDetailsPane.expandedProperty().set(waterPlaneEnabled.isSelected());
+    waterWorldModeDetailsPane.managedProperty().set(waterPlaneEnabled.isSelected());
+
```
leMaik, line 175: I really liked the binding more: https://github.com/chunky-dev/chunky/pull/1455/files#diff-4441e3bab76e7447488f3d3b6642671320550af35f87af8976250f0d533f75dbR177

Could we use that instead of having to call `.set` everywhere?

leMaik (changes requested):

chunky/src/java/se/llbit/chunky/ui/render/tabs/WaterTab.java
Peregrine05: The issues with using `.bind` were that attempting to collapse a panel that was expanded would fail and would cause an error in the debug console, and that using `.bind().not()` to manage the 'collapsibleProperty' would cause one of a few possible issues, one of them being that the panel contents would fail to render.

nib9888:
Any thoughts on potentially collapsing and hiding the sun controls and cloud controls if `Draw sun` and `Enable clouds` are disabled respectively? I don't think we should go overboard on hiding options, but particularly `Cloud X`, `Cloud Y`, and `Cloud Z` feel like useless visual clutter if clouds are disabled.

Also I'm new here, hi! :)

leMaik:
@nib9888 I'll merge this. If you want to, feel free to open an issue or create a PR for the sun and cloud changes. :muscle:

leMaik (approved):

PR #1459, opened 2022-10-06
Increase the range of cloud offset sliders and change default cloud size

### LINKED ISSUE #1388: Increase the upper limit of the `Cloud Y` slider
The current upper limit of the `Cloud Y` slider (not the input field) is *100*. It should be increased to 320 or above, considering that the default cloud level in Minecraft was once at 128, and is now at 192, and that users may wish to use higher cloud levels.

Peregrine05:
Fixes #1388.
- Changed default cloud size to 12, to match Minecraft Fancy clouds.
- Changed cloud X offset range to (-256, 256).
- Changed cloud Y offset range to (-64, 320).
- Changed cloud Z offset range to (-256, 256).
- Added a tooltip to `Cloud size` control.

**New:**
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
@@ -65,7 +65,7 @@ public class Sky implements JsonSerializable {
    */
   protected static final int DEFAULT_CLOUD_HEIGHT = 128;
 
-  protected static final int DEFAULT_CLOUD_SIZE = 64;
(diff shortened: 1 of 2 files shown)


ThatRedox (approved):

NotStirred (approved):
Perhaps the cloudY range could match the world height similar to how how the min and max clip are done?

Though that could be a separate PR

leMaik (approved):

leMaik:
lgtm, as @NotStirred said, we can do the range clamping in another PR

PR #1461, opened 2022-10-06
Add `Select chunks in radius` button, with configuration menu

NotStirred:
image

Adds the ability to select a circle of chunks, within a radius. It's added to the right click menu on the map view
image

Additionally there's a config dialog available in the right click menu.
image
I can't figure out how to get the button to be square though, so it's a little chunkier than the rest of the icons. If any javafx wizard wants to fix that it'd be much appreciated

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/dialogs/SelectChunksInRadiusDialog.java
+++ b/chunky/src/java/se/llbit/chunky/ui/dialogs/SelectChunksInRadiusDialog.java
@@ -16,7 +16,7 @@ public class SelectChunksInRadiusDialog extends Dialog<ButtonType> {
   protected IntegerTextField selectionZ;
 
   public SelectChunksInRadiusDialog() {
-    this.setTitle("Adjust chunk selection radius");
+    this.setTitle("Select chunks in radius");
 
     this.selectionX = new IntegerTextField();
     this.selectionZ = new IntegerTextField();
--- a/chunky/src/java/se/llbit/chunky/ui/dialogs/SelectChunksInRadiusDialog.java
+++ b/chunky/src/java/se/llbit/chunky/ui/dialogs/SelectChunksInRadiusDialog.java
@@ -1,0 +1,63 @@
+package se.llbit.chunky.ui.dialogs;
+
+import javafx.geometry.Pos;
+import javafx.scene.control.ButtonType;
+import javafx.scene.control.Dialog;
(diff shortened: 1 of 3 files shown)


leMaik:
I like @Peregrine05's suggestion. @NotStirred What do you think, do you want to change it?

jackjt8:
Just thinking - Given that right click may not be the most accurate, especially when fully zoomed out, could the dialog also have X,Z coordinates so that the user can double check and adjust the centre point?

leMaik:
@NotStirred Would you like to include the two suggestions above into this PR or should we leave that to a future PR? :pray:

NotStirred:
Sure, will implement that now

NotStirred:
New dialog with suggestions:
image

JustinTimeCuber:
Maybe this should have an option to only select chunks within a certain angle? I could imagine loading a bunch of chunks in non-visible directions would use a lot of memory unnecessarily.

NotStirred:
That's what `Select camera visible chunks` is for, I think that does its job well enough

JustinTimeCuber:
That doesn't let you select chunks at a certain circular distance, just a rough "rectangular" distance. I'd find it useful to be able to make a circular selection at a specific distance (aka what this PR does) but without loading non-visible chunks for memory reasons.

leMaik:
@JustinTimeCuber Feel free to create a new issue or submit a PR for this idea. :+1:

chunky/src/java/se/llbit/chunky/ui/dialogs/SelectChunksInRadiusDialog.java
```diff
@@ -0,0 +1,63 @@
...
+import se.llbit.chunky.ui.IntegerTextField;
+
+public class SelectChunksInRadiusDialog extends Dialog<ButtonType> {
+  protected float chunkSelectionRadius = 10;
+
+  protected IntegerTextField selectionX;
+  protected IntegerTextField selectionZ;
+
+  public SelectChunksInRadiusDialog() {
+    this.setTitle("Adjust chunk selection radius");
```
leMaik, line 19:
```suggestion
    this.setTitle("Select chunks in radius");
```

leMaik (approved):

JustinTimeCuber:
> @JustinTimeCuber Feel free to create a new issue or submit a PR for this idea. +1

Will look into it probably this weekend

PR #1462, opened 2022-10-06
Add an `Apparent sky brightness` control

### LINKED ISSUE #514: Too bright sky through transparent leaves
When the sky light setting is increased above 1.0, transparent leaves get very bright spots. This only happens when the texture is not fully transparent - only if it is partially transparent.

The following render demonstrates the effect.

This was rendered using a custom texture pack where the space between leaves is not completely transparent. Here is the exact same scene rendered with the same settings but using Minecraft 1.13.2 textures:

This bug may be affecting all objects with transparent textures.

Peregrine05:
Closes #514.

- Fixed the `Sky light` control changing apparent sky brightness in render preview, but not in the path-traced render.

- Renamed `Sky light` to `Sky brightness`.

- Added an `Apparent sky brightness` control which acts independently from the `Sky brightness` control.

**New:**
image

Please tell me if I did anything wrong, and feel free to commit fixes, if any are needed.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
@@ -54,7 +54,7 @@ public class PreviewRayTracer implements RayTracer {
     }
 
     if (ray.getCurrentMaterial() == Air.INSTANCE) {
-      scene.sky.getSkyApparentColor(ray, true);
+      scene.sky.getApparentSkyColor(ray, true);
     } else {
       scene.sun.flatShading(ray);
     }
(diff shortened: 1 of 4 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
```diff
@@ -54,7 +54,7 @@ public class PreviewRayTracer implements RayTracer {
     }
 
     if (ray.getCurrentMaterial() == Air.INSTANCE) {
-      scene.sky.getSkyColor(ray, true);
+      scene.sky.getSkyApparentColor(ray, true);
```
leMaik, line 57:
```suggestion
      scene.sky.getApparentSkyColor(ray, true);
```

leMaik (approved):
I like the change, I'd just suggest changing the variables to `apparentSkyBrightness` instead of `skyApparentBrightness` etc?

PR #1463, opened 2022-10-06
Remove effects of `Sun intensity` and `Sun color` on preview

### LINKED ISSUE #993: Sun intensity affects preview render
Left is sun intensity = 1, right is sun intensity = 5.

I think that was intentional, to show how strong the sun will look directly in the preview, but values higher than 3 just make it difficult to see what the camera is pointed to and doesn't reflect what the end result will look like.

Peregrine05:
Fixes #993.

The preview now always appears as though the `Sun color` were white and the `Sun intensity` were 1.25.

**New:**
image

Please tell me if I did anything wrong, and feel free to commit fixes, if any are needed.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Sun.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Sun.java
@@ -137,6 +137,10 @@ public class Sun implements JsonSerializable {
 
   protected final Vector3 emittance = new Vector3(1, 1, 1);
 
+  protected final Vector3 previewEmittance = new Vector3(1, 1, 1);
+
+  protected final Vector3 previewColor = new Vector3(1, 1, 1);
+
   // final to ensure that we don't do a lot of redundant re-allocation
   private final Vector3 color = new Vector3(1, 1, 1);
 
@@ -200,6 +204,9 @@ public class Sun implements JsonSerializable {
 
     emittance.set(color);
     emittance.scale(FastMath.pow(intensity, Scene.DEFAULT_GAMMA));
+
+    previewEmittance.set(previewColor);
+    previewEmittance.scale(FastMath.pow(DEFAULT_INTENSITY, Scene.DEFAULT_GAMMA));
 
     Sky sky = ((Scene) scene).sky();
     if (sky.getSkyMode() == Sky.SkyMode.SIMULATED) {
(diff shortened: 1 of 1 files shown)


NotStirred (approved):

leMaik:
Not sure about the sun color, but not changing the intensity is a good idea :+1:

chunky/src/java/se/llbit/chunky/renderer/scene/Sun.java
```diff
@@ -201,6 +205,9 @@ private void initSun() {
     emittance.set(color);
     emittance.scale(FastMath.pow(intensity, Scene.DEFAULT_GAMMA));
 
+    previewEmittance.set(previewColor);
+    previewEmittance.scale(FastMath.pow(DEFAULT_INTENSITY, Scene.DEFAULT_GAMMA));
```
leMaik, line 209: Given that these never change, why not initialize them in the constructor?

Peregrine05:
Depending on the sun intensity and many other factors, such as sky light, the effect that the sun color has on the preview can be quite inaccurate compared to the actual render, and can simply make it difficult to see the preview properly.

Peregrine05:
I don't really know what I'm doing (I'm not a Java expert), but it works, supposedly. I don't know if this is what leMaik meant, though, or if this is how it is best done.

chunky/src/java/se/llbit/chunky/renderer/scene/Sun.java
```diff
@@ -137,6 +137,10 @@ public class Sun implements JsonSerializable {
 
   protected final Vector3 emittance = new Vector3(1, 1, 1);
 
+  private double pE = FastMath.pow(DEFAULT_INTENSITY, Scene.DEFAULT_GAMMA);
+
+  protected final Vector3 previewEmittance = new Vector3(pE, pE, pE);
```
NotStirred, line 142:
```suggestion
  private static final double pE = FastMath.pow(DEFAULT_INTENSITY, Scene.DEFAULT_GAMMA);

  protected static final Vector3 previewEmittance = new Vector3(pE, pE, pE);
```

NotStirred: does this even make sense? it may actually make more sense for these to be configurable for each sun instance :thinking:

certainly can add final to `pE` though

Peregrine05: I don't know if it makes sense, but I'll go with your suggestion, since you know more about it than I do.

ThatRedox (approved):

PR #1465, opened 2022-10-09
Add most controls from `Map` and `Render preview` right-click menus to the menu bar

### LINKED ISSUE #1387: Add more controls to the menu bar
Chunky should make more use of the menu bar.

Chunky should have `Undo` and `Redo` controls located in an `Edit` dropdown menu, and they should be able to undo or redo any undoable or redoable action that the user performs. The current `Undo` and `Redo` controls apply only to input fields, and they become inaccessible once an input field loses focus.

Controls that exist in the context menus of the `Map` tab and `Render preview` tab should also exist in the menu bar under a `Map` dropdown menu and a `Render preview` dropdown menu, respectively. This is to provide an alternative location for the controls, so that new users who do not know that the map and render preview can be right-clicked can still find and access the controls. Additionally, at large canvas resolutions, the render preview becomes invisible, rendering controls located in its context menu inaccessible.

Peregrine05:
Closes #1387.

- Added to the menu bar a `Map` menu that contains most controls from the `Map` tab right-click menu.
- Added to the menu bar a `Render preview` menu that contains most controls from the `Render preview` tab right-click menu.

Controls whose effects depend on the location of the cursor at the time of the context menu being opened, such as `Set target` and `Move camera here`, are not added to the menu bar.

**Map menu:**
image

**Render preview menu:**
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/ChunkMap.java
+++ b/chunky/src/java/se/llbit/chunky/ui/ChunkMap.java
@@ -172,11 +172,7 @@ public class ChunkMap implements ChunkUpdateListener, ChunkViewListener, CameraV
 
     selectVisible.setGraphic(new ImageView(Icon.eye.fxImage()));
     selectVisible.setOnAction(event -> {
-      ChunkView chunkView = new ChunkView(view);  // Make thread-local copy.
-      if (controller.getChunky().sceneInitialized()) {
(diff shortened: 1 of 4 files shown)


leMaik (changes requested):
I like the results, but can we implement this without duplicating the menu item code?

Issue #1175, opened 2022-02-23
Y min/max clip (re)loading inconsistencies


jackjt8:
Affected versions - `2.5.0-51-gad88ccd7`,  `2.4.2-21-g1304a5fe`, `2.4.2`, `2.5.0-SNAPSHOT.75.g8bd287901`
Unaffected versions - `2.5.0-15-g1ba5d5b`
Exported region from a World Painter world - [Kingdom of Galekin_bug.zip](https://github.com/chunky-dev/chunky/files/8128264/Kingdom.of.Galekin_bug.zip)

Steps to reproduce:

1. Select a 1.18 world with terrain that is above Y 256.
2. Set `Y min clip` to -64 and `Y max clip` to 320.
3. Look for anomalies in `Map View`  and select them; such as:
![image](https://user-images.githubusercontent.com/3740917/155416112-004cbc28-dc1d-4527-8ec3-871e7be51679.png)
4. `Load selected chunks` and observe behaviour: 
![image](https://user-images.githubusercontent.com/3740917/155416326-8ed14e8e-d836-45d7-a6d6-9dfa79821755.png)
5. 
- `Load selected chunks` mixes Chunk loading at Y clip 0<Y<256 and -64<Y<320 (see above).
- Using `Reload chunks` slices the terrain back down to 0<Y<256: 
![image](https://user-images.githubusercontent.com/3740917/155416919-8d031364-f606-48a6-88ad-3ec483eed1e3.png)

---

Alternative steps:

1. Select a 1.18 world with terrain that is above Y 256.
2. Do **NOT** alter `Y min/max clip` (ie leave at 0,256).
3. Select and  `Load selected chunks`.
4. Alter, set `Y min/max clip` to -64,320.
5. `Reload chunks`.
6. Also cry here because you had it working doing it this way before.

---

And when it magically works... (Not actually sure the exact steps to do this and it breaks very easily...) 
![image](https://user-images.githubusercontent.com/3740917/155420684-470d95b8-62f6-4a59-8a7c-18f0d5e7c8a7.png)

---

Additional information - This issue is selection dependant. Below you will see one region vs two chunks without moving the camera. (2.5.0)
![image](https://user-images.githubusercontent.com/3740917/155512017-316118cc-725f-4306-8ab3-0b7ed00dcfb5.png)
![image](https://user-images.githubusercontent.com/3740917/155512028-56d7cd79-d103-4f83-9f02-24e851b46836.png)
(2.4.2)
[...]

PR #1466, opened 2022-10-10
Fix ChunkDataVersion issues with third party applications

NotStirred:
Some WorldPainter worlds have their ChunkDataVersion set to a version prior to minecraft's support for worlds with >=256 height.
This PR changes how we create `ChunkData` objects to be based on the actual block range found in a chunk, rather than its version

### DIFF
--- a/chunky/src/java/se/llbit/chunky/world/Chunk.java
+++ b/chunky/src/java/se/llbit/chunky/world/Chunk.java
@@ -17,6 +17,7 @@
 package se.llbit.chunky.world;
 
 import se.llbit.chunky.block.minecraft.Air;
+import it.unimi.dsi.fastutil.ints.IntIntImmutablePair;
 import se.llbit.chunky.block.Block;
 import se.llbit.chunky.block.minecraft.Lava;
 import se.llbit.chunky.block.minecraft.Water;
@@ -179,7 +180,8 @@ public class Chunk {
 
     surfaceTimestamp = dataTimestamp;
     version = chunkVersion(data);
-    chunkData.set(this.dimension.createChunkData(chunkData.get(), data.get(DATAVERSION).intValue()));
+    IntIntImmutablePair chunkBounds = inclusiveChunkBounds(data);
+    chunkData.set(this.dimension.createChunkData(chunkData.get(), chunkBounds.leftInt(), chunkBounds.rightInt()));
     loadSurface(data, chunkData.get(), yMin, yMax);
     biomesTimestamp = dataTimestamp;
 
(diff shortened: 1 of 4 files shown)


chunky/src/java/se/llbit/chunky/world/ImposterCubicChunk.java
```diff
@@ -70,7 +70,7 @@ public synchronized boolean loadChunk(@NotNull Mutable<ChunkData> mutableChunkDa
     }
 
     surfaceTimestamp = dataTimestamp;
-    mutableChunkData.set(this.world.createChunkData(mutableChunkData.get(), 0)); //chunk version ignored for cubic worlds
+    mutableChunkData.set(this.world.createChunkData(mutableChunkData.get(), Integer.MIN_VALUE, Integer.MAX_VALUE)); //chunk version ignored for cubic worlds
```
leMaik, line 73: Is the comment still relevant?

leMaik (approved):

chunky/src/java/se/llbit/chunky/world/ImposterCubicChunk.java
NotStirred: good point

chunky/src/java/se/llbit/chunky/world/Chunk.java
```diff
@@ -168,7 +169,8 @@ public synchronized boolean loadChunk(@NotNull Mutable<ChunkData> chunkData, int
 
     surfaceTimestamp = dataTimestamp;
     version = chunkVersion(data);
-    chunkData.set(this.world.createChunkData(chunkData.get(), data.get(DATAVERSION).intValue()));
+    IntIntImmutablePair chunkBounds = inclusiveChunkBounds(data);
+    chunkData.set(this.world.createChunkData(chunkData.get(), chunkBounds.leftInt(), chunkBounds.rightInt()));
```
ThatRedox, line 173: Could we use `yMin` and `yMax` instead of parsing the chunk bounds on every chunk? Although we would probably need to test to see which one is faster.

jackjt8:
This PR is somewhat difficult to test. We still have #1175 which may or may not be related which could be causing issues. There are certain moments when I can sometimes load 1 chunk or 20k chunks to the correct 0 to 320 height. Other times I can do 0 to 256.

Retyping the `Scene Y clip` does not fix this inconsistent behaviour. I have tried both loading/reloading with it set to 256 and 320.

Retyping the `Map Y clip` does not fix things either.

The `Reload chunks` button can sometimes give us the correct height... and sometimes not.. even without changing the selection.

---

Loading 10k chunks
image

Map view (max Map Y clip 320) gives an indication as to where terrain should be 
image

Map view with max Map Y clip at 256 shows that terrain in the preview is still cut at 256.
image

---

Loading just 1024 chunks
image

leMaik:
The screenshots in the previous comment worried me so I didn't merge this PR yet. Are these issues _caused_ by this PR or could this be merged?

If this PR is fine, could you please rebase it, @NotStirred? :pray:

PR #1468, opened 2022-10-11
Improve skymap and other error logging.

### LINKED ISSUE #1467: Failed to load skymap - Not enough information
image

No information is provided as to why the loading of a Skymap has failed nor is there anything in the verbose debug console. If possible we should provide the reason to the user.

I suspect in this instance the [40k x 20k skymap](https://cdn.sci.esa.int/documents/33580/35361/1567215018845-ESA_Gaia_DR2_AllSky_Brightness_Colour_Cartesian_40kx20k.png/2f7d729a-12d2-7835-6083-24f739a356a7?version=1.0&t=1567215134202) is likely too large to be loaded but confirmation in the UI would be useful. [Website of image](https://sci.esa.int/web/gaia/-/60196-gaia-s-sky-in-colour-equirectangular-projection)

ThatRedox:
* Add special error message for when loading a skymap with too many pixels
* Improve error logging for failing to load skymap
* Improve some other error logging

Closes #1467

### DIFF
--- a/chunky/src/java/se/llbit/chunky/resources/HDRTexture.java
+++ b/chunky/src/java/se/llbit/chunky/resources/HDRTexture.java
@@ -102,7 +102,7 @@ public class HDRTexture extends AbstractHdriTexture {
         }
       }
     } catch (IOException e) {
-      Log.error("Error loading HRD image: " + e.getMessage());
+      Log.error("Error loading HRD image:", e);
       e.printStackTrace();
     }
   }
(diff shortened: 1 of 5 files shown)


chunky/src/java/se/llbit/chunky/resources/HDRTexture.java
```diff
@@ -102,7 +102,7 @@ public HDRTexture(File file) {
         }
       }
     } catch (IOException e) {
-      Log.error("Error loading HRD image: " + e.getMessage());
+      Log.error("Error loading HRD image:", e);
```
jackjt8, line 105: Should this not be `HDR`? I assume this was a typo as the file name is the correct order.

ThatRedox: Fixed.

leMaik (approved):

PR #1469, opened 2022-10-12
Add a control to lock the camera angle.

### LINKED ISSUE #1396: Lock camera angle
Essentially having the ability to lock your camera angle so that you don't accidentally destroy a camera angle that took hours to setup to perfection. (And contrastly, also being able to unlock your camera angle would be great too :p)

Peregrine05:
Fixes #1396.

- Added a `Lock selected camera` control, which is associated with the separate cameras.
- The control disables many other controls in the `Camera` tab, and disables camera input from the `Render preview`.

**Disabled:**
image

**Enabled:**
image

Testing is appreciated.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
+++ b/chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
@@ -81,6 +81,8 @@ public class RenderCanvasFx extends ScrollPane implements Repaintable, SceneStat
   private Vector2 target = new Vector2(0, 0);
   private Tooltip tooltip = new Tooltip();
 
+  private CheckMenuItem lockCameraAngle = new CheckMenuItem("Lock camera angle");
+
   private boolean fitToScreen = PersistentSettings.getCanvasFitToScreen();
 
   private RenderStatusListener renderListener;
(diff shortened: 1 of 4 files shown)


jackjt8:
This PR works as intended.

Though I would also like it to be a more general `Lock camera` button as currently you can also change FoV via the scroll wheel. Additional functionality would be to also ignore keyboard commands... but that's probably too far out of scope for the issue and this PR.

ThatRedox:
I agree with Jack, I think more camera locking controls would be useful. Perhaps put them in the `Camera` tab and associate them with different cameras?

chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
```diff
@@ -152,8 +154,8 @@ public RenderCanvasFx(ChunkyFxController chunkyFxController,
     });
 
     canvas.setOnMouseDragged(e -> {
-      if (e.isSecondaryButtonDown()) {
-        // do not drag when right-clicking
+      if (e.isSecondaryButtonDown() | lockCameraAngle.isSelected()) {
```
ThatRedox, line 157:
```suggestion
      if (e.isSecondaryButtonDown() || lockCameraAngle.isSelected()) {
```

ThatRedox, line 157:
```suggestion
      if (e.isSecondaryButtonDown() || lockCameraAngle.isSelected()) {
```

ThatRedox (changes requested):

ThatRedox (approved):
Looks good! I would see if we could get the lock UI from the canvas size lock but it's not super important.

PR #1489, opened 2022-10-21
Refactoring of camera projectors

ShirleyNekoDev:
- moved camera/projector related stuff into own package
- moved projector creation code out of Camera class (-> ProjectorFactory)
- add Camera interface for future extendability - implementation is now MutableCamera
- ~~ProjectorMode~~ ProjectorPreset is now a Registerable in the ProjectorPreset.REGISTRY


chunky/src/java/se/llbit/chunky/renderer/scene/camera/Camera.java
```diff
@@ -0,0 +1,66 @@
...
+  double getShiftX();
+  double getShiftY();
+
+  double getFov();
+
+  double getDof();
+  boolean isInfiniteDoF();
+  double getSubjectDistance();
+
+  ProjectionPreset getProjectionPreset();
```
leMaik, line 61: I do like this MR. The only thing I don't understand is this rename. Why call it a "projection preset"?

ShirleyNekoDev: My thoughts were that current projectors can be nested (e.g. parallel is a parallel projector inside a forward displacement projector inside a DoF inside a shift). It should be possible (in future plugins or PRs) to create such compositions from JSON definitions, which can be saved in the scene. At this point the name projection mode would be weird IMO, because you combine different existing projection modes to create a new projector factory. Therefore the currently provided projector factories are presets. (Does this make sense?)

ShirleyNekoDev: Also, CameraPresets will get refactoring treatment from me soon (and are named preset) :)

leMaik: `Projector` instead of `ProjectorPreset`? :thinking:

PR #1497, opened 2022-10-23
Add a moving average for SPS

### LINKED ISSUE #697: Use a moving average for samples per second
Render time and ETA does not count the time it takes to do a render dump/snapshot time.
If there are many render dumps/snapshots that contribute to a lot of the time, they are not accounted for in ETA or render time.

ETA also does not take into account CPU %. If CPU is run at 1% for an hour, then changed to 100%, its ETA will remain huge and slowly crawl down towards a reasonable estimate, while always being an over estimate.

Possibly: Keep track of overall sampling speed (over entire render) and calculate how much longer it will take from that. If saving time is included in calculating sample speed, the ETA will be more accurate. Or just use the last few minutes of SPS instead. (might be more accurate, but a little less precise)

nib9888:
Closes #697. I can confirm that the SPS value displayed in the ui is now more stable and consistent. Currently this PR averages over the past 150 frames' SPS values, but this is completely arbitrary and subject to change. Plus, I've not set this new behavior to be configurable, but if there is enough demand, I'll try to add that.

I'm completely new to java coding, so let me know if anything isn't correct or is against style guides etc.
Also, I've marked this as a draft because I'm not used to chunky's codebase, so had no idea where to add statements to clear the list used in computing the averages on scene reset (which I think would be preferable), so am after a bit of guidance in that respect.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
+++ b/chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
@@ -193,7 +193,9 @@ public class ChunkyFxController
   static class GUIRenderListener implements RenderStatusListener {
     private final ChunkyFxController gui;
     private int spp;
-    private int sps;
+    private LinkedList<Integer> sps_queue = new LinkedList<Integer>();
+    private static final int AVERAGE_PERIOD = 150;
+    private int sps_average;
 
     public GUIRenderListener(ChunkyFxController renderControls) {
       this.gui = renderControls;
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
```diff
@@ -193,7 +193,9 @@ public void showRenderPreview() {
   static class GUIRenderListener implements RenderStatusListener {
     private final ChunkyFxController gui;
     private int spp;
-    private int sps;
+    private LinkedList<Integer> sps_queue = new LinkedList<Integer>();
+    private static final int AVERAGE_PERIOD = 150;
```
leMaik, line 197: A `LinkedList` has a lot of overhead. We could simply use a `int[AVERAGE_PERIOD]` instead.

chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
```diff
@@ -210,7 +212,24 @@ public GUIRenderListener(ChunkyFxController renderControls) {
     }
 
     @Override public void setSamplesPerSecond(int sps) {
-      this.sps = sps;
+      // Rolling average of SPS
```
leMaik, line 215: Isn't this method already called with the averages `sps`?  Shouldn't this be calculated from the spp and the time delta one step earlier?

leMaik (changes requested):

chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
nib9888: Oh yeah - That was left over from when I was calculating the average slightly differently

chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
nib9888: Whoops - missed that
I thought we didn't calculate the average at all

nib9888:
I've tried again, and this should now work properly and be ready for merging

leMaik:
Sorry, I've been pretty busy lately. The code looks fine, I'll refactor the moving average logic into a new class and merge it. Thanks a lot! 👍

leMaik:
Superseded by #1911 but this inspired me to find a solution without a buffer, so thank you! :clap:

PR #1502, opened 2022-11-23
Rename Skymap modes; fix scene loading issue

Peregrine05:
- Renamed `Skymap (panoramic)` to `Skymap (equirectangular)`.
- Renamed `Skymap (spherical)` to `Skymap (angular)`.
- Fixed failure to load an angular skymap of a saved scene.
- Retain backwards-compatibility with scenes saved in older versions of Chunky.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
@@ -681,11 +681,11 @@ public class Sky implements JsonSerializable {
     skyExposure = json.get("skyExposure").doubleValue(skyExposure);
     skyLightModifier = json.get("skyLight").doubleValue(skyLightModifier);
     apparentSkyLightModifier = json.get("apparentSkyLight").doubleValue(apparentSkyLightModifier);
-    if (!(Objects.equals(json.get("mode").stringValue(mode.name()), "SKYMAP_PANORAMIC") || Objects.equals(json.get("mode").stringValue(mode.name()), "SKYMAP_SPHERICAL"))) {
+    if (!(json.get("mode").stringValue(mode.name()).equals("SKYMAP_PANORAMIC") || json.get("mode").stringValue(mode.name()).equals("SKYMAP_SPHERICAL"))) {
       mode = SkyMode.get(json.get("mode").stringValue(mode.name()));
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
```diff
@@ -681,7 +681,13 @@ public void importFromJson(JsonObject json) {
     skyExposure = json.get("skyExposure").doubleValue(skyExposure);
     skyLightModifier = json.get("skyLight").doubleValue(skyLightModifier);
     apparentSkyLightModifier = json.get("apparentSkyLight").doubleValue(apparentSkyLightModifier);
-    mode = SkyMode.get(json.get("mode").stringValue(mode.name()));
+    if (!(Objects.equals(json.get("mode").stringValue(mode.name()), "SKYMAP_PANORAMIC") || Objects.equals(json.get("mode").stringValue(mode.name()), "SKYMAP_SPHERICAL"))) {
```
leMaik, line 684: What's the benefit of using `Objects.equals` over the `.equals` method?

chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
```diff
@@ -699,7 +705,8 @@ public void importFromJson(JsonObject json) {
     color.set(JsonUtil.vec3FromJsonArray(json.get("color")));
 
     switch (mode) {
-      case SKYMAP_PANORAMIC: {
+      case SKYMAP_EQUIRECTANGULAR:
+      case SKYMAP_ANGULAR: {
```
leMaik, line 709: Nice!

chunky/src/java/se/llbit/chunky/renderer/scene/Sky.java
NotStirred: Objects.equals first does a reference equality `==`, then checks for null and then does `.equals`
Though in this case they can never be `==` as (if I'm reading the brackets correctly) one is a string literal,
so `"SKYMAP_PANORAMIC".equals(json.get("mode").stringValue(mode.name()))` would probably be preferred

Peregrine05: I have set it to be `json.get("mode").stringValue(mode.name()).equals("SKYMAP_PANORAMIC")`.

PR #1504, opened 2022-11-25
Add render region mode.

ThatRedox:
Add settings to crop the camera frame (similar to lens shift).
image
These allow for the advanced user to easily render multiple parts of an image and recombine them later:
split (Medium)
sunset (Medium)

The `Width` and `Height` entries are the new "virtual" image width and height, of which Chunky will render a portion of. The width & height in the `Scene` tab along with the `Crop X Offset` and `Crop Y Offset` determine the exact portion of the full image to render.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/PathTracingRenderer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/PathTracingRenderer.java
@@ -51,13 +51,19 @@ public class PathTracingRenderer extends TileBasedRenderer {
   @Override
   public void render(DefaultRenderManager manager) throws InterruptedException {
     Scene scene = manager.bufferedScene;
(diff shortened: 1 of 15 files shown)


ThatRedox:
Note: Currently builds are failing due to an expired certificate.

leMaik:
I wonder if we should switch _full width_ and _width_ (and height). The _width_ in the crop settings is now the "scene width", whereas the scene width is the width of the cropped part that is rendered… I find that a bit confusing. :thinking:

ThatRedox:
New UI:
image
image

leMaik:
Is the canvas size the virtual size now?

ThatRedox:
Yes, the canvas size is now the virtual size.

Issue #608, opened 2020-06-07
2.1 Glass reduces intensity of emitters significantly


jackjt8:
Similar to https://github.com/llbit/chunky/issues/528 but with glass instead. 

1.4.5
![NoiseTest_a_1_water_glass-256](https://user-images.githubusercontent.com/3740917/83980998-ec2afd80-a911-11ea-8a2f-1d8a8cd9ebe4.png)

2.0beta6
![NoiseTest_a_2_water_glass-256](https://user-images.githubusercontent.com/3740917/83980999-edf4c100-a911-11ea-868f-1f951f7e3358.png)

2.1-274
![NoiseTest_a_274_water_glass-256](https://user-images.githubusercontent.com/3740917/83981003-f3520b80-a911-11ea-950f-f951c1faba98.png)

2.1-308
![NoiseTest2_water_glass-256](https://user-images.githubusercontent.com/3740917/83981028-21cfe680-a912-11ea-8461-b96336a20711.png)

jackjt8:
Scene
[scenes.zip](https://github.com/llbit/chunky/files/4743927/scenes.zip)

aTom3333:
![NoiseTest2_water_glass-256](https://user-images.githubusercontent.com/23342398/84033814-be899700-a999-11ea-9000-e598e683893c.png)
I went back to see when that was introduced. It was when `localIntersect` for class was made to be `false`. Setting it back to `true` fixes this issue but we don't get refraction anymore. So i'll have to look further

leMaik:
@aTom3333 We would need to remember that the ray already hit a glass block of that color (`previousMaterial` of the ray) and then skip intersection in the `intersect` method of the glass model. This could make glass blocks behind glass blocks (glass air glass, like in your image) invisible, I'll have to try it out.

Refraction shouldn't be affected by this change. :thinking: It works fine for e.g. slime which also uses local intersection.

Stenodyon:
If `localIntersect` is `true` then the ray goes through the glass instead of generating an intersection that can refract. This is because `TexturedBlockModel.intersect(…)` makes a decision about intersecting the glass block based on the texture alpha but not the refraction index.
https://github.com/llbit/chunky/blob/ed0d57550d6f836690b4ae5ade465f2b7eac42ed/chunky/src/java/se/llbit/chunky/model/TexturedBlockModel.java#L60-L65
[...]

PR #1513, opened 2022-12-10
Improve how translucent materials handle color

JustinTimeCuber:
Instead of using the texture's opacity directly to determine how much to filter the inbound ray color, the path tracer should first determine how much total light would be transmitted out of white inbound light if each color channel were simply multiplied, that is, transmitted light = (R + G + B)/3. Then, this is scaled based on the desired opacity as given by the texture. The result is that at least for white light, the total proportion of light that is transmitted is equal to 1 minus the alpha value of the texture. For example, a pure red texture with 33% transparency (alpha = 0.67) would transmit 100% of red light and 0% of other light. The way Chunky currently works, it would transmit 100% of red light and 33% of green and blue light, which is less accurate since 33% total transparency is already achieved by allowing the red light to pass.

One caveat is that when a texture has an unusually low opacity but a high color saturation, it could result in that color being amplified as it passes through while the others are blocked. This could be realistic in some situations, i.e. absorbing light at one frequency and re-emitting it at another, but depending on whether this has unwanted effects within Minecraft's block palette, the algorithm could be modified to better handle such cases.

@ThatRedox also wrote a potential fix to #608 which I originally included in my fork but isn't currently included in this PR due to it needing more testing. However, I have included examples both with and without this fix. For reference, [here](https://github.com/chunky-dev/chunky/commit/72aecbe51b3925991bb2cd51a69e4c29cbbdf3d1) is the commit where this fix was removed.

[default_2022-12-10_13-57-28.zip](https://github.com/chunky-dev/chunky/files/10200843/default_2022-12-10_13-57-28.zip)
Before:
image
After:
image
After (with Redox's fix):
image

[default_2022-12-10_01-27-54.zip](https://github.com/chunky-dev/chunky/files/10200845/default_2022-12-10_01-27-54.zip)
Before:
image
After:
image
After (with Redox's fix):
image

(scene is too big to upload)
Before:
image
After:
image
After (with Redox's fix):
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
@@ -343,9 +343,23 @@ public class PathTracer implements RayTracer {
                 }
 
                 if (pathTrace(scene, refracted, state, 1, false)) {
-                  ray.color.x = ray.color.x * pDiffuse + (1 - pDiffuse);
-                  ray.color.y = ray.color.y * pDiffuse + (1 - pDiffuse);
-                  ray.color.z = ray.color.z * pDiffuse + (1 - pDiffuse);
+                  // Color-based transmission value
+                  double colorTrans = (ray.color.x + ray.color.y + ray.color.z)/3;
+                  // Total amount of light we want to transmit (overall transparency of texture)
+                  double shouldTrans = 1 - pDiffuse;
+                  // Amount of each color to transmit - default to overall transparency if RGB values add to 0 (e.g. regular glass)
+                  double rTrans = shouldTrans, gTrans = shouldTrans, bTrans = shouldTrans;
+                  if(colorTrans > 0) {
+                    // Amount to transmit of each color is scaled so the total transmitted amount matches the texture's transparency
+                    rTrans = ray.color.x * shouldTrans / colorTrans;
+                    gTrans = ray.color.y * shouldTrans / colorTrans;
+                    bTrans = ray.color.z * shouldTrans / colorTrans;
+                  }
+                  // Set transparent and opaque components of each color channel
+                  ray.color.x = (1 - rTrans) * ray.color.x + rTrans;
+                  ray.color.y = (1 - gTrans) * ray.color.y + gTrans;
+                  ray.color.z = (1 - bTrans) * ray.color.z + bTrans;
+                  // Scale by results from refracted ray
                   ray.color.x *= refracted.color.x;
                   ray.color.y *= refracted.color.y;
                   ray.color.z *= refracted.color.z;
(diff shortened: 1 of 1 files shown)


JustinTimeCuber:
New transmissivity cap [commit](https://github.com/chunky-dev/chunky/pull/1513/commits/8b031ae3164e115ffc7d795ef313b7d8566bc1d7):
After:
image
Before:
image

leMaik:
@JustinTimeCuber These changes look great! :clap: I allowed actions to run and just merged #1514, so this PR requires a rebase.

JustinTimeCuber:
I rebased the branch, it was a bit tricky since we were modifying the same file, but it seems like it's working as expected.

As for the per-color cap, currently there is a constant `TRANSMISSIVITY_CAP = 1` (line 458 in PathTracer.java). A straightforward option would be to add a slider for this value from 1 to 3 (I think I personally prefer it in the 1.25 to 1.5 range, maybe 1.25 would be a good default?) Below are some examples. Note that even with the minimum cap of 1, the colors are noticeably more saturated than in master. Also, emitters shining through colored glass are moderately brighter than in master, regardless of the cap. This is due to forcing each block to pass through a proportion of white light equal to 1 - opacity, while in master, less total light is passed due to color filtering.

**Sea Lanterns through various translucent blocks**
Master
image
Cap = 1
image
Cap = 1.25
image
Cap = 1.5
image
Cap = 2
image

**Higher-saturation emitters through red, yellow, lime, and blue stained glass**
Master
image
Cap = 1
image
Cap = 1.25
image
Cap = 1.5
image
Cap = 2
image

**Light filtering**
Master
image
Cap = 1 
image
Cap = 1.25
image
Cap = 1.5
image
Cap = 2
image

leMaik:
I do like cap 1.5, I'm just wondering... Is a cap of >1 creating energy?

JustinTimeCuber:
> I do like cap 1.5, I'm just wondering... Is a cap of >1 creating energy?

In some cases involving colored light passing through its own color glass, it could do that. For white light, it is only redistributing energy of one color to a different color. In other words, it would allow per-channel transmissivity to be R=1.3, G=0.7, B=0.4 whereas cap = 1 would normalize this to R=1, G=0.8, B=0.6 (which gives you the same hue of color and adds to the same amount).

I think it would be relatively straightforward to check the ratio between energy in and out, and then in the (unlikely) case that it's greater than 1, normalize it. This would prevent any energy gains.

leMaik:
@JustinTimeCuber Thank you for adding the additional check, this makes me more comfortable with merging this PR. :upside_down_face:

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
```diff
@@ -459,6 +441,71 @@ public static boolean pathTrace(Scene scene, Ray ray, WorkerState state, int add
     return hit;
   }
 
+  private static void translucentRayColor(Scene scene, Ray ray, Ray next, double opacity) {
+    // Color-based transmission value
+    double colorTrans = (ray.color.x + ray.color.y + ray.color.z)/3;
+    // Total amount of light we want to transmit (overall transparency of texture)
+    double shouldTrans = 1 - opacity;
+    // Amount of each color to transmit - default to overall transparency if RGB values add to 0 (e.g. regular glass)
+    double rTrans = shouldTrans, gTrans = shouldTrans, bTrans = shouldTrans;
```
leMaik, line 450: Would the code get better with `trans = new Vector3(shouldTrans, shouldTrans, shouldTrans)` ?

chunky/src/res/style.css
```diff
@@ -31,6 +31,12 @@ Label {
     -fx-text-fill: #faebd7;
 }
 
+Label.group-label {
```
leMaik, line 34: Nice! :clap:

leMaik (approved):

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
JustinTimeCuber: I just did that, was able to eliminate a few repeated statements

JustinTimeCuber:
@leMaik Do you think the default transmissivity cap should stay at 1, which is arguably a bit more realistic (and less arbitrary) or should I change it to e.g. 1.25 or 1.5, which generally looks better? With my changes a couple weeks ago, there should be no chance of energy being created regardless of the cap value.

JustinTimeCuber:
Also I did something in IntelliJ that I thought was only updating the local copy but for some reason it pushed it to github? Whatever I did, hopefully it didn't break anything. Still figuring this out

leMaik:
@JustinTimeCuber I don't really know :thinking: Maybe stick to 1.0 then?

JustinTimeCuber:
Okay, let me know if there are any other changes I should make

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
```diff
@@ -105,7 +105,7 @@ public static boolean pathTrace(Scene scene, Ray ray, WorkerState state, int add
 
       float pSpecular = currentMat.specular;
 
-      double pDiffuse = ray.color.w;
+      double pDiffuse = 1 - Math.sqrt(1 - ray.color.w);
```
leMaik, line 108: This has probably come up on Discord before, but is this physically accurate? If not, should we make this change configurable?

leMaik (approved):

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
JustinTimeCuber: I think it comes down to interpretation so maybe it wouldn't hurt to add it as an option? But I think it physically makes the most sense, as one would expect 50% opacity to mean 50% of light gets reflected OR absorbed, whereas the way Chunky currently interprets it is "50% gets reflected, 25% gets transmitted, and the remainder is partially transmitted depending on the texture color".

PR #1518, opened 2022-12-28
Reduce size of light block model

Peregrine05:
- Reduced the size of the light block model from a full block to 1/2 block length cubed.
- Quadrupled default emittance values for light blocks to compensate.

Previously, light blocks would unnaturally fully illuminate any texture directly adjacent to it. By reducing the size of the model, this effect is eliminated.

**Old:**
light_block_test-32

**New: (actual results are similar; this is an older render)**
light_block_test-32 2


leMaik:
This looks way better, thank you! :clap: 
I'll also backport this to the next Chunky 2.4 release.

PR #1478, opened 2022-10-19
More details in error messages in MCA chunk loading

ShirleyNekoDev:
- rewrite error handling of chunk loading
- add some more documentation

PR #1522, opened 2022-12-31
Implement entity regions

NotStirred:
Previously I'd wanted to wait until I'd moved CubicChunks support out, and could rewrite the region parsing altogether.

This implementation isn't ideal but should be fine for now.

Ideally the first three commits should be squashed with the last being separate, let me know if I should do that or if it'll be done in the merge

image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/world/region/MCRegion.java
+++ b/chunky/src/java/se/llbit/chunky/world/region/MCRegion.java
@@ -29,6 +29,8 @@ import se.llbit.nbt.ErrorTag;
 import se.llbit.nbt.NamedTag;
 import se.llbit.nbt.Tag;
 import se.llbit.util.Mutable;
+import se.llbit.util.annotation.NotNull;
+import se.llbit.util.annotation.Nullable;
 
 /**
  * Abstract region representation. Tracks loaded chunks and their timestamps.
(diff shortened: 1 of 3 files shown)


ShirleyNekoDev (commented):
Some changes here might have minor conflicts with #1478, which can be fixed here after merging it, but nothing serious as far as I can see. Nice to have paintings back 🎉

NotStirred:
I see - I'll rebase onto master once #1478 is merged.
I'll also review it to hopefully inspire a merge xd

leMaik:
Thanks @ShirleyNekoDev and also for the code review @NotStirred. :+1:
I'll merge this once it's rebased. I'd also like to backport this PR into the next 2.4 release; could you re-open this PR against the chunky-2.4.x branch? I'm not sure if I should backport #1478.

NotStirred:
Not sure about backporting #1478. It would be nice to have and I don't see much possibility for bugs, but it is an unnecessary change for an old version, as it's just a refactor.

Let me know if you're planning on it though, as I'll wait until that PR is opened/merged before backporting this one

leMaik:
@NotStirred Allright, I won't backport #1478 then.

chunky/src/java/se/llbit/chunky/world/region/MCRegion.java
```diff
@@ -253,6 +284,11 @@ private static ChunkDataSource readChunkData(RandomAccessFile file, ChunkPositio
     long index = getMCAChunkIndex(chunkPos);
 
     long length = file.length();
+
+    if (length == 0) {
+      return null; // vanilla will occasionally save empty region files, we shouldn't warn the user about these.
```
leMaik, line 289: Nice! :clap:

chunky/src/java/se/llbit/chunky/world/region/MCRegion.java
```diff
@@ -263,7 +299,7 @@ private static ChunkDataSource readChunkData(RandomAccessFile file, ChunkPositio
     int locationEntry = file.readInt();
     int sectorCount = locationEntry & 0xFF;
     int sectorOffset = locationEntry >> 8;
-    if (sectorOffset == 0 && sectorCount == 0) {
+    if (sectorOffset == 0 || sectorCount == 0) {
```
leMaik, line 302: This change doesn't seem right. Is it bad if the sectorOffset is 0 but the sectorCount is not zero?
```suggestion
    if (sectorOffset == 0 && sectorCount == 0) {
```

NotStirred: If either is zero, the header entry is invalid.
Was concerned that we would be running through the later code with an invalid header entry.

NotStirred: Also now that I think about it, a `sectorOffset` of `1` is also invalid as that points to the timestamp sector.
We probably don't need to think about that though.

leMaik (approved):

PR #1527, opened 2023-01-06
fix NPE when loading scenes with missing worlds

ShirleyNekoDev:


leMaik:
Thank you; this also seems to affect 2.4 and I'll backport the fix.

PR #1528, opened 2023-01-06
Move entity related stuff into separate class

ShirleyNekoDev:
Another PR in the quest for a cleaner Scene class :)


NotStirred (approved):
Haven't tested locally yet, but the changes look fab :+1:

leMaik:
Moving code out of `Scene` is always much appreciated. :+1: Not sure if we should move the BVHs back into the scene eventually. The octree is in the scene too, and not in some `SceneBlocks` class. :thinking:

PR #1529, opened 2023-01-07
Fix issue where the auto-update popup appears over first time setup.

### LINKED ISSUE #1530: First time setup and auto-update run simultaneously
While testing #1292 we discovered that since probably #1320 there is a regression that breaks the first time setup due to running auto-update at the same time, which already creates directories.

This bug is not present in 1.13.1 (ie. the latest release version of the launcher).

ThatRedox:
Fixes #1530


leMaik:
_looks at issue number_ ... _looks at pr number_ ...

image

PR #1531, opened 2023-01-07
Do not limit water plane y-values and change `Sun Radius` to degrees.

### LINKED ISSUE #1517: Display external values for `Sun radius` and `Apparent sun radius` in degrees.
Currently, these values are displayed with a base value of 1, with an internal value of 0.03 radians. Displaying this value in degrees would be more intuitive.

### LINKED ISSUE #1524: Do not clamp the range of the Water plane height.

ThatRedox:
Closes #1517 
Closes #1524


ShirleyNekoDev (approved):

leMaik:
Does this make the scene files incompatible?

ThatRedox:
No, sun radius was always stored in radians.

PR #1535, opened 2023-01-13
Increase sea lantern light level

### LINKED ISSUE #1534: Set default emittance value for `minecraft:sea_lantern` to 1
As sea lanterns are a full block with a light level of *15*, they should have a default emittance value of *1*, not *0.5*.

JustinTimeCuber:
As mentioned in the issue, sea lanterns are light level 15 and therefore should have an emittance of 1.0.


NotStirred (approved):

leMaik:
Nice, thannk you!

PR #1543, opened 2023-02-09
Add a choice box for fog type

Peregrine05:
This does not implement a GUI for layered fog. It only adds a simpler way to enable and disable fog.

**New:**
image

image

image


Peregrine05:
I should probably change that hint.

PR #1548, opened 2023-02-15
(WIP) Add blocks from 22w42a and 23w07a

JustinTimeCuber:
Here's my current progress
image
Here's a screenshot of the test world

Test scene
[1.20-angle1.zip](https://github.com/chunky-dev/chunky/files/10747306/1.20-angle1.zip)

What's currently missing: piglin heads, pink petals, and hanging signs.


chunky/src/java/se/llbit/chunky/resources/Texture.java
```diff
@@ -1122,6 +1122,100 @@ public class Texture {
...
+  public static final Texture chiseledBookshelfTop = new Texture();
+
+  public static final Texture[] chiseledBookshelfCombinations = new OverlaidTexture[64];
+  static {
+    for(int i = 0; i < chiseledBookshelfCombinations.length; i++) {
+      int j = i; // it complains if I don't do this
+      chiseledBookshelfCombinations[i] = new OverlaidTexture(Texture.chiseledBookshelfEmpty, Texture.chiseledBookshelfOccupied, (x, y) ->
+        ((j & 1) != 0 && x < 1f/3 && y < 1f/2) || ((j & 2) != 0 && x >= 1f/3 && x < 2f/3 && y < 1f/2) || ((j & 4) != 0 && x >= 2f/3 && y < 1f/2)
+        || ((j & 8) != 0 && x < 1f/3 && y >= 1f/2) || ((j & 16) != 0 && x >= 1f/3 && x < 2f/3 && y >= 1f/2) || ((j & 32) != 0 && x >= 2f/3 && y >= 1f/2)
+      );
```
leMaik, line 1193: Can you refactor this so that it is readable?

chunky/src/java/se/llbit/chunky/model/QuadModel.java
```diff
@@ -15,6 +16,43 @@
...
+  public static final Quad FULL_BLOCK_TOP_SIDE = new Quad(
+    new Vector3(1, 1, 0),
+    new Vector3(0, 1, 0),
+    new Vector3(1, 1, 1),
+    new Vector4(1, 0, 1, 0));
+  public static final Quad FULL_BLOCK_BOTTOM_SIDE = new Quad(
+    new Vector3(0, 0, 0),
+    new Vector3(1, 0, 0),
+    new Vector3(0, 0, 1),
+    new Vector4(0, 1, 0, 1));
```
leMaik, line 48: Note that these quads were rotated specifically such that `TopBottomOrientedTexturedBlockModel` textures are rotated correctly on all sides. This may not apply to any full cube blocks.

chunky/src/java/se/llbit/chunky/resources/Texture.java
JustinTimeCuber: Done, let me know if it is good

chunky/src/java/se/llbit/chunky/model/QuadModel.java
JustinTimeCuber: is there already a model that rotates the sides but not the top/bottom, or would that need to be created?

ShirleyNekoDev: As far as I understand, the Quad's orientation should each currently match the correct coordinate axis when not rotated.
I wanted to work on more utilities for block models in the future - having these constants should also be helpful for them. (And should reduce the global object count if reused in multiple models, right?)

chunky/src/java/se/llbit/chunky/model/DecoratedPotModel.java
```diff
@@ -0,0 +1,169 @@
...
+      case "minecraft:pottery_shard_skull":
+        return Texture.decoratedPotPatternSkull;
+      case "minecraft:pottery_shard_prize":
+        return Texture.decoratedPotPatternPrize;
+      case "minecraft:pottery_shard_archer":
+        return Texture.decoratedPotPatternArcher;
+      case "minecraft:pottery_shard_arms_up":
+        return Texture.decoratedPotPatternArmsUp;
+      default:
+        throw new IllegalArgumentException("Unknown pottery shard: " + shard);
```
leMaik, line 156: For better resilience, we could instead lof a warning and use `Texture.decoratedPotSide`.

PR #1555, opened 2023-02-28
Add apparent emitter brightness controls

Peregrine05:
- Renamed `Emitter intensity` to `Emitter brightness`, and changed it to scale both emitter light intensity and apparent emitter brightness. Changed its default value to *1*.
- Added `Emitter light intensity modifier`, which scales only the emitter light intensity. Its default value is *13*, to reflect the default value of `Emitter intensity` from previous versions.
- Added `Apparent emitter brightness modifier`, which scales only the apparent brighness of emitters. Its default value is *1*.
- Added an `Apparent brightness` material property, which changes the apparent brightness of the material when it behaves as an emitter. Its default value is *1*
- When loading scenes saved in an older Chunky version, Chunky will load the value of `Emitter brightness` (previously `Emitter intensity`) as the value of `Emitter light intensity modifier`, and set `Emitter brightness` to *1*.

**New:**
image
image

This PR could still use some testing, specifically with ESS. The PathTracer code might also be more complicated than it need be; we should check if there is any way to achieve the same results with simpler code (This may no longer be the case).


Peregrine05:
Fixed by #1627 and #1621.

PR #1557, opened 2023-03-04
Add an API to get the canvas / map coordinate in the context menu callback.

ThatRedox:
This sets the user data of the context menu to the `Vector2` representing the coordinate.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/main/Chunky.java
+++ b/chunky/src/java/se/llbit/chunky/main/Chunky.java
@@ -479,7 +479,7 @@ public class Chunky {
 
   /**
    * Get the mutable list of context menu transformers in the main map view.
-   * The map X and Y coordinates are supplied as a `Vector2` in ContextMenu.getUserData()
+   * The map view X and Y coordinates are supplied as a `Vector2` in the context menu's `mapviewPosition` property.
    */
   @PluginApi
   public List<ContextMenuTransformer> getMapContextMenuTransformers() {
@@ -489,7 +489,7 @@ public class Chunky {
 
   /**
    * Get the mutable list of context menu transformers in the main render view.
-   * The canvas X and Y coordinates are supplied as a `Vector2` in ContextMenu.getUserData()
+   * The canvas X and Y coordinates are supplied as a `Vector2` in the context menu's `canvasPosition` property.
    */
   @PluginApi
   public List<ContextMenuTransformer> getRenderContextMenuTransformers() {
--- a/chunky/src/java/se/llbit/chunky/main/Chunky.java
+++ b/chunky/src/java/se/llbit/chunky/main/Chunky.java
@@ -480,8 +480,8 @@ public class Chunky {
   /**
    * Get the mutable list of context menu transformers in the main map view. The following are supplied as context menu properties:
    * <ul>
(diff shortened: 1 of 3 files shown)


chunky/src/java/se/llbit/chunky/ui/RenderCanvasFx.java
```diff
@@ -247,6 +247,7 @@ public RenderCanvasFx(ChunkyFxController chunkyFxController,
         double halfWidth = canvas.getWidth() / (2.0 * canvas.getHeight());
         target.set(-halfWidth + event.getX() * invHeight,
             -0.5 + event.getY() * invHeight);
+        contextMenu.setUserData(new Vector2(event.getX(), event.getY()));
```
NotStirred, line 250: Any reason not to use properties? imo it's more readable
```suggestion
        contextMenu.getProperties().put("canvasPos", new Vector2(event.getX(), event.getY()));
```

ThatRedox: Didn't know you could do that 👍

chunky/src/java/se/llbit/chunky/main/Chunky.java
```diff
@@ -479,6 +479,7 @@ public TabTransformer getMainTabTransformer() {
 
   /**
    * Get the mutable list of context menu transformers in the main map view.
+   * The map view X and Y coordinates are supplied as a `Vector2` in the context menu's `mapviewPosition` property.
```
ShirleyNekoDev, line 482: Could you please clarify the unit for the coordinate system. Is it in pixels of the FX pane or in Minecraft blocks? Ideally I'd like to have access to both in different properties :)

chunky/src/java/se/llbit/chunky/main/Chunky.java
```diff
@@ -488,6 +489,7 @@ public List<ContextMenuTransformer> getMapContextMenuTransformers() {
 
   /**
    * Get the mutable list of context menu transformers in the main render view.
+   * The canvas X and Y coordinates are supplied as a `Vector2` in the context menu's `canvasPosition` property.
```
ShirleyNekoDev, line 492: Also, could you please clarify the origin for the coordinate system (which corner) and if it is for the surrounding FX element or the canvas itself (and how does scaling affect these? Do you get the pixel location for lookup in the render buffer or something on which you have conversion to do?).

ShirleyNekoDev (changes requested):

chunky/src/java/se/llbit/chunky/main/Chunky.java
ThatRedox: Looks like its in pixels of the FX pane. I updated the Javadocs and also added a property with the chunk coordinates (multiply by 16 to get blocks).

chunky/src/java/se/llbit/chunky/main/Chunky.java
ThatRedox: It's the coordinate of the canvas which is the same as the pixel location on the render buffer. I updated the javadocs.

ShirleyNekoDev (approved):

NotStirred (approved):
New additions look great :+1:

NotStirred:
Thoughts on adding a comment next to any added properties that they're PluginAPI?

leMaik (approved):

PR #1558, opened 2023-03-06
more resilient error handling in AsynchronousSceneManager

ShirleyNekoDev:
- use an single thread ExecutorService instead of self-made thread handling
- also move scene name sanitization out into separate class

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
@@ -30,7 +30,8 @@ import se.llbit.util.TaskTracker;
 import java.io.File;
 import java.io.IOException;
 import java.util.Collection;
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 /**
  * This scene manager is used for asynchronous loading and saving of scenes.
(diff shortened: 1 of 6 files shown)


leMaik (approved):

chunky/src/java/se/llbit/chunky/renderer/scene/AsynchronousSceneManager.java
```diff
@@ -39,16 +40,39 @@
...
 
   public AsynchronousSceneManager(RenderContext context, RenderManager renderManager) {
-    super("Scene Manager");
-
     sceneManager = new SynchronousSceneManager(context, renderManager);
-    taskQueue = new LinkedBlockingQueue<>();
+  }
+
+  /**
+   * Schedule a task to be run soon.
```
leMaik, line 72: Maybe we should add a note on why one would want to use this, i.e. to safely modify the scene without race conditions?

ShirleyNekoDev: Feel free to suggest a descriptive JavaDoc ~ I'm not sure myself when to use a task with this, I just updated the code ^^
Seems like all operations in the AsyncSceneManager are enqueued with this.

leMaik: Probably because that allows it to be async and not do expensive operations in the UI thread. :brain:

ShirleyNekoDev: Well yes, I see that 😅 
But this seems too obvious to state in a note here (?) Did you have anything else in mind? 🙈

PR #1559, opened 2023-03-06
Add ability to reuse initial rays multiple times (performance improvement)

JustinTimeCuber:
## Overview
The first rays being cast from the camera are always more or less* the same. Therefore, it is wasteful to re-cast each initial ray for every sample. Instead, this PR creates a setting for how many "branches" each ray should take after its first hit (default: 10). My testing indicates no visual drawback in most* situations doing e.g. 50 initial rays with 10 branches compared to a standard 500 SPP in master. However, such a render would complete faster - in my testing, typical performance gains range from about 5% on the low end (indoor scenes with high ray depth) up to about 50% on the high end (outdoor scenes with low ray depth). Additionally, I have found that more than 10-20 branches generally makes little to no additional performance benefit, while potentially being detrimental* to visual quality.

Here are a few performance comparisons. Of course, these results may vary based on hardware, OS, JVM, etc. Note that "samples per second" in the context of the PR refers to the "true" SPP multiplied by the branch count, and the UI has been updated to reflect that as well. (There is also a UI option under the Advanced tab to control the branch count.)
AnotherTestScene
Cornell Box
Dodecahedron
I can provide the scene files if needed, although it would probably be best to test with a wider variety of scenes.

## *Caveats
Each initial ray being cast from a particular pixel is not exactly the same. They are randomized within the bounds of the pixel, which, after many samples, creates an anti-aliasing effect. Therefore, it is still necessary to randomize the initial rays _sometimes_. From my testing, it seems as though the PR produces pretty much indistinguishable results as master (except for the specific pattern of noise, of course) when the "true" SPP (equivalent SPP divided by branch count) is about 40 or more. Of course, these results will vary depending on scene, resolution, etc. But for an extreme example, here is the Dodecahedron render with 2000 branches and a "true" SPP of 1:
image

A small caveat is that this would increase the amount of time it takes Chunky to respond to e.g. a camera angle change while in path tracing mode (preview mode should be unaffected). The workaround would be setting the slider to 1 before modifying the scene in this way.

Another potential issue I have noticed is that the denoiser plugin seems to render 1 SPP before switching to the normal and albedo maps, and then it discards that initial SPP and starts over with the main render. With this PR, the denoiser plugin therefore takes much longer to start a render (roughly proportional to your branch count) unless you use the workaround mentioned above. However, I think this should be considered a bug in the denoiser plugin and fixed there.


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -140,6 +140,16 @@ public class Scene implements JsonSerializable, Refreshable {
...
 
+  /**
+   * Minimum branch count.
+   */
+  public static final int MIN_BRANCH_COUNT = 1;
+
+  /**
+   * Maximum branch count.
+   */
+  public static final int MAX_BRANCH_COUNT = 50;
```
leMaik, line 151: These are pretty much GUI-only constants, should we move them into the advanced tab?

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
```diff
@@ -123,270 +123,31 @@ public static boolean pathTrace(Scene scene, Ray ray, WorkerState state, int add
         // TODO: material color may change here.
         continue;
       }
-
+      if(ray.depth + 1 >= scene.rayDepth) {
+        break;
+      }
```
leMaik, line 128: A note on how this is the Roussian Roulette technique would be nice.

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
```diff
@@ -123,270 +123,31 @@ public static boolean pathTrace(Scene scene, Ray ray, WorkerState state, int add
...
-              } else if(indirectEmitterColor.x > Ray.EPSILON || indirectEmitterColor.y > Ray.EPSILON || indirectEmitterColor.z > Ray.EPSILON) {
-                hit = true;
-                ray.color.x *= indirectEmitterColor.x;
-                ray.color.y *= indirectEmitterColor.y;
-                ray.color.z *= indirectEmitterColor.z;
-              }
-            }
-          }
+      int count = firstReflection ? scene.getBranchCount() : 1;
+      for (int i = 0; i < count; i++) {
```
leMaik, line 134: A comment on this technique (is there a paper or some other source on this?) would be great.

Just an idea: Could we re-use more than the first ray? Eg. re-use entire ray paths _but_ the last ray?

chunky/src/java/se/llbit/chunky/ui/render/tabs/AdvancedTab.java
```diff
@@ -101,6 +102,16 @@ public void initialize(URL location, ResourceBundle resources) {
     rayDepth.setRange(1, 25);
     rayDepth.clampMin();
     rayDepth.onValueChange(value -> scene.setRayDepth(value));
+
+    branchCount.setName("Branch count");
+    branchCount.setTooltip("Sets the number of rays cast after the first intersection.");
```
leMaik, line 107: This description doesn't mention the reuse of initial rays and the fact that this could affect the quality of low spp renders.

leMaik: Is `branchCount` a good name?

leMaik (changes requested):

leMaik:
Great to see these performance improvements! :clap: :rocket:

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
JustinTimeCuber: Seems reasonable, will do.

chunky/src/java/se/llbit/chunky/ui/render/tabs/AdvancedTab.java
JustinTimeCuber: Well, I was thinking of it as the number of times the ray branches at its first intersection. Do you have a better name in mind?

And I will try to edit the description to be more complete.

PR #1519, opened 2022-12-28
Add UE4 filmic tone mapping and make Hable tone mapping configurable.

leMaik:
I just wanted to render a few example images for the post-processing section of the docs, and somehow went down this rabbit hole.

1. [UE4 Filmic Tonemapping](https://docs.unrealengine.com/4.26/en-US/RenderingAndGraphics/PostProcessEffects/ColorGrading/) is now implemented; basically configurable ACES, defaults to matching the ACES curve (UE4 legacy tone mapping preset is also in the code already)
2. Hable was wrong (missing Gamma Correction and arbitrary x16 exposure)
3. Compared [Hable's GDC talk](https://dokumen.tips/documents/hable-john-uncharted2-hdr-lighting.html?page=1) with the [blog post](http://filmicworlds.com/blog/filmic-tonemapping-operators/) and noticed that there were different parameters; added presets for both
4. Hable and UE4 Filmic Tonemaps can now be configured ~~(JSON only until someone adds a UI)~~

PR #1560, opened 2023-03-06
Extract canvas size code from Scene into separate CanvasConfig class

ShirleyNekoDev:
This helps to reduce the size of the Scene class and is a step for implementing support for custom canvas buffers in the future.


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2043,95 +2019,21 @@ public void setTargetSpp(int value) {
...
-    return 0;
+  public synchronized void setCanvasCropSize(
+    int canvasWidth, int canvasHeight,
+    int fullWidth, int fullHeight,
+    int cropX, int cropY
+  ) {
+    boolean changedBuffers = canvasConfig.setSize(canvasWidth, canvasHeight);
+    boolean cropChanged = canvasConfig.setCropSize(fullWidth, fullHeight, cropX, cropY);
+    if(changedBuffers) initBuffers();
+    if(changedBuffers || cropChanged) refresh();
```
NotStirred, line 2036: chunky style
```suggestion
    if(changedBuffers) { 
      initBuffers();
    }
    if(changedBuffers || cropChanged) {
      refresh();
    }
```

NotStirred (approved):
lgtm - `Scene` is getting smaller!

chunky/src/java/se/llbit/chunky/renderer/scene/CanvasConfig.java
```diff
@@ -0,0 +1,191 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  public int getPixelCount() {
+    return width * height;
+  }
+
+  public void writeJsonData(JsonObject json) {
```
leMaik, line 163: We could use the `Configurable` interface from #1519 here. :thinking:

ShirleyNekoDev: This would require #1519 to be merged :D
Will do in a separate PR 👍

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2043,97 +2019,27 @@ public void setTargetSpp(int value) {
...
-    }
-    return 0;
-  }
-
-  public int getCropY() {
-    if (isCanvasCropped()) {
-      return cropY;
-    }
-    return 0;
-  }
```
ShirleyNekoDev, line 2135: deprecate them for plugin compatability

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -370,10 +349,10 @@ public static void exportToZip(File sceneDirectory, String name, File targetFile
...
    */
   public synchronized void initBuffers() {
-    frontBuffer = new BitmapImage(width, height);
-    backBuffer = new BitmapImage(width, height);
-    alphaChannel = new byte[width * height];
-    samples = new double[width * height * 3];
+    frontBuffer = new BitmapImage(canvasConfig.getWidth(), canvasConfig.getHeight());
+    backBuffer = new BitmapImage(canvasConfig.getWidth(), canvasConfig.getHeight());
+    alphaChannel = new byte[canvasConfig.getPixelCount()];
+    samples = new double[canvasConfig.getPixelCount() * 3];
```
leMaik, line 355: Maybe we could add a `Canvas` class that has these buffers and then one could apply the config to it?
`CanvasConfig` currently is the only class where we split config and logic, eg. sky bundles both. Not sure what's the better approach, but we should be consistent.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1909,95 +1882,73 @@ public void setTargetSpp(int value) {
...
 
-  public boolean isCanvasCropped() {
-    return fullWidth != 0 && fullHeight != 0;
-  }
-
   /**
-   * @return Canvas width
+   * @deprecated use {@link CanvasConfig}
    */
+  @Deprecated
```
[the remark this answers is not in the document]
ShirleyNekoDev: They were reinserted by request from @ThatRedox [see here](https://discord.com/channels/541221265512464394/545374333883777037/1138503126509420595).
Plugins depending on the fields or getters have to be updated either way, but I would have not problem with keeping them for another version before removing them.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
ShirleyNekoDev: That was the plan for another upcoming PR, but I wanted to keep the changes small here and do the big canvas buffer overhaul separately (see PR description).

leMaik: Sounds good. So `CanvasConfig` will be replaced by `Canvas` eventually?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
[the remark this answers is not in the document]
jackjt8: I wonder if we should add a comment to deprecated methods to state a removal version? That way we would know when to remove them. Otherwise, we could attempt to track down all plugins that use said method and track compliance via an issue?

chunky/src/java/se/llbit/chunky/renderer/scene/CanvasConfig.java
leMaik: I'll add the UI for that PR soon and then we just have to remember this. :sweat_smile:

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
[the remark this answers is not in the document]
leMaik: Let's keep these methods for the 2.5.x life cycle then.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
ShirleyNekoDev: Yes, something like that is the plan :)

chunky/src/java/se/llbit/chunky/renderer/scene/CanvasConfig.java
leMaik: Ping :upside_down_face:

PR #1561, opened 2023-03-06
Refactor RenderConfig interface

ShirleyNekoDev:
- renames methods to be more precise
- move render options from ChunkyOptions into separate interface/class
  (long term goal is to reduce the amount of context which is pushed around for rendering with the RenderManager/RenderContext)
- add customization factories for Rays and WorkerState in TileBasedRenderer (for plugins)


NotStirred (commented):
I like the idea in general, just a few comments: (github removed my comments for some reason)

lib/src/se/llbit/chunky/renderer/RenderOptions.java
```diff
@@ -16,36 +16,49 @@
...
+  /**
+   * Default CPU load
+   */
+  int TARGET_CPU_LOAD_PERCENTAGE = 100;
+
   /**
    * Default number of worker threads.
    * Is set to the number of available CPU cores.
    */
   int NUM_RENDER_THREADS_DEFAULT = Runtime.getRuntime().availableProcessors();
```
NotStirred, line 41: As `NUM_RENDER_THREADS_MIN` changed to `RENDER_THREADS_COUNT_MIN`
should `NUM_RENDER_THREADS_DEFAULT` change to `RENDER_THREADS_COUNT_DEFAULT`?

or perhaps `RENDER_THREADS_DEFAULT_COUNT` and `RENDER_THREADS_MIN_COUNT`

chunky/src/java/se/llbit/chunky/renderer/TileBasedRenderer.java
```diff
@@ -57,6 +58,9 @@ public void setPostRender(BooleanSupplier callback) {
     postRender = callback;
   }
 
+  protected Supplier<WorkerState> workerStateFactory = WorkerState::new;
+  protected Supplier<Ray> rayFactory = Ray::new;
```
ShirleyNekoDev, line 62: Any arguments against adding PluginAPI to these factory fields?

chunky/src/java/se/llbit/chunky/main/ChunkyOptions.java
```diff
@@ -65,11 +65,24 @@ public static ChunkyOptions getDefaults() {
...
   }
 
+  /**
+   * Returns the current state of the render options to by the renderer.
+   * It is not specified whether it is a copy or a view of the current options and
+   * does not guarantee to update on change.
+   * <p>Cast it to {@link ModifiableRenderOptions}, if you have to edit it.
+   * Edits are only safe on startup and might not propagate while a render is in progress.
+   *
+   * @return Options to use to start the render system.
```
leMaik, line 80: That sounds flaky to me.

chunky/src/java/se/llbit/chunky/renderer/DefaultRenderManager.java
```diff
@@ -323,6 +325,10 @@ public void run() {
     }
   }
 
+  protected void startRender(Renderer renderer) throws InterruptedException {
+    renderer.render(this);
+  }
```
leMaik, line 330: Is this for plugins?

chunky/src/java/se/llbit/chunky/renderer/ModifiableRenderOptions.java
```diff
@@ -0,0 +1,40 @@
+package se.llbit.chunky.renderer;
+
+public class ModifiableRenderOptions implements RenderOptions {
```
leMaik, line 3: See above. Can we improve this so that modifying the render options is safe and gets applied after the user confirms (or reset otherwise)?

chunky/src/java/se/llbit/chunky/renderer/TileBasedRenderer.java
leMaik: Why do we use the factory pattern here in the first place? Are there plugins that need this? If we add PluginApi, we have to maintain compatibility, which may not be what we want during refactoring.

ShirleyNekoDev: I had a renderer plugin idea which needed a way to add additional worker state / ray state --> answering your question: yes, plugins could need this.
But I would not mind it being internal (allowing for breaking changes) for now instead of PluginAPI.

leMaik: I have a PR with a per-worker pool of rays, which are re-used. Maybe I'll move the ray factory into the worker state eventually.

chunky/src/java/se/llbit/chunky/main/ChunkyOptions.java
ShirleyNekoDev: Same behaviour as before, but now documented. I wanted to make clear that this should typically only be set on startup (command line args) otherwise requiring a restart (of the render).

leMaik: Great! :clap:

chunky/src/java/se/llbit/chunky/renderer/DefaultRenderManager.java
ShirleyNekoDev: I believe I added this to be overridable by custom RenderManagers to add pre-start logic.

chunky/src/java/se/llbit/chunky/renderer/ModifiableRenderOptions.java
ShirleyNekoDev: Pretty sure that this would be the job of the UI logic or something in the render code which copies its state from this object. At the moment, the user can only set the `tile size` and `spp per pass` via CMD argument, right? And render thread count had some special handling somewhere which already required restarting the rendering or Chunky (?). Definitively a useful suggestion for a follow-up issue + PR :)

chunky/src/java/se/llbit/chunky/renderer/TileBasedRenderer.java
ShirleyNekoDev: Sounds good, I have no problem with updating plugin code for unstable/snapshot features ^^

chunky/src/java/se/llbit/chunky/renderer/DefaultRenderManager.java
ShirleyNekoDev: e.g. setting data in the new worker state / ray factories

PR #1573, opened 2023-04-03
Fix sun radius control on sun sampling.

ThatRedox:
Previously sun radius control was not properly updating in sun sampling. Moved sun radius initialization to `Sun.initSun()` to initialize the radius values properly.
Untitled1


leMaik:
Was this a snapshot only bug?

Peregrine05:
The `Sun size` control only exists in snapshot, so it would be snapshot-only.

PR #1578, opened 2023-04-15
Allow render threads to be changed without restarting

NotStirred:
I also removed the `CPU Load` option, as it wasn't accurate. This is a better replacement

Render threads changing from `1 -> 3 -> 6 -> 9 -> 12 -> 9 -> 6 -> 3 -> 1`
image
(The drops around 12 are from the brief period I had `1` typed in when typing `12`)


ThatRedox:
Duplicate of #1113.

PR #1579, opened 2023-04-15
Fix fluids shape on octree edge

NotStirred:
This is not the fastest way to do this, perhaps we could add in a single value cache for the most recently checked chunk?
Though as it's only on water in finalisation I decided against optimising for now

image


Peregrine05:
Related are #1247 and #1293.

leMaik:
Nice fix, we can optimize it later :rocket:

PR #1593, opened 2023-06-20
Allow empty chunks to be selected if their region exists

NotStirred:
On slower computers loading very large maps can take a long time, this PR allows chunks to be selected even if they're empty.

To clarify: `EmptyChunk`s only exist if the region they're within exists. `EmptyRegionChunk`s still cannot be selected.


leMaik:
@NotStirred Could you solve the conflicts? :pray: 

This PR allows loading empty chunks so that users don't have to zoom in, which would take a lot of time, I guess?

PR #1599, opened 2023-07-02
Add a ray pool and misc other optimizations (requires benchmarking)

leMaik:


jackjt8:
Performance appears to be roughly the same as all previous versions, maybe a few percent faster. Some minor weird initial run variation on SNAPSHOT.268 and this PR however repeat measurements appear to resolve the issue.
image


The is a major issue however. The `3D Render Workers` do not stop and keep using CPU thread time even after the SPP target has been reached. This only occurs when using this PR. I'm not sure how to track down what is causing this issue, I can currently only report it.

image


```
"3D Render Worker 0" #38 prio=5 os_prio=0 cpu=581187.50ms elapsed=623.87s tid=0x000001cf2c5113d0 nid=0x1f44 runnable  [0x0000005b1cefe000]
   java.lang.Thread.State: RUNNABLE
        at se.llbit.chunky.renderer.RenderWorkerPool$RenderWorker.run(RenderWorkerPool.java:51)

   Locked ownable synchronizers:
        - None
```

leMaik:
I removed some of the locking logic trying to improve the performance. I'll revert that.

PR #1601, opened 2023-07-05
Improve Refraction

### LINKED ISSUE #1515: IOR is incorrect

ThatRedox:
Fixes #1515
There was a check in the octree intersection code ignoring intersections with air. Removing this broke the water octree. I put a hack that coerces the ray's current block to be air if it isn't water when intersecting with the water octree. This seems to work for now but probably needs more testing.


ThatRedox:
My Test Scene: [Refraction Test.zip](https://github.com/chunky-dev/chunky/files/11953840/Refraction.Test.zip)
image

PR #1602, opened 2023-07-09
Fix decorated pot model

JustinTimeCuber:
MC screenshot (with custom texture to make rotation easier to see):

After changes:
image

Before:
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/model/DecoratedPotModel.java
+++ b/chunky/src/java/se/llbit/chunky/model/DecoratedPotModel.java
@@ -1,6 +1,5 @@
 package se.llbit.chunky.model;
 
-import se.llbit.chunky.entity.SporeBlossom;
 import se.llbit.chunky.resources.Texture;
 import se.llbit.log.Log;
 import se.llbit.math.Quad;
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/chunky/model/DecoratedPotModel.java
```diff
@@ -137,7 +165,7 @@ public DecoratedPotModel(String facing, String[] sherds) {
       // sherds[2] right crafting slot -> east
       getTextureForsherd(sherds[2]),
       Texture.decoratedPotBase, // top
-      Texture.decoratedPotBase  // bottom
+      Texture.decoratedPotSide  // bottom
```
ShirleyNekoDev, line 168: is the "// bottom" comment still correct?

JustinTimeCuber: Yes, for some reason Mojang is sampling a seemingly arbitrary pixel from the side texture to determine the bottom:
image

JustinTimeCuber: wait... I can't reproduce this behavior in a fresh unmodded instance, wtf is going on

JustinTimeCuber: sodium bug (!?)

JustinTimeCuber:
Okay, seems like an unusual [bug](https://github.com/CaffeineMC/sodium-fabric/issues/1872) in Sodium causing it to do that. I've changed it to match Vanilla behavior.

image

leMaik (approved):
Thank you, i know this is pretty laborious to get right. :)

PR #1603, opened 2023-07-09
Fix Javadoc for Quad constructors

JustinTimeCuber:
When I was working on the Decorated Pot model (#1602), I noticed that the Javadocs for the Quad constructors seem wrong, since as far as I can tell, the first three arguments are the bottom left, bottom right, and top left coordinates of the quad, respectively. Hopefully this helps explain - these are `Quad`s defined in `QuadModel`:

```
public static final Quad FULL_BLOCK_NORTH_SIDE = new Quad(
    new Vector3(1, 0, 0),
    new Vector3(0, 0, 0),
    new Vector3(1, 1, 0),
    new Vector4(0, 1, 0, 1));
```
image

```
public static final Quad FULL_BLOCK_EAST_SIDE = new Quad(
    new Vector3(1, 0, 1),
    new Vector3(1, 0, 0),
    new Vector3(1, 1, 1),
    new Vector4(0, 1, 0, 1));
```
image

```
public static final Quad FULL_BLOCK_TOP_SIDE = new Quad(
    new Vector3(1, 1, 0),
    new Vector3(0, 1, 0),
    new Vector3(1, 1, 1),
    new Vector4(1, 0, 1, 0));
```
image

I confirmed that the other sides work the same way although I don't think more pictures are needed.

While I'm pretty sure something is wrong with the Javadocs, it is also possible that I'm wrong in which case I'd like to know how I am misunderstanding this.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/model/QuadModel.java
+++ b/chunky/src/java/se/llbit/chunky/model/QuadModel.java
@@ -16,6 +16,7 @@ import java.util.Random;
 @PluginApi
 public abstract class QuadModel implements BlockModel {
 
+  // For some visualizations, see this PR: https://github.com/chunky-dev/chunky/pull/1603
   public static final Quad FULL_BLOCK_NORTH_SIDE = new Quad(
     new Vector3(1, 0, 0),
     new Vector3(0, 0, 0),
--- a/chunky/src/java/se/llbit/math/Quad.java
+++ b/chunky/src/java/se/llbit/math/Quad.java
@@ -100,6 +100,8 @@ public class Quad {
 
   /**
    * Create a new single sided quad
+   * <br>
+   * For some visualizations, see this PR: <a href="https://github.com/chunky-dev/chunky/pull/1603">#1603</a>
    *
    * @param v0 Bottom left vector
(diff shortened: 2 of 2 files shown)


leMaik:
Great, this would have helped a lot back when i did all the 1.14-1.15 stuff. Could you add a link to this PR, ie. the drawings, to the javadoc?

leMaik (approved):

JustinTimeCuber:
> Great, this would have helped a lot back when i did all the 1.14-1.15 stuff. Could you add a link to this PR, ie. the drawings, to the javadoc?

Good idea, done. Also added a similar comment in QuadModel since it is relevant.

leMaik:
Regarding the top/bottom change: The y coordinate of textures is mirrored everywhere in Chunky (ie. 0 is at the bottom of a texture). Does this explain the confusion or is the vector actually flipped?

JustinTimeCuber:
I think the connection there is that since the texture starts from the bottom left, that's why the first vector is the bottom left coordinate. And then the second vector is parallel to the "u-axis" and the third is parallel to the "v-axis" if that makes sense.

PR #1606, opened 2023-07-19
Fix dimension bugs in the map view

NotStirred:
Previously players would always be rendered on the map view regardless of dimension.
Additionally, the map view wouldn't update on switching dimension and required the camera to move first


leMaik:
Nice! Does this need to be backported to 2.4?

leMaik (approved):

NotStirred:
Very likely, yes

Issue #1563, opened 2023-03-22
Minecraft 23w12a Snapshot Features


ShirleyNekoDev:
This is also a follow up issue for snapshot changes after #1550

- [x] new decorated pot spout model
- [x] new pottery shards
- [x] calibrated sculk sensor block
- [x] sus gravel
- [x] pitcher plant
- [x] pitcher crop
- [x] sniffer egg
- [x] text on both sides of a sign
- [x] flowerbed model changes

Possible new entities:
- [ ] armor trims on armor stand entities
- [ ] display entity

leMaik:
The decorated pot model was updated in #1602 :tada:

PR #1608, opened 2023-07-28
Add the pitcher crop and calibrated sculk sensor.

leMaik:
Related to #1563

### DIFF
--- a/chunky/src/java/se/llbit/chunky/model/PitcherCropTopModel.java
+++ b/chunky/src/java/se/llbit/chunky/model/PitcherCropTopModel.java
@@ -6,21 +6,23 @@ import se.llbit.math.Vector3;
 import se.llbit.math.Vector4;
 
 public class PitcherCropTopModel extends QuadModel {
-  private static final Texture pitcher_top = Texture.pitcherCropTop;
-  private static final Texture pitcher_side = Texture.pitcherCropSide;
-  private static final Texture pitcher_bottom = Texture.pitcherCropBottom;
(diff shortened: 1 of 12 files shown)


ShirleyNekoDev (approved):
Looks good to me 👍

chunky/src/java/se/llbit/chunky/model/PitcherCropTopModel.java
```diff
@@ -0,0 +1,168 @@
...
+  private static final Texture pitcher_bottom = Texture.pitcherCropBottom;
+  private static final Texture stage_2 = Texture.pitcherCropBottomStage2;
+
+  private static final Texture[][] textures = new Texture[][]{
+    null, null,
+    new Texture[]{
+      stage_2, stage_2, stage_2, stage_2, pitcher_top, pitcher_bottom, pitcher_side, pitcher_side, pitcher_side, pitcher_side
+    },
+    new Texture[]{
+      Texture.pitcherCropTopStage3, Texture.pitcherCropTopStage3, Texture.pitcherCropTopStage3, Texture.pitcherCropTopStage3
```
ShirleyNekoDev, line 20: Any reason why stage_2 is an instance variable above, but not these stage textures?

leMaik: First, this model didn't have a model for age=2. But then I saw that the bottom model of age=2 is higher than one block, so I added the same model to the top model (translated by 1 block to the bottom). Same trick as for pitcher plant.

I'll make these instance variables, too.

PR #1611, opened 2023-08-02
Split World and Dimension into their own classes

NotStirred:
Split the concepts of Dimension and World into their own class, 
World handles per-world data, and loading Dimensions
Dimension handles regions, chunks, loading chunks, and any per-dimension data.

There is no API for detecting the format of a world or dimensions within it, hopefully that can come in the next PR

This PR properly loads a world with one or more vanilla dimensions, and one or more CubicChunks dimensions (previously the entire world was either vanilla or CubicChunks)


NotStirred:
This PR has some reasonable change of breaking plugin api actually
many of the listeners `ChunkDeletionListener`, `ChunkTopographyListener`, `ChunkUpdateListener` are now registered on `Dimension` instead of `World`. I'll add a deprecated method to `World` that just redirects to the current loaded dimension

chunky/src/java/se/llbit/chunky/world/Dimension.java
```diff
@@ -0,0 +1,312 @@
...
+   */
+  protected Dimension(World world, int dimensionId, File dimensionDirectory, Set<PlayerEntityData> playerEntities, long timestamp) {
+    this.world = world;
+    this.dimensionId = dimensionId;
+    this.dimensionDirectory = dimensionDirectory;
+    this.playerEntities = playerEntities;
+    this.timestamp = timestamp;
+  }
+
+  public int dimensionId() {
```
leMaik, line 59:
```suggestion
  public int getDimensionId() {
```

leMaik: Also, is a `int` really the API we want to have here? Is that compatible with Bedrock?

chunky/src/java/se/llbit/chunky/world/Dimension.java
```diff
@@ -0,0 +1,312 @@
...
+  public synchronized File getRegionDirectory() {
+    return new File(getDimensionDirectory(), "region");
+  }
+
+  /**
+   * Get the current player position as an optional vector.
+   *
+   * <p>The result is empty if this is not a single player world.
+   */
+  public synchronized Optional<Vector3> playerPos() {
```
leMaik, line 189:
```suggestion
  public synchronized Optional<Vector3> getPlayerPosition() {
```

chunky/src/java/se/llbit/chunky/world/Dimension.java
```diff
@@ -0,0 +1,312 @@
...
+      return Optional.of(new Vector3(pos.x, pos.y, pos.z));
+    } else {
+      return Optional.empty();
+    }
+  }
+
+  /**
+   * @return The chunk heightmap
+   */
+  public Heightmap heightmap() {
```
leMaik, line 201:
```suggestion
  public Heightmap getHeightmap() {
```

chunky/src/java/se/llbit/chunky/world/Dimension.java
```diff
@@ -0,0 +1,312 @@
...
+   *
+   * @param chunk The chunk
+   */
+  public void chunkTopographyUpdated(Chunk chunk) {
+    for (ChunkTopographyListener listener : chunkTopographyListeners) {
+      listener.chunksTopographyUpdated(chunk);
+    }
+  }
+
+  public Optional<Vector3i> spawnPos() {
```
leMaik, line 269:
```suggestion
  public Optional<Vector3i> getSpawnPosition() {
```

chunky/src/java/se/llbit/chunky/world/Dimension.java
```diff
@@ -0,0 +1,312 @@
...
+    return Optional.ofNullable(this.spawnPos);
+  }
+
+  public void spawnPos(@Nullable Vector3i spawnPos) {
```
leMaik, line 273:
```suggestion
  public void setSpawnPosition(@Nullable Vector3i spawnPos) {
```

chunky/src/java/se/llbit/chunky/world/Dimension.java
```diff
@@ -0,0 +1,312 @@
...
+
+  public Date getLastModified() {
+    return new Date(this.dimensionDirectory.lastModified());
+  }
+
+  /**
+   * Load entities from world the file.
+   * This is usually the single player entity in a local save.
+   */
+  public synchronized Collection<PlayerEntity> playerEntities() {
```
leMaik, line 295:
```suggestion
  public synchronized Collection<PlayerEntity> getPlayerEntities() {
```

chunky/src/java/se/llbit/chunky/world/World.java
```diff
@@ -131,12 +96,61 @@ public enum LoggedWarnings {
...
+    currentDimension.reloadPlayerData();
+  }
+
+  /**
+   * Add a chunk deletion listener.
+   *
+   * @deprecated Remove in 2.7.
+   * Replaced by {@link World#currentDimension()} -> {@link Dimension#addChunkDeletionListener}
+   */
+  @Deprecated
```
leMaik, line 110: It didn't have `@PluginApi`, so you could as well just delete these methods.

If you don't, it's probably fine to remove them in 2.6. Are there any plugins that actually use these APIs?

leMaik (approved):
Mostly nit-picking (remove abbreviations like pos, change getters/setters to Java convention).

Everything else looks good, I like the split of Worlds and Dimensions. :clap:

chunky/src/java/se/llbit/chunky/world/World.java
NotStirred: Beyond my own, no I don't think so. I'll remove it :+1:

chunky/src/java/se/llbit/chunky/world/Dimension.java
NotStirred: No, `int` is definitely not the API we want, but changing this requires a little more fiddling with internals of other classes, this will be the next PR

NotStirred:
(force push to drop redundant plugin api commit)

PR #1614, opened 2023-08-09
Replace SimplexNoise license link with CC0 instead of Unlicense

ShirleyNekoDev:
The linked source code did not state Unlicense, but is released as public domain software. While I do not know the legal aspects in detail, I'll assume that CC0 is a better description of the library author's intents. 
also see https://chrismorgan.info/blog/unlicense/

### DIFF
--- a/chunky/src/java/se/llbit/chunky/ui/controller/CreditsController.java
+++ b/chunky/src/java/se/llbit/chunky/ui/controller/CreditsController.java
@@ -171,7 +171,7 @@ public class CreditsController implements Initializable {
     simplexnoise.setBorder(Border.EMPTY);
     simplexnoise.setOnAction(e -> launchAndReset(simplexnoise, "https://github.com/keijiro/sketches2016/blob/master/Simplex2/SimplexNoise.java"));
     simplexnoiseLicense.setBorder(Border.EMPTY);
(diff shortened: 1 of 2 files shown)


jackjt8:
Looking at the project readme the code was taken from clearly states the license is Unlicense. 

> Code: Unlicense - You can use it for whatever you want.

https://github.com/keijiro/sketches2016/tree/master

ShirleyNekoDev:
Oh, you are right. I took the licensing statement information from the file itself [SimplexNoise](https://github.com/keijiro/sketches2016/blob/master/Simplex2/SimplexNoise.java) (which is the only part we use), isolated from the repository. Not sure if this would count as separately licensed.

jackjt8:
After reviewing the original code there is no mention of GNU. That appears to be an addition made by someone else. The original code is public domain. It's not got a well known waiver or licence, but it's still public domain. As our implementation is based off a derivative work we should maintain the license presented (Unlicense).

Otherwise.. we could always readapt things from the original public domain code. That would solve our license concerns.

leMaik:
We could as well not mention the license name and just link to the code we adapted.

leMaik:
The Java port we link to is licensed under the _unlicense_. It is a port of public domain code.

PR #1428, opened 2022-08-09
Add freely adjustable layered fog

almrausch:
Completely rewrote the code for the fog so that fog can e.g. concentrate at the bottom, allowing for a wider range of effects. Fog density is modelled as a function of y. It is a sum of bell shaped sigmoid functions. Each fog layer has a center y coordinate, a density and a breadth value. Scattering works only for one layer so far. The scene JSON structure has been altered so that all fog data is stored in one JsonObject. The previous fog behavior is still supported and old scenes will be saved in the new format.

Import the following settings to test the new features:
`{"fog":{"mode":"LAYERED","layers":[{"y":62,"breadth":10,"density":1}]}}`
`{"fog":{"mode":"UNIFORM","uniformDensity":0.05}}`
`{"fog":{"mode":"NONE"}}`

A GUI implementation is still missing. So far the GUI still works the same except that the mode must be set to UNIFORM via JSON. I'd like to leave the GUI part to someone else.

PR #1618, opened 2023-08-21
Add a GUI for layered fog

### LINKED ISSUE #1479: Add UI for layered fog
#1428 needs a UI, and since this would be fairly complex it's being split from the functionality and worked on separately. Here's a potential mockup:
notlabelled
Labelled:
labelled

There has been [some discussion](https://discord.com/channels/541221265512464394/1032297295795982396/1032297839776251914) on the discord about this UI, and this issue can serve as a tracker for work related to this.
Hopefully @ShirleyNekoDev can work some magic on this, and open a PR when we're closer to a working prototype.

Peregrine05:
Closes #1479.
Instead of creating a more complicated GUI as shown in the issue, I decided to create a simpler one.

**New:**
image
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Fog.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Fog.java
@@ -1,7 +1,8 @@
 package se.llbit.chunky.renderer.scene;
 
-import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.Random;
+import java.util.stream.Collectors;
 
 import se.llbit.chunky.PersistentSettings;
 import se.llbit.json.JsonArray;
@@ -19,7 +20,7 @@ public final class Fog implements JsonSerializable {
   protected boolean fastFog = true;
   protected double uniformDensity = Scene.DEFAULT_FOG_DENSITY;
   protected double skyFogDensity = 1;
-  protected FogLayer[] layers = new FogLayer[0];
+  protected ArrayList<FogLayer> layers = new ArrayList<>(0);
   protected Vector3 fogColor = new Vector3(PersistentSettings.getFogColorRed(), PersistentSettings.getFogColorGreen(), PersistentSettings.getFogColorBlue());
 
   private static final double EXTINCTION_FACTOR = 0.04;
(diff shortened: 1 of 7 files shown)


megamasterbloc:
maybe change layer breath to a more meaningful term like layer dispersion or layer spreading

Peregrine05:
It is called `breadth` internally, and it is an accurate term in my opinion.

chunky/src/java/se/llbit/chunky/renderer/scene/Fog.java
```diff
@@ -48,6 +54,47 @@ public boolean fastFog() {
...
+    int currentIndex = 0;
+    int newIndex = 0;
+    for (FogLayer layer : layers) {
+      if (!(currentIndex == index)) {
+        layers2[newIndex] = layer;
+        newIndex++;
+      }
+      currentIndex++;
+    }
+    layers = Arrays.copyOf(layers2, layers2.length);
```
leMaik, line 79: Wouldn't it be way easier to use a list?

Peregrine05: I'll attempt the conversion. At some point, I did store it as an ArrayList, which was much easier, but I was tricked into thinking that it was causing an issue, where the issue was actually due to the minimum value of *0.01* for fog breadth, which makes fog rendering very unstable and likely to crash or hang forever. By then, I had already changed it back to Array, though.

Peregrine05:
Should I change the name of the `Layer breadth` control, or leave it as it is?

jackjt8:
> Should I change the name of the `Layer breadth` control, or leave it as it is?

The definition of Breadth is "the the distance from one side to another". I haven't looked at the code as to exactly what it's used for but I'm pretty sure Breadth controls the vertical spread of the fog layer. Which.. technically would mean the wrong term is use?

Peregrine05:
I wonder if `Veritcal spread` would be a suitable replacement.

leMaik:
How about `layer height`?

Peregrine05:
> How about `layer height`?

That would be too easily confused with `Layer altitude`, I believe.

Peregrine05:
I changed it to `Layer thickness`, which should be easily understandable.

leMaik (approved):

leMaik:
@Peregrine05 Somehow, picking the fog color of any layer changes the color of all layers.

Peregrine05:
> @Peregrine05 Somehow, picking the fog color of any layer changes the color of all layers.

@leMaik This is actually a limitation of the layered fog, and is mentioned in #1428.

leMaik:
Indeed, it wasn't clear how to blend the layers iirc. Can we make it obvious in the UI that the color affects all layers? Or is it obvious and I'm just blind?

Peregrine05:
It does not seem obvious to me. Another hint could be useful.

megamasterbloc:
while it's a limitation of chunky I think that it's perfectly reasonable to only have one color for all fog layers. layered fog is only meant to change the distribution of fog and not it's properties

Peregrine05:
I have added a new note / hint to the fog GUI.

PR #1620, opened 2023-08-31
Workaround for some potential drawbacks of ray branching

JustinTimeCuber:
Don't branch on the first few SPP to avoid some potential performance-related issues (such as pointed out [here](https://discord.com/channels/541221265512464394/1144538849582792847/1144538849582792847) on Discord) based on the square root of the branch count (arbitrary, but seems to work fine)

Examples:
If BC=5, sequence will be 1, 1, 1, 2, 5, 5, 5, ...
If BC=10, sequence will be 1, 1, 1, 1, 6, 10, 10, 10, ...
If BC=20, sequence will be 1, 1, 1, 1, 1, 15, 20, 20, 20, ...

Also, this sets the default branch count to 10 instead of 5 (since there are now fewer issues with that.)

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -1920,11 +1920,19 @@ public class Scene implements JsonSerializable, Refreshable {
   }
 
   /**
-   * @param modified Whether to alter the branch count based on current spp
-   * @return The branch count
+   * @return The "nominal" value of the branch count.
    */
-  public int getBranchCount(boolean modified) {
-    if(modified && spp < branchCount) {
+  public int getBranchCount() {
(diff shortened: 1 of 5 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1920,10 +1920,19 @@ public void setTargetSpp(int value) {
   }
 
   /**
+   * @param modified Whether to alter the branch count based on current spp
```
leMaik, line 1923: This boolean changes the behavior of the whole method. Can we create a new method instead? `getCurrentBranchCount` maybe, with the same explanation?

JustinTimeCuber: Done.

PR #1622, opened 2023-09-05
Add RegionPosition class, replace all usages of ChunkPosition as a region with it.

NotStirred:
This should have zero impact on performance, but makes things much clearer and adds a nice layer of type safety on top.

I think the name of `ChunkPosition#asRegionPosition` is worth discussing. I didn't want to change the return type of the existing `getRegionPosition` in case any plugins use it as I'm unsure. Maybe none do?

### DIFF
--- a/chunky/src/java/se/llbit/chunky/world/ChunkPosition.java
+++ b/chunky/src/java/se/llbit/chunky/world/ChunkPosition.java
@@ -51,21 +51,11 @@ public class ChunkPosition {
     return String.format("r.%d.%d.mca", x, z);
   }
 
-  public RegionPosition asRegionPosition() {
-    return new RegionPosition(x >> 5, z >> 5);
-  }
-
   /**
    * @return The region position of this chunk position
-   *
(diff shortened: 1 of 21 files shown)


chunky/src/java/se/llbit/chunky/world/ChunkPosition.java
```diff
@@ -43,13 +43,22 @@ public long getLong() {
...
+  public RegionPosition asRegionPosition() {
+    return new RegionPosition(x >> 5, z >> 5);
+  }
+
   /**
    * @return The region position of this chunk position
+   *
+   * @deprecated Use {@link ChunkPosition#asRegionPosition()}. Remove in 2.6
    */
   public ChunkPosition getRegionPosition() {
```
leMaik, line 63: After much discussion on Discord, I guess that a breaking change here is fine for 2.5

NotStirred:
Rebased onto master (auto rebase was failing horribly)

PR #1623, opened 2023-09-09
Remove unused amethyst "lit" tag

JustinTimeCuber:
The `lit` tag was removed in a later 1.17 snapshot due to being unused; all amethyst clusters should glow. This change also makes a separate `AmethystCluster extends SpriteBlock` class redundant, so I removed it.


leMaik:
Thank you :+1:

PR #1624, opened 2023-09-09
Improve octree, bvh, renderer and output mode dropdowns.

leMaik:


chunky/src/java/se/llbit/chunky/ui/RegisterableCellAdapter.java
```diff
@@ -0,0 +1,26 @@
...
+    @Override
+    public String getLabel(Registerable item) {
+        return item.getName();
+    }
+
+    @Override
+    public Tooltip getTooltip(Registerable item) {
+        String description = item.getDescription();
+        if (description != null && !description.isEmpty()) {
+            return new Tooltip(item.getDescription());
```
leMaik, line 22:
```suggestion
            return new Tooltip(description);
```

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -836,8 +836,8 @@ public OctreeImplementation getImplementation() {
     return implementation;
   }
 
-  public static void addImplementationFactory(String name, ImplementationFactory factory) {
-    factories.put(name, factory);
+  public static void addImplementationFactory(ImplementationFactory factory) {
```
leMaik, line 839: This breaking change is bad and we should find a way to work around this (maybe a backward-compatible `ImplementationFactory` and a new `RegisterableImplementationFactory`).

chunky/src/java/se/llbit/chunky/renderer/RenderManager.java
```diff
@@ -39,12 +39,12 @@ public interface RenderManager {
   /**
    * Get all available {@code Renderer}s.
    */
-  Collection<? extends Registerable> getRenderers();
+  Collection<Renderer> getRenderers();
```
ThatRedox, line 42: This would make it impossible for plugins to implement `RenderManager` and have their own type of swappable renderer that is incompatible with Chunky's renderers. (Which kinda defeats the point of changing the RenderManager implementation).

leMaik: Oh… Okay, so a doesn't have to be a `Renderer`?

ThatRedox: Yeah, `Renderer` is just the interface to work with Chunky's `DefaultRenderManager`.

Issue #789, opened 2021-01-21
21w03a


leMaik:
New year, new snapshots! :partying_face: 

- [ ] Glow lichen
- [ ] Glow item frame

Due to the new 1.17 world format (entities are not stored in the same files as the blocks anymore), we can't load entities currently. This breaks loading painting and prevents us from loading (glow) item frames from 1.17 worlds.

PR #1626, opened 2023-09-10
Add dyed signs.

### LINKED ISSUE #1625: Dyed text on signs
Minecraft allows you to use dye to color the text on signs. Given that Chunky supports color codes already, this should be fairly straightforward to implement.
Example:

leMaik:
This PR adds dyed signs. Dyes can be different on the front and back side of a sign.

Note that the 16 dye colors are different from the usual sign text colors (except for black). This means that we can't use the existing 4-bit palette for the dye color. In order to not have to add more bits to the palette, I made it so that if the palette contains a `1` but the text mask is `0`, the sign texture treats the pixel as _dye color_, which requires only 32 or 64 bits (ie. an enum value) per sign texture.

I extracted the colors via screenshots from dyed signs. They may be incorrect.

Closes #1625 and related to #789.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/entity/HangingSignEntity.java
+++ b/chunky/src/java/se/llbit/chunky/entity/HangingSignEntity.java
@@ -253,22 +253,26 @@ public class HangingSignEntity extends Entity {
   private final boolean attached;
   private final SignTexture frontTexture;
   private final SignTexture backTexture;
+  private final SignEntity.Color frontDye;
+  private final SignEntity.Color backDye;
   private final Texture texture;
   private final String material;
 
   public HangingSignEntity(Vector3 position, CompoundTag entityTag, int rotation, boolean attached, String material) {
-    this(position, SignEntity.getFrontTextLines(entityTag), SignEntity.getBackTextLines(entityTag), rotation, attached, material);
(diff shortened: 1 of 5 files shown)


ThatRedox:
These should be the text colors:
| Color           | ARGB       |
|-----------------|------------|
| WHITE           | 0xFFFFFFFF |
| ORANGE          | 0xFFFF681F |
| MAGENTA         | 0xFFFF00FF |
| LIGHT_BLUE      | 0xFF9AC0CD |
| YELLOW          | 0xFFFFFF00 |
| LIME            | 0xFFBFFF00 |
| PINK            | 0xFFFF69B4 |
| GRAY            | 0xFF808080 |
| LIGHT_GRAY      | 0xFFD3D3D3 |
| CYAN            | 0xFF00FFFF |
| PURPLE          | 0xFFA020F0 |
| BLUE            | 0xFF0000FF |
| BROWN           | 0xFF8B4513 |
| GREEN           | 0xFF00FF00 |
| RED             | 0xFFFF0000 |
| BLACK           | 0xFF000000 |

leMaik:
The text of white dyed signs is light gray in game. Do we need to multiply all of them by some factor if the sign doesn't glow?

ThatRedox:
Yes, it looks like each channel is multiplied by `0.4` if the sign isn't glowing.

leMaik:
Note to self: The outline on glowing signs is 0.4 * dye color.

PR #1627, opened 2023-09-13
Color-based, per-material emittance mapping

JustinTimeCuber:
NOTE: this PR depends on/includes the changes from #1621, so it is currently marked as a draft.
One useful feature Chunky is currently missing is the ability to differentiate between the parts of blocks that should emit light and the parts that shouldn't. Here is an example with furnaces:
image
As can be seen, the entirety of the blocks are emitting light despite the fact that physically, only the "flame" portions should be providing light. 
Ideally this could be done with PBR maps, but besides the fact that we don't have those yet, that basically requires either us to provide them for every emitter (potentially a lot of work) or users to provide them (not really an ideal solution in my opinion). This PR provides 3 methods for determining how much light should be emitted from each pixel as a function of pixel color rather than an emittance map texture. One of these is `REFERENCE_COLORS`, which allows for specific color range(s) to be specified which should emit light while all other colors will not do so. In this mode, the strength of the emitted light is also dependent on the HSV Value of the pixel color raised to some exponent (default: 1.5). This mode is well-suited to furnaces, as well as some other blocks like lanterns, torches, glow berries, and end rods. There is currently no GUI for adjusting the reference colors, but they can be specified in JSON format and defaults are included for the aforementioned emitters.
image
For certain other emitters like magma blocks, sea lanterns, and redstone lamps, the darker pixels should emit _less_ light, but not _none_. In this case, a simpler method (`BRIGHTEST_CHANNEL`) is used, which simply takes the HSV value raised to the same exponent and uses that as the relative emittance of that pixel:
image
There is also one more method included called (`INDEPENDENT_CHANNELS`) which raises each individual RGB component to the provided exponent. This isn't the default because it tends to increase the saturation of emitted light at higher powers making it a bit less realistic, but it is still included as an option for more artistic freedom.
image

The current version of this PR doesn't include a GUI option for setting `REFERENCE_COLORS` which can only be customized by editing the scene JSON, but everything else is customizable through the GUI. I'll look into adding a GUI option (let me know if this is needed before merging) based on the layered fog GUI in #1618.


leMaik:
We can add a new GUI in a future PR.

leMaik:
Rebased against latest `master`. Mapping for copper bulbs would be great. :bulb:

leMaik:
@JustinTimeCuber Given that #1621 is merged now, is this one ready for merge?

JustinTimeCuber:
I think so, as long as there isn't another new light source I haven't heard about

leMaik (approved):

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
```diff
@@ -507,11 +532,12 @@ private static void sampleEmitterFace(Scene scene, Ray ray, Grid.EmitterPosition
...
         e *= pos.block.surfaceArea(face);
-        e *= emitterRay.getCurrentMaterial().emittance;
-        e *= scene.emitterIntensity;
         e *= scaler;
+        Vector3 emittance = new Vector3();
+        doEmittanceMapping(emittance, emitterRay.color, scene, emitterRay.getCurrentMaterial());
+        emittance.scale(e);
 
-        result.scaleAdd(e, emitterRay.color);
+        result.add(new Vector4(emittance, 0));
```
leMaik, line 540: Can we do this without allocating `Vector3` and `Vector4`?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -3183,6 +3240,27 @@ public void setEmittance(String materialName, float value) {
...
+  }
+
+  /**
+   * Modifies the emittance property for the given material.
+   */
+  public void setEmitterMappingTypeOverride(String materialName, EmitterMappingType value) {
+    JsonObject material = materials.getOrDefault(materialName, new JsonObject()).object();
+    material.set("emitterMappingType", Json.of(value.toString()));
+    materials.put(materialName, material);
+    System.out.println(materialName + ", " + value);
```
leMaik, line 3260: Remove debug output

```suggestion
```

chunky/src/java/se/llbit/chunky/ui/render/tabs/LightingTab.java
```diff
@@ -101,7 +106,21 @@ public LightingTab() throws IOException {
...
+    boolean showEmitterSettings = scene != null && scene.getEmittersEnabled();
+    emitterSettings.setVisible(showEmitterSettings);
+    emitterSettings.setManaged(showEmitterSettings);
+
+//    fancierTranslucency.selectedProperty()
+//      .addListener((observable, oldValue, newValue) -> {
+//        scene.setFancierTranslucency(newValue);
+//        transmissivityCap.setVisible(newValue);
+//        transmissivityCap.setManaged(newValue);
+//      });
```
leMaik, line 123: This seems unrelated?

chunky/src/java/se/llbit/chunky/world/Material.java
```diff
@@ -60,6 +66,22 @@ public abstract class Material {
...
+  /**
+   * Overrides the global emitter mapping type unless set to NONE.
+   */
+  public EmitterMappingType emitterMappingType = EmitterMappingType.NONE;
+
+  /**
+   * (x, y, z): The color to use for the REFERENCE_COLORS emitter mapping type.
+   * w: The range surrounding the specified color to apply full brightness.
+   */
+  public ArrayList<Vector4> emitterMappingReferenceColors = new ArrayList<>();
```
leMaik, line 83: We allocate lots of materials (one per block type, one per triangle in the bvh). Will this become an issue?

leMaik (changes requested):
Looks good, there's some debug code left and we should make sure that we don't increase memory usage too much (especially when color mapping isn't used).

chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
JustinTimeCuber: I think the `Vector3 emiitance` is definitely useful because otherwise we'd need a `doEmittanceMappingR` etc. but the `Vector4` could be removed

chunky/src/java/se/llbit/chunky/ui/render/tabs/LightingTab.java
JustinTimeCuber: Huh I wonder how that got there, I'll remove it

chunky/src/java/se/llbit/chunky/world/Material.java
JustinTimeCuber: Maybe I could set it to `null` and only initialize when needed?

leMaik: Sounds good

leMaik: Also, maybe set it to `null` again when it's not needed anymore

PR #1634, opened 2023-09-26
Use a ChoiceBox for water shader

Peregrine05:
Removed the `Still water` and `Procedural water` checkboxes and replaced them with a `Water shader` dropdown that contins options: `SIMPLEX` (default), `LEGACY`, and `STILL`. Also updated the persistent settings to support the new GUI. Scenes from older versions should have their water settings preserved when loaded.

**New:**
image
image
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/WaterShadingStrategy.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/WaterShadingStrategy.java
@@ -1,0 +1,48 @@
+/* Copyright (c) 2023 Chunky Contributors
+ *
+ * This file is part of Chunky.
+ *
+ * Chunky is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Chunky is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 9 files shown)


chunky/src/java/se/llbit/chunky/renderer/WaterShadingStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with Chunky.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package se.llbit.chunky.renderer;
+
+import se.llbit.util.Registerable;
+
+public enum WaterShadingStrategy implements Registerable {
+    SIMPLEX("Simplex", "Uses configurable noise to shade the water, which prevents tiling at great distances."),
+    LEGACY("Legacy", "Uses a tiled normal map to shade the water"),
```
leMaik, line 23: Why not call it `builtinNormalMap` then?

Peregrine05: What is the determining factor of what case the ID should be?

Peregrine05: I notice that some may be camel case and others may be all caps.

leMaik: It's not about the case (didn't notice that) but about whether we consider the hard-coded normal map to be legacy.

leMaik (approved):
Looks good. I renamed `LEGACY` and improved the tooltip. Let me know what you think!

Peregrine05:
The changes look good to me. :+1:

PR #1637, opened 2023-09-27
Improve biome blending in 3D

aTom3333:
This PR improves performance of doing biome blending during chunk loading. 
The combined time of step 3 and 4 (loading chunks and finalization ; biome blending occurs during step 4) has roughly been halved based on my rough timings. (Sorry, no precise timings provided, I've been winging it a bit). 
The improved performance and the improved asymptotic complexity of blur make it reasonable to do blur bigger than 3*3*3 so that's now possible and exposed in the UI. 
As an added bonus that wasn't really intentional, memory usage has decreased when using trivial 3D biome structure (by virtue of not storing biome color for every block, more on that later).

The improved performances are achieved for 3 reasons:
 - When possible a 2D blur is used instead of a 3D one. To know when using a 2D blur can be used (will give the same result as the 3D one), I keep track of vertical biome transition during chunk loading, ie when the biome at a block is different than the biome at the block just under. When a transition occurs, for every y level that is close to transition (closer than the blur radius), a 3D blur is needed as the biomes are not vertically uniform. Basically, for each y at which a (vertical) biome transition occurs, there is an interval of y value for which a 3D blur must be used. But for every y level outside of all of those intervals, a 2D blur is enough. Even better, there is usually a lot of consecutive y levels with (vertically) uniform biomes, this means the 2D blur can be computed for only one of those levels and copied for the other levels. In practice I don't track biome transition for each (x, z) column but for each chunk, this means that if there is a biome transition in a single column of the chunk, the 3D blur will be used for the entire chunk (for the y levels around the transition). And even more, because neighboring chunks affect biome blending, a single biome transition in a column will affect the whole chunk of the block and the neighboring chunks. To be effective this optimization relies on the fact that there isn't a lot of vertical biome transition in a minecraft world (which seems to be the case in my very limited testing). (this optimization does nothing to help performance when using 2D biome only)
 - The blur computations, both 2D and 3D, have been changed to use a summed-area table to reduce the complexity, especially with bigger blur radius. (while writing this message I realized that in asymptotic sense, it doesn't actually improve the complexity the way I've done it. It's better then naive blur but maybe an separated blur is even better but that would be a separated PR). This optimization benefits the 2D biome only case as well.
 - Only compute the biome colors when needed. That's a very rough way to do it (it could be done more finaly later I guess), during loading chunk I keep track of the min and max y value for which at least one block of the chunk needs biome color. When comes times to compute biome color, I don't compute any color outside of this interval. This simple test can lead to big wins on some chunks (like when there is no water underground, the biome colors will only be computed around surface level). (doesn't help in the 2D biome case). This is the change that cause the memory usage reduction, because in some cases a lot less biomes colors are stored.

The setting for enabling/disabling biome blending has been replace for a setting to choose the biome blending radius (and a radius of 0 means biome blending is disabled)

Sorry no screenshot of bigger radius, I'm too lazy and my test scene is stupid looking anyway

### DIFF
--- a/chunky/src/java/se/llbit/chunky/block/legacy/blocks/LegacyFlowerPot.java
+++ b/chunky/src/java/se/llbit/chunky/block/legacy/blocks/LegacyFlowerPot.java
@@ -85,7 +85,7 @@ public class LegacyFlowerPot extends UnfinalizedLegacyBlock {
   }
 
   @Override
-  public boolean useBiomeTint() {
+  public boolean isBiomeDependant() {
     return true; // (in reality it is only used for fern)
   }
 }
--- a/chunky/src/java/se/llbit/chunky/block/legacy/blocks/LegacyFlowerPot.java
+++ b/chunky/src/java/se/llbit/chunky/block/legacy/blocks/LegacyFlowerPot.java
@@ -83,4 +83,9 @@ public class LegacyFlowerPot extends UnfinalizedLegacyBlock {
 
     return tag; // keep empty
   }
+
+  @Override
+  public boolean useBiomeTint() {
+    return true; // (in reality it is only used for fern)
+  }
 }
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -930,9 +930,9 @@ public class Scene implements JsonSerializable, Refreshable {
         int wz0 = cp.z * 16;
         BiomeData biomeData = chunkData.getBiomeData();
         ChunkBiomeBlendingHelper chunkBiomeHelper = new ChunkBiomeBlendingHelper();
+        biomeBlendingHelper.put(cp, chunkBiomeHelper);
 
         if (use3dBiomes) {
-          biomeBlendingHelper.put(cp, chunkBiomeHelper);
           // We need to load biome data for the full height of the chunk
           // and not only limited to the sections where there are blocks
           // because it is possible that a neighboring chunks has blocks
@@ -999,7 +999,7 @@ public class Scene implements JsonSerializable, Refreshable {
                   int octNode = currentBlock;
                   Block block = palette.get(currentBlock);
 
-                  if(block.useBiomeTint())
+                  if(block.isBiomeDependant())
                     chunkBiomeHelper.makeBiomeRelevant(y);
 
                   if(block.isEntity()) {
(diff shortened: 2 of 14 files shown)


chunky/src/java/se/llbit/chunky/block/legacy/blocks/LegacyFlowerPot.java
```diff
@@ -83,4 +83,9 @@ public Tag getNewTagWithBlockEntity(Tag blockTag, CompoundTag entityTag) {
 
     return tag; // keep empty
   }
+
+  @Override
+  public boolean useBiomeTint() {
+    return true; // (in reality it is only used for fern)
```
leMaik, line 89: It should be fixed then.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -930,14 +933,25 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
         int wx0 = cp.x * 16; // Start of this chunk in world coordinates.
         int wz0 = cp.z * 16;
         BiomeData biomeData = chunkData.getBiomeData();
+        ChunkBiomeBlendingHelper chunkBiomeHelper = new ChunkBiomeBlendingHelper();
 
         if (use3dBiomes) {
-          for (int y = chunkData.minY(); y < chunkData.maxY(); y++) {
+          biomeBlendingHelper.put(cp, chunkBiomeHelper);
+          // We need to load biome data for the full height of the chunk
+          // and not only limited to the sections where there are blocks
```
leMaik, line 941: Can this be limited to `chunkData.minY() - blendRadius` … `chunkData.maxY() + blendRadius`?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -989,6 +1003,9 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
                   int octNode = currentBlock;
                   Block block = palette.get(currentBlock);
 
+                  if(block.useBiomeTint())
```
leMaik, line 1006: Maybe?

```suggestion
                  if(block.isBiomeDependant())
```

chunky/src/java/se/llbit/math/structures/Position3d2IntPackedArray.java
```diff
@@ -57,7 +64,7 @@ public int get(int x, int y, int z) {
     if(arr != null) {
       return arr[packedIndex(x, y, z)];
     }
-    return 0;
+    return -1;
```
leMaik, line 67: Does this change affect any other existing code?

leMaik (approved):
Awesome! :star_struck: 

The third optimization is smart… Could that save memory in 2d biomes too because we could have a "don't care" biome then? Similar how we have `any` nodes in the octree?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
aTom3333: No, even if a chunk is full of air (so there are no block to load) its biome can affect neighboring chunks. Like you could have a leave bloc in a chunk at y=120 (near the border of a chunk) and have no clock higher than y=70 in the neighboring chunks, you still need to load the biome of the neighboring chunk at height around 120 to compute the correct biome color.
It was vital to load before I changed Position3d2IntPackedArray.java to return -1 in case of not found because before this change, it was impossible to make the difference between not biome information not loaded and biome loaded that happens to have index of 0 so we ended up blurring with a chunk that wasn't there. Now that we can detect, and ignore, unloaded biome during blur this isn't as vital because I won't produce obviously wrong result anymore but the correct way is still to fully load

chunky/src/java/se/llbit/chunky/block/legacy/blocks/LegacyFlowerPot.java
aTom3333: Returning true only for fern would be better. But from the function just above it seem the entityTag is needed to determine whether this is a fern or not. And because returning true when it should be false doesn't lead to a big adverse effect (it may lead to unnecessary work as more biome blending than necessary is performed), I just was lazy and made it simpler

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
aTom3333: Why not, wasn't too sure of the name myself

chunky/src/java/se/llbit/math/structures/Position3d2IntPackedArray.java
aTom3333: I'm pretty sure the biome loading is the only user of this class so it shouldn't affect anything else

aTom3333:
> The third optimization is smart… Could that save memory in 2d biomes too because we could have a "don't care" biome then? Similar how we have `any` nodes in the octree?
It could be applicable with limited changes to the 2D case as well, if no block in the chunk uses biome color, the color for the biome doesn't need to be stored. We don't even need a "don't care" thing, we just won't store the color in the world texture in the first place.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
ShirleyNekoDev: How about `isBiomeTinted`? Biome dependant sounds like a block "_design_" can change depending on the biome.

chunky/src/java/se/llbit/chunky/block/legacy/blocks/LegacyFlowerPot.java
leMaik: We could rename it to `canHaveBiomeTint` and merge this.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
leMaik: Oh yeah, that makes sense. Thanks for explaining it!

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
leMaik: How about `canHaveBiomeTint`?

leMaik (approved):

chunky/src/java/se/llbit/chunky/block/legacy/blocks/LegacyFlowerPot.java
leMaik: Actually, this shouldn't be a problem at all. We replace all legacy blocks while loading a legacy chunk and let it create a new tag that takes the entity tag's information into account.

All the biome stuff is handled after converting the legacy blocks.

chunky/src/res/se/llbit/chunky/ui/render/tabs/TexturesTab.fxml
```diff
@@ -3,11 +3,15 @@
...
 <?import javafx.scene.layout.VBox?>
+<?import se.llbit.chunky.ui.IntegerAdjuster?>
+<?import javafx.scene.layout.HBox?>
 <fx:root type="javafx.scene.control.ScrollPane" xmlns="http://javafx.com/javafx/8.0.40"
          xmlns:fx="http://javafx.com/fxml/1">
   <VBox spacing="10.0">
     <CheckBox fx:id="biomeColors" mnemonicParsing="false" text="Enable biome colors" />
-    <CheckBox fx:id="biomeBlending" mnemonicParsing="false" text="Enable biome blending" />
+    <HBox alignment="CENTER_LEFT" spacing="10.0">
+      <IntegerAdjuster fx:id="biomeBlendingRadiusInput" name="Biome Blending Radius" />
```
leMaik, line 13:
```suggestion
      <IntegerAdjuster fx:id="biomeBlendingRadiusInput" name="Biome blending radius" />
```

leMaik:
Something strange is happening with the biome colors…
image

leMaik:
Rebased and resolved conflicts, I'll test again and merge this if it works now. :+1:

PR #1638, opened 2023-10-01
Add a ground plane.

ThatRedox:
This adds an infinite ground plane similar to the water plane.
image

Currently the texture is set to stone and isn't configurable. Currently the controls are in the `Water` tab so suggestions for a better place are welcome.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
@@ -95,6 +95,7 @@ public class PreviewRayTracer implements RayTracer {
     if (scene.isWaterPlaneEnabled()) {
       hit = waterPlaneIntersection(scene, ray) || hit;
     }
+    hit |= groundPlaneIntersection(scene, ray);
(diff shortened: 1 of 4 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
```diff
@@ -95,6 +95,9 @@ public static boolean nextIntersection(Scene scene, Ray ray) {
     if (scene.isWaterPlaneEnabled()) {
       hit = waterPlaneIntersection(scene, ray) || hit;
     }
+    if (scene.isGroundPlaneEnabled()) {
+      hit |= groundPlaneIntersection(scene, ray);
```
leMaik, line 99: This won't call the `groundPlaneIntersection` method if `hit` is already true, eg. if the water plane was hit before. If this gets fixed, we could have ocean grounds for water world mode. :star_struck:

ThatRedox: I think it should, `|=` uses the bitwise or operator so it can't get short circuited.

PR #1639, opened 2023-10-01
Add a 3D world texture implementation.

### LINKED ISSUE #1607: `TRIVIAL_3D` BiomeStructure implementation uses too much memory

ThatRedox:
Closes #1607.
image
2,115 chunks of my Terralith test world with the max heap set to 1 GB.

This is a 3D biome implementation that uses significantly less memory than `TRIVIAL_3D`. It is similar to `WorldTexture2D` where it only stores 8 bits per channel and deduplicates textures.

Testing with 1024 chunks of my Terralith test world:
| Implementation | Biome Texture Size (bytes) | Loading Time (sec) | Chunks loaded with 512mb heap |
|-------------------|------------------------------|----------------------|--------------------------------------|
| World Texture 3D (this) | 61,735,768 | 56.772 | 942 / 1024 |
| Trivial 3D | 7,250,510,208 | 50.801 | 31 / 1024 |
| World Texture 2D | 345,600 | 2.898 | 1024 / 1024 |
| Trivial 2D | 28,434,816 | 2.816 | 1024 / 1024 |

Other changes:
* Separated `compact` and `endFinalization` for `BiomeStructure`
* Added dedicated `Interner` interface (and tests!) for deduplication
* Refactored `WorldTexture2D` to share code with `WorldTexture3D`


chunky/src/java/se/llbit/util/interner/Interner.java
```diff
@@ -0,0 +1,41 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with Chunky.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package se.llbit.util.interner;
+
+/**
+ * This deduplicates objects similar to `String.intern()`, but for arbitrary objects.
+ */
+public interface Interner<T> {
```
leMaik, line 24: I'm not sure about this name. Maybe `ObjectDeduplicator`?

ThatRedox: `Interner` seems to be the generally used name for this. Google Guava has a similar interface also named [`Interner`](https://guava.dev/releases/22.0/api/docs/com/google/common/collect/Interner.html) and there is also [`String.intern()`](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html?is-external=true#intern--).

PR #1640, opened 2023-10-01
Small tasktracker improvments.

ThatRedox:
* Log completed steps and elapsed time when a task is closed
* Add more detailed step information in scene loading


leMaik (approved):
Nice improvement.
I was about to suggest a subtask API, but maybe we shouldn't over-engineer this.

Issue #1208, opened 2022-03-20
Minecraft sky simulation


Peregrine05:
Add an option to simulate the Minecraft sky in Chunky. The sky should change according to the sun position, and even display stars and maybe the moon (and add a setting to change the moon phase?) when the sun is below the horizon. The color of the sky can be changed slightly using biome temperature and humidity settings, or any other factors that change the sky blueness, as in Minecraft, and the weather can also be changed, such as to rain or thunder, to make the sky gray.

jackjt8:
Related: https://github.com/chunky-dev/chunky/issues/497 https://github.com/chunky-dev/chunky/issues/731

Issue #731, opened 2020-11-21
Native Night Support


RocketNinja15:
If night support was build in, aka, you lower the sun altitule, and then it is automatically into the night. So that would include sky brightness, sun intensity, the moon. It would not just set the sky brightness to 0.1, it would change the engine for the sky. Lighting would change to the moon, the fog would auto-enable, the Fullbright horizon effect would go away.

leMaik:
Great idea! Why I wouldn't necessarily make this automated, being able to optionally enable moon, stars and have different moon phases, would be a really cool feature.

On a different note: Moon and sun are actually visible at the same time most of the days.

RocketNinja15:
Well, I think that if you lower the sun altitude below the horizon, that it would bring up the moon and the stars, compared to having it stay above as it does right now

jackjt8:
Related to issue #86 however as this one has more information I'll close the original issue.

Regarding setting the Sun altitude below 0 degrees there is issue #497 which is open. Making sure to reference it here in case this gets implemented. Both the Simulated Sky and Fog are tied into the sun position and may experience unexpected behaviour.

PR #1655, opened 2023-10-22
Feature/moon celestial

ShirleyNekoDev:
Adds logic for different sun textures (moon / custom) including moon phases.
Only in code at the moment including loading/saving into scene config, UI can be added later.
This was recently requested in Discord and in previous issues (#1208, #731).
based on top of #1654 (diff is way smaller)


chunky/src/java/se/llbit/chunky/renderer/scene/sky/CelestialBodyType.java
```diff
@@ -0,0 +1,219 @@
...
+import se.llbit.resources.ImageLoader;
+import se.llbit.util.Registerable;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Supplier;
+
+// TODO: use https://openjdk.org/jeps/409 sealed classes
```
ThatRedox, line 17: Maybe we shouldn't make this sealed so plugins in the future have the option to add their own celestial bodies

leMaik:
This needs a rebase after the recent refactorings.

ShirleyNekoDev:
Should be ready now :)

> This needs a rebase after the recent refactorings.

leMaik:
Do we need to reset the moon texture after reloading the resource packs? wouldn't it be better to have a MoonPhaseTextureLoader that handles this?

ShirleyNekoDev:
> Do we need to reset the moon texture after reloading the resource packs? wouldn't it be better to have a MoonPhaseTextureLoader that handles this?

Yeah, probably a good idea to refresh the moon texture after a resource pack reload!
How would you expect the MPTextureLoader to work? I haven't had a look at this part of the code base for a while. Do you want to append it to this PR? :)

PR #1656, opened 2023-10-24
Add sha256 support for library verification to the launcher and a flag to disable it.

### LINKED ISSUE #1509: Add an option to disable md5 checks
One way to implement this could be a command line argument such as `--dangerouslyDisableVerification`.

leMaik:
It's 2023 and md5 is basically broken. This PR adds sha256 verification support. The `latest.json` for Chunky 2.4.5 on my server already contains the sha256 hashes.

Using `--dangerouslyDisableLibraryValidation`, checksum validation of libraries can now be disabled. This is a potential security issue and thus not recommended. It does help with local debugging and sharing custom Chunky versions though. Closes #1509

### DIFF
--- a/lib/src/se/llbit/util/Util.java
+++ b/lib/src/se/llbit/util/Util.java
@@ -60,7 +60,7 @@ public class Util {
   }
 
   /**
-   * @return the MD5 hash sum of the given file, in hexadecimal format.
+   * @return the SHA256 hash sum of the given file, in hexadecimal format.
    * Returns an error message if there was an error computing the checksum.
    */
   public static String sha256sum(File library) {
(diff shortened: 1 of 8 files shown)


leMaik:
Having that flag as system property means that the checksum validation could be disabled by environment variables (ie. `JAVA_TOOL_OPTIONS`. I'll change this to a command line flag.

lib/src/se/llbit/util/Util.java
```diff
@@ -59,6 +59,29 @@ public static String md5sum(File library) {
     }
   }
 
+  /**
+   * @return the MD5 hash sum of the given file, in hexadecimal format.
```
leMaik, line 63:
```suggestion
   * @return the SHA256 hash sum of the given file, in hexadecimal format.
```

PR #1658, opened 2023-10-28
Prevent locking camera from resetting render

### LINKED ISSUE #1546: Allow the selected camera to be locked without refreshing the scene

Peregrine05:
Fixes #1546.


chunky/src/java/se/llbit/chunky/renderer/scene/Camera.java
```diff
@@ -788,7 +789,10 @@ public ApertureShape getApertureShape() {
 
   public void setCameraLocked(boolean value) {
     lockCamera = value;
-    scene.refresh();
+    if (scene.mode == RenderMode.PREVIEW) {
```
leMaik, line 792: Why would we need to refresh the scene at all just because the camera is locked/unlocked?

leMaik (approved):

Issue #1651, opened 2023-10-20
Exhance TIFF format export


ShirleyNekoDev:
After working on the TIFF writer in #1649 I'm tempted to implement some additional features besides baseline TIFF export:
- compression (https://www.awaresystems.be/imaging/tiff/tifftags/compression.html)
- int32 / int16 format (https://www.awaresystems.be/imaging/tiff/tifftags/sampleformat.html)
- embedded alpha channel (https://www.awaresystems.be/imaging/tiff/tifftags/extrasamples.html)
- alpha mask & preview layer (https://www.awaresystems.be/imaging/tiff/tifftags/newsubfiletype.html)
- selectable embedded metadata
- integrate with #1650

ShirleyNekoDev:
TODO: add UI compatibility warning for non-baseline features (https://github.com/chunky-dev/chunky/pull/1661#issuecomment-1785792674)

PR #1659, opened 2023-10-29
TIFF compression

ShirleyNekoDev:
- refactored TIFF export (again)
- easier to extend (for #1651)
- added compression (part of #1651)
- should be faster too

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/export/Tiff32ExportFormat.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/export/Tiff32ExportFormat.java
@@ -17,9 +17,11 @@
  */
 package se.llbit.chunky.renderer.export;
 
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import se.llbit.chunky.renderer.scene.Scene;
+import se.llbit.imageformats.tiff.CompressionType;
 import se.llbit.imageformats.tiff.TiffFileWriter;
 import se.llbit.util.TaskTracker;
 
@@ -50,8 +52,12 @@ public class Tiff32ExportFormat implements PictureExportFormat {
 
   @Override
   public void write(OutputStream out, Scene scene, TaskTracker taskTracker) throws IOException {
+    assert(out instanceof FileOutputStream);
     try (TaskTracker.Task task = taskTracker.task("Writing TIFF");
-        TiffFileWriter writer = new TiffFileWriter(out)) {
+        TiffFileWriter writer = new TiffFileWriter(
+          ((FileOutputStream) out).getChannel(),
+          CompressionType.DEFLATE
+        )) {
       writer.export(scene, task);
     }
   }
--- a/chunky/src/java/se/llbit/chunky/renderer/export/Tiff32ExportFormat.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/export/Tiff32ExportFormat.java
@@ -19,7 +19,14 @@ package se.llbit.chunky.renderer.export;
 
 import java.io.FileOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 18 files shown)


ShirleyNekoDev:
```
RepresentationInformation: [...]\snapshots\default_2023-10-20_00-28-45-100.tiff
 LastModified: 2023-10-29 05:26:10 CET
 Size: 2666360
 Format: TIFF
 Version: 6.0
 Status: Well-Formed and valid
 SignatureMatches:
  TIFF-hul
 MIMEtype: image/tiff
 TIFFMetadata:
  ByteOrder: big-endian
  IFDs:
   Number: 1
   IFD:
    Offset: 8
    Type: TIFF
    Entries:
     NisoImageMetadata:
      FormatName: image/tiff
      ByteOrder: big_endian
      CompressionScheme: PKZIP-style Deflate encoding
      ImageWidth: 500
      ImageHeight: 500
      ColorSpace: RGB
      ReferenceBlackWhite: 0, 0, 0, 0, 0, 0
      DateTimeCreated: 2023-10-29T05:26:10
      ScanningSoftware: Chunky ?-snapshot
      Orientation: normal
      SamplingFrequencyUnit: no absolute unit
      XSamplingFrequency: 1
      YSamplingFrequency: 1
      BitsPerSample: 32, 32, 32
      BitsPerSampleUnit: f32
      SamplesPerPixel: 3
     DateTime: 2023:10:29 05:26:10
     NewSubfileType: 0
     SampleFormat: 3, 3, 3
```

chunky/src/java/se/llbit/chunky/renderer/export/Tiff32ExportFormat.java
```diff
@@ -50,8 +52,12 @@ public boolean isTransparencySupported() {
 
   @Override
   public void write(OutputStream out, Scene scene, TaskTracker taskTracker) throws IOException {
+    assert(out instanceof FileOutputStream);
```
leMaik, line 55: This is an assumption that violates the method signature. From what I see, a FOS is used to create a file channel to create a buffered output stream on top of it. Why?

leMaik (changes requested):

chunky/src/java/se/llbit/chunky/renderer/export/Tiff32ExportFormat.java
ShirleyNekoDev: TIFF format is annoying and uses pointers to addresses which are typically only known after writing parts of the data. I had 2 ways to implement this and chose not to buffer everything in memory before writing it to an unspecified OutputStream. Instead the FileChannel is used to reposition the writer to replace placeholder values for the pointers after writing everything.
Because we currently only use FileOutputStream for PictureExportFormats, it might be better to replace the first argument with a File instead? From that I can create the custom Tiff[...]OutputStream.

leMaik: Allright, but then let's make this a proper runtime exception to yell loudly at whoever dares to mis-use this:


```suggestion
  if (!(out instanceof FileOutputStream)) {
    throw new UnsupportedOperationException("Tiff export is only implemented for FileOutputStream");
  }
```

ShirleyNekoDev: Added a fallback using a temp file

leMaik: That's even better, of course. 👏

leMaik (approved):

leMaik:
Thank you for your work on fixing and improving Chunky's tiff support! 🎉

PR #1660, opened 2023-10-29
More accurate fog implementation

### LINKED ISSUE #665: Limitations of the current fog implementation
Below is a list of issues with the current fog implementation.

---

Fog is illuminated only by the suns position. Whether it is enabled or disabled does not matter.

Sun enabled
image

Sun disabled
image

Sun disabled + Sky Black 
image

---

Emitters do not illuminate fog.

No Fog
image

Fog
image

Realworld example
image

---

Fog has a uniform color and is not tinted by the color of light. ie- Fog color on either side of stained glass is the same.

Fast Fog (Enabled)
glass_fog2_ff-512

(Slow) Fog
glass_fog2_sf-512

Realworld example
image

---

Peregrine05:
This PR is draft because changes are pending.

Fixes #665.

This PR adds more accurate fog implementation that interacts with indirect lighting and emitter lighting correctly *(It does not yet work with Emitter Sampling Strategy). This fog implementation works independently of the old fog implementation, so both can be used together if so desired.

**Old**:
image

**New**:
image

The bounds of the fog is controlled by fog volumes. Each fog volume has its own independent properties, which include material and color properties. Four shapes of volume may be used. These are `EXPONENTIAL`, `LAYER`, `SPHERE`, and `CUBOID`.

This PR also applies the fog implementation to the clouds (which is toggleable), and adds more configuration options to the clouds. These include the ability to add multiple layers of clouds to the scene, the ability to scale the clouds per-dimension rather than just on the X- and Z-dimensions in conjunction, and the ability to set the color and material per cloud layer.

A new material property specific to the fog material, called `Anisotropy`, controls the direction light that hits the fog is likely to be scattered. The range of the property is `[-1, 1]`. Negative values increase the probability that light is scattered toward the direction from which it came, and positive values increase the probability that light is scattered into the direction to which it goes.


chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
```diff
@@ -83,21 +85,27 @@ public static double skyOcclusion(Scene scene, WorkerState state) {
 
   /**
    * Find next ray intersection.
+   * @param random Used for particle fog, can be null if particleFog is false
    * @return Next intersection
    */
-  public static boolean nextIntersection(Scene scene, Ray ray) {
+  public static boolean nextIntersection(Scene scene, Ray ray, Random random, IntersectionConfig config) {
```
ThatRedox, line 91: I think it would be better for this to give the intersection with the entire fog volume instead of needing to pass in a random and intersect in here. That would give ray tracers more flexibility on what it wants to do with the fog (ie ignore it, intersect with fog, calculate attenuation of fog, etc).

Issue #1650, opened 2023-10-20
Image export options


ShirleyNekoDev:
I'd like to have a UI to modify additional options of an image format when exporting a render. These should be saved globally for each format (will be restored after switching to a different format).
To modify these options, a button is added alongside the format dropdown. When a format has options, it will open a dialog box or expand a collapsed field, otherwise it will be disabled.

PR #1661, opened 2023-10-30
TIFF alpha channel

ShirleyNekoDev:
- refactored alpha channel computation and storage
- renamed some PictureExportFormat stuff and replaced `isTransparencySupported` with `transparencyType`
- added TIFF alpha channel export (#1651)
- PR based on top of PR 1659

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/export/PictureExportFormat.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/export/PictureExportFormat.java
@@ -53,7 +53,7 @@ public interface PictureExportFormat {
 
   /**
    * @return true, if this export format supports exporting the alpha channel
-   * @deprecated Replaced by {@link #transparencyType()} and usage of {@link AlphaBuffer}
+   * @deprecated Replaced by {@link #getTransparencyType()} and usage of {@link AlphaBuffer}
    */
   @Deprecated(forRemoval = true)
   default boolean isTransparencySupported() {
@@ -65,7 +65,7 @@ public interface PictureExportFormat {
    *
    * @return the required format for the alpha buffer or {@link AlphaBuffer.Type#UNSUPPORTED} if alpha is not supported.
    */
-  default AlphaBuffer.Type transparencyType() {
+  default AlphaBuffer.Type getTransparencyType() {
     return isTransparencySupported() ? AlphaBuffer.Type.UINT8 : AlphaBuffer.Type.UNSUPPORTED;
   }
 
--- a/chunky/src/java/se/llbit/chunky/renderer/export/PictureExportFormat.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/export/PictureExportFormat.java
@@ -54,10 +54,10 @@ public interface PictureExportFormat {
   /**
    * Note: It depends on the scene settings if the alpha buffer will be available on export.
    *
-   * @return the required format for the alpha buffer or DISABLED if alpha is not supported.
+   * @return the required format for the alpha buffer or {@link AlphaBuffer.Type#UNSUPPORTED} if alpha is not supported.
    */
   default AlphaBuffer.Type transparencyType() {
-    return AlphaBuffer.Type.DISABLED;
+    return AlphaBuffer.Type.UNSUPPORTED;
   }
 
   /**
--- a/chunky/src/java/se/llbit/chunky/renderer/export/PictureExportFormat.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/export/PictureExportFormat.java
@@ -18,6 +18,8 @@ package se.llbit.chunky.renderer.export;
 
 import java.io.IOException;
 import java.io.OutputStream;
+
+import se.llbit.chunky.renderer.scene.AlphaBuffer;
 import se.llbit.chunky.renderer.scene.Scene;
 import se.llbit.util.TaskTracker;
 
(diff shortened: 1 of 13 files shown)


ShirleyNekoDev:
Note: The TIFF tag "extra samples" is an extension to the TIFF baseline. Some software might not support it. 
From my experiments even software which can decode alpha samples, might not accept in fp32 format (e.g. IrfanViewdoesn't work; Gimp works). We should implement a UI warning when working on #1650 in another PR.

chunky/src/java/se/llbit/chunky/renderer/export/PictureExportFormat.java
```diff
@@ -50,12 +52,20 @@ default String getDescription() {
...
+   */
+  default AlphaBuffer.Type transparencyType() {
+    return AlphaBuffer.Type.UNSUPPORTED;
+  }
+
+  /**
+   * @return true, if the export formats wants preprocessed buffer data<br>
+   *         false, if the export format uses the unprocessed sampling data
    */
-  default boolean isTransparencySupported() {
```
leMaik, line 57: This breaks plugins that adds export formats.

chunky/src/java/se/llbit/chunky/renderer/export/PictureExportFormat.java
```diff
@@ -50,12 +52,20 @@ default String getDescription() {
...
-    return false;
+  default boolean wantsPostprocessing() {
```
leMaik, line 67:
```suggestion
  default boolean isPostprocessed() {
```

chunky/src/java/se/llbit/chunky/renderer/export/PictureExportFormat.java
```diff
@@ -50,12 +52,20 @@ default String getDescription() {
   String getExtension();
 
   /**
-   * Check if this format supports transparency (used for transparent sky).
+   * Note: It depends on the scene settings if the alpha buffer will be available on export.
    *
-   * @return True if this format supports transparency, false otherwise
+   * @return the required format for the alpha buffer or {@link AlphaBuffer.Type#UNSUPPORTED} if alpha is not supported.
+   */
+  default AlphaBuffer.Type transparencyType() {
```
leMaik, line 59: Rename to keep Java getter style:

```suggestion
  default AlphaBuffer.Type getTransparencyType() {
```

chunky/src/java/se/llbit/chunky/renderer/scene/PreviewRayTracer.java
```diff
@@ -83,7 +83,7 @@ public static double skyOcclusion(Scene scene, WorkerState state) {
 
   /**
    * Find next ray intersection.
-   * @return Next intersection
+   * @return true if intersected, false if no intersection has been found
```
leMaik, line 86: Nice :+1:

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -486,9 +485,9 @@ public synchronized void copyState(Scene other, boolean copyChunks) {
       height = other.height;
       backBuffer = other.backBuffer;
       frontBuffer = other.frontBuffer;
-      alphaChannel = other.alphaChannel;
       samples = other.samples;
     }
+    alphaBuffer.reset();
```
leMaik, line 490: Previously, the alpha buffer of the other scene was copied. Why was this removed?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -2631,7 +2563,7 @@ public void setUseCustomWaterColor(boolean value) {
       ((Configurable) postProcessingFilter).storeConfiguration(postprocessJson);
       json.add("postprocessSettings", postprocessJson);
     }
-    json.add("outputMode", outputMode.getName());
+    json.add("outputMode", pictureExportFormat.getName());
```
leMaik, line 2566: Make sure to keep the old names to not make this a breaking change for existing scene files.

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -37,8 +37,8 @@
 import se.llbit.chunky.renderer.scene.Scene;
 import se.llbit.chunky.world.Material;
 import se.llbit.log.Log;
-import se.llbit.util.PositionalInputStream;
-import se.llbit.util.PositionalOutputStream;
+import se.llbit.util.io.PositionalInputStream;
```
leMaik, line 40: Is moving these classes really necessary?

leMaik (changes requested):

chunky/src/java/se/llbit/chunky/renderer/export/PictureExportFormat.java
ShirleyNekoDev: This would break plugin compatability

ShirleyNekoDev: Also I like this name because it describes if the format expects a buffer with postprocessed data before exporting. See `Scene::writeFrame()`

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
ShirleyNekoDev: Scene resetting/updating is pretty complicated right now and I'm not sure if the alphaBuffer is cleared every time something changes for the camera perspective. Also when the export format changes, it might have to be recalculated now.
Better save to not keep it here and recalculate this 1 frame before export.
(But if everybody is sure that the buffer can be copied, we can bring it back -- and make the scene-updating easier to understand :)

chunky/src/java/se/llbit/math/Octree.java
ShirleyNekoDev: I like structure, especially in the util-package :)
Not really necessary, but there are many classes in it which do very different things. And I did not expect plugins to use these IO-classes for now.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
leMaik:
```suggestion
    alphaBuffer.reset(); // TODO could we copy it without resetting if the export format didn't change?
```

chunky/src/java/se/llbit/chunky/renderer/export/PictureExportFormat.java
leMaik: Oh wait, we had `wantsPostprocessing` before? :grimacing: Weird name, but allright.

leMaik (approved):

leMaik (approved):

ShirleyNekoDev:
I'll rebase this once #1659 is merged 😵‍💫

PR #1663, opened 2023-11-03
fix exposed, weathered, and oxidized copper grates

JustinTimeCuber:
Due to an oversight, I only made regular copper grates support transparency, not any of the oxidized variants:
image
image
(second image is on the emitter mapping branch just because that's what I had easily available)


leMaik:
Oh... The image hides that very well, didn't even notice it before.

Issue #1591, opened 2023-06-14
Smoothness should affect transmitted light


Peregrine05:
Currently, the smoothness property affects only specularly-reflected light, but it should have an affect on light that is transmitted through the material.

NotStirred:
:+1: I believe blender has a specific "transmission roughness" property for this

PR #1668, opened 2023-11-05
Rransmission roughness

ConsueTerra:
This pr is the first of a series to implement a better BTDF in chunky. This pr adds a linear lerping like in specular reflections in other to enable glossy reflections, a new material parameter called `trasmissionRoughness` which is set independently from `roughness` the UI parameter is Transmission smoothness. Fixes Issue #1591

|Before (1 smoothness,1 transmission smoothness)|After (1 smoothness,0.8 transmission smoothness)| 
|---|---|
|image|image|

A scene example 80k SSP, the white glass has a transmission smoothness of 1, red glass of 0.75, blue glass of 0.5, notice the increase spread of the caustics (sun sampling is OFF)
image

This pr is a draft, there are 2 pending issues:

- [ ] fix the bias caused from lerping b/w diffuse and specular (also present in reflections)
- [ ] fix intersections with air, currently refracted rays don't intersect air and "exit" materials, this causes 1. incorrect refractions with air gap, 2. eliminated internal refection events. This problem was okay with pure specular reflections but is a serious concern when introducing roughness.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/PathTracer.java
@@ -265,8 +265,8 @@ public class PathTracer implements RayTracer {
       double directLightB = 0;
 
       boolean frontLight = next.d.dot(ray.getNormal()) > 0;
-
-      if (frontLight || (currentMat.subSurfaceScattering
+      //check if normal faces the sun direction, if so do sampling
+      if (frontLight || (false
         && random.nextFloat() < Scene.fSubSurface)) {
 
         if (!frontLight) {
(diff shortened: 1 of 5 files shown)


ConsueTerra:
I have included diffuse transmission as well, 

|Before )|After (0.3 transmission)| 
|---|---|
|image|image|

This is only ray depth 5 so the effect is suble but you can see softer shadows on the leaves (some of the light bleeds to the back) and the shadows have a green tint. Even with a high Ray depth it was near impossible to light these forests scenes properly. The new parameter is `additionalTransmission`.0.3 is a good value for allowing back facing light without taking to much brightness from the front. Currently diffusion is split b/w refection and transmission, this is energy conserving on some block models but not others, probably due to how intersections currently work. 

A look at ray depth 20, 72 samples (1600 by 1600)
forest-72

PR #1672, opened 2023-11-10
Homogenize Depth

ConsueTerra:
Chunky would rely `Ray.set(Ray)` to increase the depth, however different routines calls this a different amount of times leading to different ray depths in a scene. Here I have removed that change and made ray depth increase explicitly to successive frames in `PathTracer.pathTrace`.  Currently this does  not explicitly change the output, but it could be a easy mistake in the future or if the ray depth is needed in other parts of the code.


leMaik:
Yeah, stuff like this makes understanding the path tracing in Chunky even harder. Thanks for changing it. :+1:

PR #1673, opened 2023-11-12
Fix refraction and material transitions.

### LINKED ISSUE #1515: IOR is incorrect

ConsueTerra:
This PR fixes #1515, in a way that is less hacky than #1601 by defining octree type and modifying the intersection logic based on which octree we are in. Also recovers total internal reflections. This is because Air has different contexts depending on location. Current changes fix solid block refractions, and partially addresses sub block refractions like glass panes. Currently I cant figure out a good methodology for fixing glass panes that does not break intersection logic. Additionally there are quite a few edge cases to consider like partially water logged glass panes. 

|Before|After| 
|---|---|
|image|image|
|image|image|
|image|image|

Now this makes for some *weird* views, dont know if these are correct but the above tests makes it seems so, in any case that is a function of refraction and Fresnel code than intersection code
image

Glass panes are still wrong example:
image

### DIFF
--- a/chunky/src/java/se/llbit/math/Octree.java
+++ b/chunky/src/java/se/llbit/math/Octree.java
@@ -244,10 +244,25 @@ public class Octree {
     }
   }
 
+  /** The type of Octree, used to adjust intersection code based on contents*/
+  public enum OctreeType {
+    /**
+     * Contains all world geometry excluding water, water is removed from water logged blocks
+     * and full water blocks are replaced with air
+     */
+    WORLD,
+    /**
+     * Contains all the water geometry only, other blocks are air
+     */
+    WATER
+  }
+
   /**
    * Timestamp of last serialization.
    */
   private long timestamp = 0;
+
+  public OctreeType type;
 
   private OctreeImplementation implementation;
 
@@ -257,9 +272,10 @@ public class Octree {
    *
    * @param octreeDepth The number of levels in the Octree.
    */
-  public Octree(String impl, int octreeDepth) {
+  public Octree(String impl, int octreeDepth, OctreeType type) {
     Log.infof("Building new octree (%s)", impl);
     implementation = getImplementationFactory(impl).create(octreeDepth);
+    this.type = type;
   }
 
   protected Octree(OctreeImplementation impl) {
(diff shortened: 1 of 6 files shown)


ConsueTerra:
I would suggest changing `Ray.prevMaterial` to `Ray.insideMaterial` or something like that..? why? because prev material is confusing (ie the past surface or where we are?), and because even though refractions are fixed, it is not applying the right material at the boundary.

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -539,9 +557,14 @@ public boolean enterBlock(Scene scene, Ray ray, BlockPalette palette) {
         ray.distance += distance;
         distance = 0;
         if (currentBlock.intersect(ray, scene)) {
-          if (prevBlock != currentBlock)
+          if (prevBlock != currentBlock) { //|| (currentBlock.refractive)) {
+            //for things like glass panes coerce the current mat to be air or water
+            // TODO testing
+            if (currentBlock.refractive) {
+              //ray.setCurrentMaterial(currentBlock.waterlogged ? Water.INSTANCE : Air.INSTANCE);
+            }
```
leMaik, line 565: Either remove this or comment it alltogether.

```suggestion
            // if (currentBlock.refractive) {
            //    ray.setCurrentMaterial(currentBlock.waterlogged ? Water.INSTANCE : Air.INSTANCE);
            // }
```

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -539,9 +557,14 @@ public boolean enterBlock(Scene scene, Ray ray, BlockPalette palette) {
         ray.distance += distance;
         distance = 0;
         if (currentBlock.intersect(ray, scene)) {
-          if (prevBlock != currentBlock)
+          if (prevBlock != currentBlock) { //|| (currentBlock.refractive)) {
```
leMaik, line 560: ?

chunky/src/java/se/llbit/math/Octree.java
```diff
@@ -484,6 +500,8 @@ public boolean enterBlock(Scene scene, Ray ray, BlockPalette palette) {
     int depth = implementation.getDepth();
 
     double distance = 0;
+    //tread air as a null block depending on octreetype and were we are in the trace
```
ThatRedox, line 503:
```suggestion
    // Treat air as a null block depending on octreetype and were we are in the trace
```

leMaik:
Given that #1830 seems to replace this PR, I'd like to maybe merge this into 2.5.0 but not 2.6.0? :thinking: @Peregrine05 and @ConsueTerra, what do you think? What's the difference between the fixes in #1830 (which I still refuse to figure out due to the vast amount of unrelated commits) and this PR?

PR #1677, opened 2023-12-18
Fix for possible crash in SceneChooserController.java

jyoo980:
An unchecked call on the Optional result of `Alert#showAndWait()` may throw an exception at run time.

This commit refactors it to one of the recommended approaches to safely get its value [here](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Alert.html).


leMaik (approved):

jyoo980:
Will push a fix for the build failure!

leMaik:
Thank you. Not sure if an error would ever occur with this particular dialog, but it is certainly the proper way to handle this.

PR #1683, opened 2024-01-16
Clarify render status UI

### LINKED ISSUE #1213: The SPP counter is inaccurate
Currently, for example, it states `Rendering 1 of 100`. That is inaccurate because while it states "Rendering 1 of 100", it is actually rendering 2 of 100. Therefore, to be accurate, it should either state, instead of "Rendering 1 of 100", `Rendered 1 of 100`, OR `Rendering 2 of 100`.

imsyphia:
The current render status pane is somewhat messy with odd and inconsistent formatting, leading to issues like #1213, where `Rendering: 5 of 7` can be misinterpreted as _currently_ rendering 5 of 7. I tried to make it overall more readable and intuitively organized.

This (mostly) fixes #1213. It can still be misinterpreted, but it should be somewhat clearer without having to add additional visual clutter like "5 of 7 done".

I've attached before/after images below for comparison.

chunkyuibefore

chunkyuiafter

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/PreviewRenderer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/PreviewRenderer.java
@@ -54,7 +54,7 @@ public class PreviewRenderer extends TileBasedRenderer {
   @Override
   public void render(DefaultRenderManager manager) throws InterruptedException {
     TaskTracker.Task task = manager.getRenderTask();
-    task.update("Preview", 2, 0, "");
+    task.update("Displaying preview", 2, 0, "");
 
     Scene scene = manager.bufferedScene;
 
--- a/chunky/src/java/se/llbit/chunky/renderer/PreviewRenderer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/PreviewRenderer.java
@@ -54,7 +54,7 @@ public class PreviewRenderer extends TileBasedRenderer {
   @Override
   public void render(DefaultRenderManager manager) throws InterruptedException {
     TaskTracker.Task task = manager.getRenderTask();
-    task.update("Displaying preview", 2, 0, "");
+    task.update("Preparing preview", 2, 0, "");
 
     Scene scene = manager.bufferedScene;
 
--- a/chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
+++ b/chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
@@ -162,8 +162,11 @@ public class ChunkyFxController
     @Override public void setProgress(String task, int done, int start, int target) {
       Platform.runLater(() -> {
         progressBar.setProgress((double) done / (target - start));
-        progressLbl.setText(String.format("%s: %s of %s", task, decimalFormat.format(done),
-            decimalFormat.format(target)));
+        if (target - start > 0) {
+          progressLbl.setText(String.format("%s - %s%%", task, Math.round(100 * (double) done / (target - start))));
+        } else {
+          progressLbl.setText(String.format("%s", task));
+        }
         etaLbl.setText("ETA: N/A");
       });
     }
@@ -171,8 +174,11 @@ public class ChunkyFxController
     @Override public void setProgress(String task, int done, int start, int target, String eta) {
       Platform.runLater(() -> {
         progressBar.setProgress((double) done / (target - start));
-        progressLbl.setText(String.format("%s: %s of %s", task, decimalFormat.format(done),
-            decimalFormat.format(target)));
+        if (target - start > 0) {
+          progressLbl.setText(String.format("%s - %s%%", task, Math.round(100 * (double) done / (target - start))));
+        } else {
+          progressLbl.setText(String.format("%s", task));
+        }
         etaLbl.setText("ETA: " + eta);
       });
     }
@@ -205,8 +211,7 @@ public class ChunkyFxController
         int seconds = (int) ((time / 1000) % 60);
         int minutes = (int) ((time / 60000) % 60);
         int hours = (int) (time / 3600000);
-        gui.renderTimeLbl.setText(String
-            .format("Render time: %d hours, %d minutes, %d seconds", hours, minutes, seconds));
+        gui.renderTimeLbl.setText(String.format("Time: %d:%02d:%02d", hours, minutes, seconds));
       });
     }
 
@@ -222,7 +227,7 @@ public class ChunkyFxController
 
     private void updateSppStats() {
       Platform.runLater(() -> gui.sppLbl.setText(String
-          .format("%s SPP, %s SPS", gui.decimalFormat.format(spp),
+          .format("%s SPP | %s SPS", gui.decimalFormat.format(spp),
               gui.decimalFormat.format(sps))));
     }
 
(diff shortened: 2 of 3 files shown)


jackjt8:
This is definitely an improvement. Having Render time and ETA on the same line in the same format is really nice.

I am still not 100% happy with the "Rendering" line... but as you have said.. I'm not sure how we can "fix" it without adding more visual clutter. I wonder if hover text with more verbose information would be a good or bad idea.. It's clutter.. just not visible? Or would a button to toggle verbose status UI be silly?

leMaik:
Great first PR! 🎉 

How about calculating a percentage? Eg. "Rendering (42%)", if that number is not zero? The numbers are spp for rendering but otherwise pretty arbitrary anyways, and a percentage is nicer in formatting imho.

imsyphia:
> Great first PR! 🎉
> 
> How about calculating a percentage? Eg. "Rendering (42%)", if that "of" number is not zero? The numbers are spp for rendering but otherwise pretty arbitrary anyways, and a percentage is nicer in formatting imho.

I actually did try adding a percentage, but the combination of counter and percentage looked a little clunky. Doing `Rendering | 42%` or something similar could totally work, though. And it would fix the "Rendering" issue as @jackjt8 stated. I'll try it when I get around to it.

My only concern is that having a specific number could be useful for debugging in certain circumstances, so it may be worth adding that information elsewhere if it's not already available.

imsyphia:
Also, this may fall slightly outside the scope of this PR, but job names are inconsistent. Some are nouns, like `Preview`, and others are active verbs, like `Rendering`. If this is a simple fix I'll likely include it, since it is mostly relevant to the status pane, but if it's more complicated I'll make a separate issue.

leMaik:
Feel free to update the task names to verbs, while you're at it. 👍

imsyphia:
I've replaced the `N of M` indicator with percentage progress, and renamed "Render time" to "Time", since it should be clear enough from context, and is not necessarily restricted to rendering.

The percentage indicator is only two digits because additional digits on the percentage progress looked awkward and was unnecessary given that there's already a timer, progress bar, and SPP counter to indicate that progress is being made.

The "Preview" task is now named "Displaying preview", since it was the only inconsistent one. I avoided "Rendering preview" because a newer user could confuse it with the actual rendering process. 

chunkyuiafter2

imsyphia:
This is definitely outside the scope of this PR, but I was thinking that the top-right of the pane doesn't need to be exclusively SPP and SPS. It could be used as a general-purpose info/status list for the current task. Should I make an issue for that?

chunky/src/java/se/llbit/chunky/renderer/PreviewRenderer.java
```diff
@@ -54,7 +54,7 @@ public String getDescription() {
   @Override
   public void render(DefaultRenderManager manager) throws InterruptedException {
     TaskTracker.Task task = manager.getRenderTask();
-    task.update("Preview", 2, 0, "");
+    task.update("Displaying preview", 2, 0, "");
```
leMaik, line 57: It's not the displaying that takes a bit of time, it's rendering the preview:

```suggestion
    task.update("Rendering preview", 2, 0, "");
```

chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
```diff
@@ -162,17 +162,23 @@ public class ChunkyFxController
     @Override public void setProgress(String task, int done, int start, int target) {
       Platform.runLater(() -> {
         progressBar.setProgress((double) done / (target - start));
-        progressLbl.setText(String.format("%s: %s of %s", task, decimalFormat.format(done),
-            decimalFormat.format(target)));
+        if (target - start > 0) {
+          progressLbl.setText(String.format("%s - %s%%", task, Math.round(100 * (double) done / (target - start))));
```
leMaik, line 166: `%s` is for strings, maybe we can even have one decimal digit (when progress is slow):

```suggestion
          progressLbl.setText(String.format("%s - %.1f%%", task, Math.round(100 * (double) done / (target - start))));
```

chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
```diff
@@ -162,17 +162,23 @@ public class ChunkyFxController
...
       });
     }
 
     @Override public void setProgress(String task, int done, int start, int target, String eta) {
       Platform.runLater(() -> {
         progressBar.setProgress((double) done / (target - start));
-        progressLbl.setText(String.format("%s: %s of %s", task, decimalFormat.format(done),
-            decimalFormat.format(target)));
+        if (target - start > 0) {
+          progressLbl.setText(String.format("%s - %s%%", task, Math.round(100 * (double) done / (target - start))));
```
leMaik, line 178:
```suggestion
          progressLbl.setText(String.format("%s - %.1f%%", task, Math.round(100 * (double) done / (target - start))));
```

chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
```diff
@@ -205,8 +211,7 @@ public GUIRenderListener(ChunkyFxController renderControls) {
         int seconds = (int) ((time / 1000) % 60);
         int minutes = (int) ((time / 60000) % 60);
         int hours = (int) (time / 3600000);
-        gui.renderTimeLbl.setText(String
-            .format("Render time: %d hours, %d minutes, %d seconds", hours, minutes, seconds));
+        gui.renderTimeLbl.setText(String.format("Time: %d:%02d:%02d", hours, minutes, seconds));
```
leMaik, line 214: Geat! 👏

leMaik (changes requested):
Great first PR 🎉
I just have some minor wording and percentage formatting suggestions. Overall, this is fine and the changes look great.

leMaik:
> My only concern is that having a specific number could be useful for debugging in certain circumstances, so it may be worth adding that information elsewhere if it's not already available.

Regarding this concern: Maybe have the full numbers as tooltip? Or maybe after the percentage, like `Rendering - 20% (200 of 1000)`?

chunky/src/java/se/llbit/chunky/renderer/PreviewRenderer.java
imsyphia: I initially chose "Rendering preview" as well, but I chose to use "Displaying preview" to avoid confusion with the actual rendering process, especially for new users. Maybe a different wording could be used?

chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
imsyphia: Yeah, `%f` makes more sense than `%s`.

Personally, I found that adding a decimal digit gave a bit too much visual weight to the element. Also, the progress bar and SPP counter should already give some indication of progress. I'll experiment with it again, though.

imsyphia: Adding replacing the hyphen with an en dash fixes the visual weight issue, so I've gone ahead and added the extra digit.

imsyphia:
Here's what it now looks like, just for reference.

chunkyuiafter3

chunky/src/java/se/llbit/chunky/renderer/PreviewRenderer.java
leMaik: Preparing preview?

leMaik (approved):

PR #1686, opened 2024-01-28
Add support for uncompressed and LZ4-compressed region files

### LINKED ISSUE #1684: Support LZ4 compressed region files
Snapshot [24w04a](https://www.minecraft.net/en-us/article/minecraft-snapshot-24w04a) added LZ4 as new region-file-compression.

The article states that there are 2 possible values: deflate (zlib) and lz4 - we also support gzip, probably from older versions (not sure)?

Relevant code:
- https://github.com/chunky-dev/chunky/blob/master/chunky/src/java/se/llbit/chunky/world/region/MCRegion.java#L347
- https://github.com/chunky-dev/chunky/blob/master/chunky/src/java/se/llbit/chunky/world/ChunkDataSource.java#L58

This feature requires an additional dependency, because Java JDK does not ship LZ4 decoders.
Probably one of those: [LZ4 Java](https://github.com/lz4/lz4-java) or [Kanzi](https://github.com/flanglet/kanzi).

leMaik:
LZ4 support was added in Minecraft snapshots recently, and uncompressed region files (albeit practically unused) were added to Minecraft sometime before 1.16. Since that format is trivial to support, I just added it.

Closes #1684


ShirleyNekoDev (approved):
Looks good 👍

PR #1697, opened 2024-03-04
Do not upload javadoc as artifact in gh actions.

leMaik:
This will speed up the Actions by ~2 minutes. If someone needs the javadoc, they can always build it by themselves.


ThatRedox:
It looks like the gradle task is still building docs. Is this time negligible?

leMaik:
@ThatRedox It takes a few seconds. I updated it to exclude docs, distZip and distTar.

PR #1703, opened 2024-03-10
Add "Quit Chunky" confirmation dialog

### LINKED ISSUE #951: Ask users to save before quitting Chunky
It may be useful to ask the user if they want to save their scene before closing Chunky, in case the scene wasn't saved previously.

ShirleyNekoDev:
image

Requests the user to confirm closing chunky when pressing `Ctrl+Q` (or using `File > Quit`).

This PR __does not__ yet track saved changes. This confirmation dialog is always shown when quitting at the moment.
Further work is needed to skip this dialog when no unsaved changes have been made as described in https://github.com/chunky-dev/chunky/issues/951#issuecomment-1987062142.

I suggest that this closes #951 and we open a separate issue for the work on "dirty state" tracking.


chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
```diff
@@ -734,6 +733,24 @@ public File getSceneFile(String fileName) {
         PersistentSettings.setSppTargetDefault(scene.getTargetSpp()));
   }
 
+  public void confirmAndClose(Event event) {
+    Alert confirmQuit = Dialogs.createAlert(Alert.AlertType.CONFIRMATION);
+    confirmQuit.setTitle("Quit Chunky");
+    confirmQuit.setHeaderText("You have unsaved changes in your scene.");
```
leMaik, line 739: Until we have #1709, I'd suggest reworing this so that users don't think this is a bug:

```suggestion
    confirmQuit.setHeaderText("You may have unsaved changes in your scene.");
```

ShirleyNekoDev: sounds good

PR #1710, opened 2024-03-16
[Launcher] Fix jumbled download status column after updating libraries

ShirleyNekoDev:
chunky-launcher-version-update

Fixes jumbled status cell entries after downloading missing dependencies / updates.
Bug was caused by incorrect use of JavaFX (common mistake, I did the same multiple times).

Also made the launcher launchable as a Gradle task (application plugin) and cached the status icons, instead of loading them on every status change.

### DIFF
--- a/launcher/build.gradle
+++ b/launcher/build.gradle
@@ -1,3 +1,7 @@
+apply plugin: 'application'
+
+mainClassName = 'se.llbit.chunky.launcher.ChunkyLauncher'
+
 configurations {
     bundled
     implementation.extendsFrom configurations.bundled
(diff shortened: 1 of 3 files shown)


Peregrine05:
Does the version need to be bumped?

launcher/build.gradle
ThatRedox, line 1: Undoing these changes fixed the build for me.

leMaik:
Nice! 
A patch bump would be nice (usually i do the bumps but i guess it's fine in this case, the other change was support for sha256 checksums iirc),

I'll merge it once ci is fixed, either by reverting the gradle changes or (preferably) fixing the issue.

leMaik (approved):

PR #1711, opened 2024-03-19
Fix HTTP Redirects in the launcher when fetching JFX versions

ThisIsBrady:
HttpURLConnection does not handle redirects from HTTP to HTTPS. This PR fixes being unable to fetch JavaFX versions from the update site when a redirect across protocols occurs. Handles the HTTP 307 and 308 responses by following the redirect.

This leaves only HTTP 300 and 304 unhandled. 300 could be handled by choosing the Location field if it exists on the response. 304 should never be seen here and would require caching a previous response.


leMaik:
Thank you! :clap: I didn't test this and just assumed that a permanent redirects would just work when I added it to my update site.

PR #1712, opened 2024-03-25
Fix not showing early logged messages in the UI

### LINKED ISSUE #1702: Errors and warnings reported in plugin loading are not presented to the UI as plugins are loaded before the receiver is set.

NotStirred:
I'm not entirely convinced that this is the right approach. See [this short discussion on discord](https://discord.com/channels/541221265512464394/545374333883777037/1215096409481805926), but anything else I come up with is much messier.

This is mainly to fix #1702, but applies to any early log message.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/main/Chunky.java
+++ b/chunky/src/java/se/llbit/chunky/main/Chunky.java
@@ -44,6 +44,7 @@ import se.llbit.chunky.ui.render.RenderControlsTabTransformer;
 import se.llbit.chunky.world.MaterialStore;
 import se.llbit.json.JsonArray;
 import se.llbit.json.JsonValue;
+import se.llbit.log.ConsoleReceiver;
 import se.llbit.log.Level;
 import se.llbit.log.Log;
 import se.llbit.log.Receiver;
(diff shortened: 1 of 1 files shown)


chunky/src/java/se/llbit/chunky/main/Chunky.java
```diff
@@ -237,6 +238,8 @@ public static void main(final String[] args) {
             break;
         }
       } catch (Throwable t) {
+        // set receiver in case an exception was thrown before it was set in one of the start modes.
+        Log.setReceiver(ConsoleReceiver.INSTANCE, Level.INFO, Level.WARNING, Level.ERROR);
```
leMaik, line 242: This would always override the receiver. Can we change it so that it only does if no other receiver is registered? Or maybe just set this receiver as default right at the top of main?

NotStirred: The default receiver is the buffered one which doesn't log anything and buffers the messages until a non-buffering receiver is set. So early messages are logged in the gui and not silently sent to the console.

The only case this `catch` block is if something fails in initialising jfx/headless in which case the respective receiver may not be set or even valid in the case of jfx

leMaik (approved):

Issue #1720, opened 2024-04-07
FirstTimeSetupDialog ignores chunky.home sytem property 


matthew55:
When Chunky is ran with `java -Dchunky.home=/path/to/your/chunky/home` to set `chunky.home` and settings are not found in the specified directory, a FirstTimeSetupDialog option is ran which requires the user to chose a directory for their settings which might not be what they specified through `chunky.home`.

PR #1721, opened 2024-04-09
Fix FirstTimeSetupDialog when chunky.home is set

matthew55:
Solves: #1720

This prevents a FirstTimeSetupDialog GUI from being opened if a `chunky.home` system property is specified by the user.


leMaik (approved):
Thank you for your first PR! Maybe this brings us a step closer to flat pack support. 🎉

Issue #1700, opened 2024-03-06
Semver4j is not included in the README


ThatRedox:
We're using semver4j in the launcher: https://github.com/chunky-dev/chunky/blob/master/launcher/build.gradle#L8
But it is not listed in the 3rd party libraries we use in the README

PR #1722, opened 2024-04-10
Include Semver4j in README

matthew55:
Solves #1700

Lists Semver4j as third party dependency in README and ads Semver4j's MIT license.


leMaik:
Thank you :+1:

PR #1332, opened 2022-07-03
Add support for custom block models

leMaik:
This PR adds support for custom block models. It can (and currently is hard-coded to do so) also replace Chunky's built-in models by models generated from the minecraft jar file.

default_2022-07-03_20-30-16-100 denoised

Caveats:
- Does not support tint colors (yet, we'll have to change the way Chunky handles tint colors)
- Randomized block variants always use the first block variant

To get the latest build of this PR, follow this guide: https://github.com/leMaik/chunky-pr-as-update-site 
Due to a bug, you currently have to open the resource pack dialog and apply the resourcepacks after launching this Chunky build. Then it will load the models. Reload the world if you only see purple blocks.

Fixes #88 and fixes #426.

Issue #751, opened 2020-12-17
PBR Texture


RocketNinja15:
Allows support for PBR texture packs, so 3d textures get rendered

leMaik:
This is a big feature and actually PBR includes multiple features:

- [ ] Normal maps
- [ ] Height maps
- [ ] Emitter maps
- [ ] Specular maps

I have [a branch](https://github.com/leMaik/chunky/tree/json-block-models-and-normal-maps) that supports normal maps, but it's pretty slow. I hope that it'll be ready to be merged some day though. Emitter maps and specular maps should be simple to add then.

Height maps are a whole different topic, but @jackjt8 and me already looked a bit into how it can be done.

PR #1734, opened 2024-06-03
Refactor resource pack loading

leMaik:
For now, this makes loading painting variant and biome registries less hacky. In the future, this enables loading other files, eg. model json files, PBR maps or animation metadata. 

Related to #751, #1332 and #1728


ShirleyNekoDev (commented):
The class LayeredResourcePacks sounds like it implements a similar interfaces as a "ResourcePack", maybe rename this to LayeredResourcePackContainer or something?

Otherwise this looks fine but I'd have to use it to see how I like it ^^

leMaik:
@ShirleyNekoDev Thanks for the review :smile: See #1728 for how it can be used. I renamed this class a few times… But in the end, it's "all resource packs you have" in one object, so I sticked with that name.

PR #1736, opened 2024-06-23
Update to JUnit5, Update to J17 and J21 for tests

NotStirred:
I switched from google truth to assertj, as truth depends heavily on junit4. No code changes thankfully, only imports (apis we use are identical)

Only thing of note: `QuickMathTest#testAbs_1` was previously incorrect (and now fails, relevant part is commented out)

Probably best not to squash as the two commits are entirely separate

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,2 +1,6 @@
+plugins {
+  id 'org.gradle.toolchains.foojay-resolver-convention' version '0.8.0' // java toolchain resolver
+}
+
 rootProject.name = 'chunky'
 include 'chunky', 'lib', 'launcher', 'releasetools'
--- a/build.gradle
+++ b/build.gradle
@@ -31,17 +31,28 @@ subprojects {
     group = 'se.llbit'
     version = rootProject.version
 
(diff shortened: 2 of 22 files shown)


settings.gradle
```diff
@@ -1,2 +1,6 @@
+plugins {
```
leMaik, line 1: What does this plugin do? Maybe add a comment

leMaik (changes requested):

settings.gradle
NotStirred: Ah this is the default toolchain resolver of gradle see [the docs](https://docs.gradle.org/current/userguide/toolchains.html#sub:download_repositories). Will add comment if you'd like

NotStirred:
I moved most things into the subproject closure so it should be much easier to understand. Everything now has junit tests and jacoco coverage enabled.

leMaik:
Maybe we can keep the launcher compatible with Java 8, just in case.

leMaik (approved):
Nice! Next step is increasing the test coverage.

PR #1748, opened 2024-08-11
Skip rendering tiles that only contain sky.

leMaik:
Many scenes contain a lot of sky. Currently, Chunky renders these areas to the full SPP, although their color will never change as we don't have any effects that would apply to a ray that doesn't hit a block.

This PR skips tiles (8x8 pixels by default) that only have immediate sky hits for the first _samples per pass_ (defaults to 1) samples for future samples. This could lead to anti-aliasing errors due to the first pass not hitting a block for a pixel but future passes (for a different sub-pixel )would hit the block. This could be mitigated by increasing _spp per pass_.

For example in the following scene, the greyed out pixels would only be rendered once whereas all other areas are rendered as usual.
image

Depending on the amount of sky in the scene, this can decrease the rendering time. Needs benchmarks, though.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/PathTracingRenderer.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/PathTracingRenderer.java
@@ -66,6 +66,8 @@ public class PathTracingRenderer extends TileBasedRenderer {
     double halfWidth = fullWidth / (2.0 * fullHeight);
     double invHeight = 1.0 / fullHeight;
 
+    resetTiles(manager); // reset previously skipped tiles
+
     double[] sampleBuffer = scene.getSampleBuffer();
 
     while (scene.spp < scene.getTargetSpp()) {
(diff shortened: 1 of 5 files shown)


JustinTimeCuber:
Might be a minor thing and difficult to notice in most situations, but couldn't this cause aliasing effects in the sky? This would probably be noticeable when using a high resolution skymap but a lower resolution canvas

Peregrine05:
I wonder if it would be good to add a separate `Sky target SPP` control that sets the target SPP for sky-only tiles.

jackjt8:
> we don't have any effects that would apply to a ray that doesn't hit a block

Depth of Field **does** apply to the sky. It's not noticeable when using a generated sky, and we could probably skip sampling in those cases... However if using a skymap (or any texture) DoF is visible. I use DoF to mask low quality skymaps all the time.

Low SPP example...

image

PR #1749, opened 2024-08-11
Re-use worker state and ray while rendering.

leMaik:
Needs benchmarking but feels like it's faster.


leMaik:
After benchmarking, this appears to be slower consistently.

PR #1750, opened 2024-08-19
Fix build issues on windows due to azul not including JFX on windows

NotStirred:
Neither oracle nor azul include jfx on both windows & linux in their java 8 jdks.

Corretto documentation says that its [mac jdk doesn't include jfx](https://aws.amazon.com/corretto/faqs/), so we still have a build problem on apple systems

If someone can tell me a mac jdk8 that includes jfx I'll PR another fix.


jackjt8:
I can confirm that Chunky now builds again on Windows. Just seeing some unmappable characters.. but nothing that stops building (and the build works fine).

PR #1762, opened 2024-09-06
Add new cocoa model

### LINKED ISSUE #1761: The cocoa model is incorrect.
As shown in the following screenshot, the cocoa model is incorrect.
image

This issue is also present in 2.4.6, but it manifests differently (note the double border on the outer edge, and the lack of a border on the inner edge.)
image

leMaik:
Turns out the model for stage 2 was changed a while ago.

Fixes #1761

### DIFF
--- a/chunky/src/java/se/llbit/chunky/block/minecraft/Cocoa.java
+++ b/chunky/src/java/se/llbit/chunky/block/minecraft/Cocoa.java
@@ -46,7 +46,7 @@ public class Cocoa extends AbstractModelBlock {
         facing = 3;
         break;
     }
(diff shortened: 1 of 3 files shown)


NotStirred:
thoughts on grouping all of the old models under chunky.oldModels.cocoa?

nicer for users and easier for us to search for if we ever refactor this to be an option in the UI

leMaik:
@NotStirred I switched to a more flexible format: `chunky.blockModels.cocoa=1.19` or `chunky.blockModels.cocoa=pre-1.19` etc.

PR #1765, opened 2024-09-17
Fix negative scene minY inconsistent behaviour

### LINKED ISSUE #1764: Negative yMinClip with numbers that are not divisible by 16 cause chunk loading errors
Possibly related to #1214 

The bottom-most blocks are not loaded in this case and the chunks "hover" over the grid. Could be a negative floor to zero error.

NotStirred:
Closes #1764


NotStirred:
Oh our gh action workflow is deprecated

PR #1768, opened 2024-09-19
Fix blocks overflowing over the top of octree bounds

NotStirred:
Due to an incorrect maxDimension calculation

This happened because we were trying to set a cube of height 16 less than 16 blocks from the edge of the octree. (which was then masked to within bounds and so wrapped)
Thoughts on adding some assertions in packedoctree that the arguments are even valid?

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -1443,18 +1443,23 @@ public class Scene implements JsonSerializable, Refreshable {
     zmin *= 16;
     zmax *= 16;
 
-    int maxDimension = Math.max(yMax - yMin, Math.max(xmax - xmin, zmax - zmin));
-    int requiredDepth = QuickMath.log2(QuickMath.nextPow2(maxDimension));
-
+    int requiredDepth;
     if(centerOctree) {
+      int maxDimension = Math.max(yMax - yMin, Math.max(xmax - xmin, zmax - zmin));
+      requiredDepth = QuickMath.log2(QuickMath.nextPow2(maxDimension));
+
       int xroom = (1 << requiredDepth) - (xmax - xmin);
       int yroom = (1 << requiredDepth) - (yMax - yMin);
       int zroom = (1 << requiredDepth) - (zmax - zmin);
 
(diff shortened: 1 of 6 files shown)


leMaik:
Assertions would be great, but why in PackedOctree? Why not throw some IllegalArgumentExceptions in Octree to fix these bugs fast? Or at least warn and tell the user to report a bug?

NotStirred:
:+1: I'd forgotten that the octree wrapper existed

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -1443,18 +1443,23 @@ private int calculateOctreeOrigin(Collection<ChunkPosition> chunksToLoad, boolea
...
+
       int xroom = (1 << requiredDepth) - (xmax - xmin);
       int yroom = (1 << requiredDepth) - (yMax - yMin);
       int zroom = (1 << requiredDepth) - (zmax - zmin);
 
       origin.set(xmin - xroom / 2, -yroom / 2, zmin - zroom / 2);
     } else {
-      // Note: Math.floorDiv rather than integer division for round toward -infinity
-      origin.set(xmin, Math.floorDiv(yMin, 16) * 16, zmin);
+      int yMin16 = yMin & ~0xf; // closest multiple of 16 below yMin
```
leMaik, line 1457: Does this actually work for negative numbers? Maybe I'm just not that fluent in bit operations, but how about moving this to a static method with some unit tests?

NotStirred: it does yes, two's complement helps us here

A helper and some tests are definitely a good idea :+1:

leMaik (approved):

NotStirred:
Bounds on the assertions are off-by-one.
if the octree is 32 size, and we're placing a 16 size cube at 16,0,0 it's checking `16+16 < 32` which is obviously not right
needs to be `<=`

NotStirred:
Thoughts on splitting `calculateOctreeOrigin` into two methods, one for legacy one for current?

Also this method is pretty horrible, internally modifies `origin` but returns `requiredDepth` with a boolean flag for legacy path

leMaik:
@NotStirred Splitting `calculateOctreeOrigin` or refactoring it not to be horrible sounds good, probably something for another PR. :+1:

PR #1770, opened 2024-09-23
Use y-clipping settings from the map view when creating a new scene or loading chunks from it

### LINKED ISSUE #1741: Fix y clipping defaults
We very frequently get questions about chunks seemingly not loading due to y clipping not being set correctly.

The problem could be the default values that are saved as a persistent setting – we should drop them imho. Instead, we should *always* use the values specified in the map view to avoid surprises.

Examples:
* https://old.reddit.com/r/chunky/comments/1dw3r8j/need_help_chunky_doesnt_load_chunks/
* https://old.reddit.com/r/chunky/comments/19c2pvb/do_you_know_guys_why_my_chunky_dont_render_above/
* https://old.reddit.com/r/chunky/comments/xmdwcv/any_idea_why_the_second_half_of_my_build_isnt/

leMaik:
Also fixes requiring two reloads after changing the y-clipping settings of a scene.

Closes #1741

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/SynchronousSceneManager.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/SynchronousSceneManager.java
@@ -235,15 +235,21 @@ public class SynchronousSceneManager implements SceneProvider, SceneManager {
 
   @Override public void loadFreshChunks(World world, Collection<ChunkPosition> chunksToLoad) {
     synchronized (scene) {
+      int yClipMin = scene.getYClipMin();
+      int yClipMax = scene.getYClipMax();
       scene.clear();
-      scene.loadChunks(taskTracker, world, chunksToLoad);
       scene.resetScene(null, context.getChunky().getSceneFactory());
+      scene.setYClipMin(yClipMin);
+      scene.setYClipMax(yClipMax);
       context.setSceneDirectory(new File(context.getChunky().options.sceneDir, scene.name));
+      scene.loadChunks(taskTracker, world, chunksToLoad);
       scene.refresh();
       scene.setResetReason(ResetReason.SCENE_LOADED);
       scene.setRenderMode(RenderMode.PREVIEW);
     }
+    System.out.println(scene.getYClipMin()+" min, 2 max "+scene.getYClipMax());
     onSceneLoaded.run();
+    System.out.println(scene.getYClipMin()+" min, 3 max "+scene.getYClipMax());
   }
 
   @Override public void loadChunks(World world, Collection<ChunkPosition> chunksToLoad) {
@@ -253,9 +259,9 @@ public class SynchronousSceneManager implements SceneProvider, SceneManager {
       if (prevChunkCount == 0) {
         scene.moveCameraToCenter();
       }
-      scene.refresh();
       scene.setResetReason(ResetReason.SCENE_LOADED);
       scene.setRenderMode(RenderMode.PREVIEW);
+      scene.refresh();
     }
     onChunksLoaded.run();
   }
@@ -263,9 +269,9 @@ public class SynchronousSceneManager implements SceneProvider, SceneManager {
   @Override public void reloadChunks() {
     synchronized (scene) {
       scene.reloadChunks(taskTracker);
-      scene.refresh();
       scene.setResetReason(ResetReason.SCENE_LOADED);
       scene.setRenderMode(RenderMode.PREVIEW);
+      scene.refresh();
     }
     onChunksLoaded.run();
   }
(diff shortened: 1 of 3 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/SynchronousSceneManager.java
```diff
@@ -235,15 +235,21 @@ public void saveScene(SceneIOProvider ioContext, Scene scene) throws Interrupted
...
       scene.resetScene(null, context.getChunky().getSceneFactory());
+      scene.setYClipMin(yClipMin);
+      scene.setYClipMax(yClipMax);
       context.setSceneDirectory(new File(context.getChunky().options.sceneDir, scene.name));
+      scene.loadChunks(taskTracker, world, chunksToLoad);
       scene.refresh();
       scene.setResetReason(ResetReason.SCENE_LOADED);
       scene.setRenderMode(RenderMode.PREVIEW);
     }
+    System.out.println(scene.getYClipMin()+" min, 2 max "+scene.getYClipMax());
```
Peregrine05, line 250: Remove debug code.

chunky/src/java/se/llbit/chunky/renderer/scene/SynchronousSceneManager.java
```diff
@@ -235,15 +235,21 @@ public void saveScene(SceneIOProvider ioContext, Scene scene) throws Interrupted
...
     onSceneLoaded.run();
+    System.out.println(scene.getYClipMin()+" min, 3 max "+scene.getYClipMax());
```
leMaik, line 252: More debug code

Peregrine05 (commented):
Things seem slightly confusing now.

Chunky appears to use the Map Y clip when `Load selected chunks` or `New scene from selection` in the `Map` tab context menu is used, but it uses the Scene Y clip when `Load selected chunks` or `Reload chunks` in the `Scene` tab is used. Additionally, changing the Map Y clip while the `Scene` tab is open causes the Scene Y clip values to apparently reset to their default values; however, the internal values are not changed, and the UI values are set properly once it refreshes.

I think that either there should be one set of sliders that controls both the map view and the scene clip, or there should be two separate sets of sliders, as Chunky was previously, but with a note that points users to the `Scene` tab to change the Scene Y clip.

As I recall, the main issue was that the Scene Y clip values were not being set properly, which resulted in portions of the world not being loaded. Users would see the Map Y clip controls, and assume that they changed the Scene Y clip, so a note next to them, I believe, would be effective.

leMaik:
> Chunky appears to use the Map Y clip when Load selected chunks or New scene from selection in the Map tab context menu is used, but it uses the Scene Y clip when Load selected chunks or Reload chunks in the Scene tab is used.

@Peregrine05 This is exactly the intended behavior. I do see that it might be confusing if you _create_ a scene from the left panel – I didn't consider that.

> Additionally, changing the Map Y clip while the Scene tab is open causes the Scene Y clip values to apparently reset to their default values; however, the internal values are not changed, and the UI values are set properly once it refreshes.

That is not supposed to happen… I'll look into this. The scene tab is supposed to represent the scene state, regardless of whatever the map view does.

> Users would see the Map Y clip controls, and assume that they changed the Scene Y clip, so a note next to them, I believe, would be effective.

Users would expect that the y clip controls from the map are applied when creating a new scene – which is what this PR does. After a scene has been created, the y clip (as well as any other scene settings) can be controlled in the scene tab.

> two separate sets of sliders, as Chunky was previously, but with a note that points users to the Scene tab to change the Scene Y clip.

A UI that needs a note to explain how it works is the worst solution we could find.

leMaik:
image

@Peregrine05 What do you think? :thinking:

Peregrine05:
Ah; so how it works now is:

`Map right-click > New scene from selection`: Use Map Y clip
`Map right-click > Load selected chunks` (First usage, equivalent to `New scene from selection): Use Map Y clip
`Scene tab > Load selected chunks`: Use Scene Y clip

If so, I think should work well for the moment (a dialog box might be a better solution in the long run, but a dialog box also requires more user input to get the same job done).

leMaik:
@Peregrine05 Exactly :100: 

Regarding the dialog box: Yes, I also considered that, but the hint is mostly for first time users who don't know about the two sliders yet. Once they know, there's no reason to bother them with a dialog. But the content of a dialog with a "Don't show this again" checkbox might be forgotten at some point.

Let's roll with this and see if the y clip confusion is reduced. :rocket: 

Also regarding "y clip": Should we consistently rename this to "Y-clip"? Same for -axis etc.?

PR #1771, opened 2024-09-24
Show time elapsed for all tasks and use Duration for the duration and eta.

### LINKED ISSUE #1754: Time elapsed for an operation not shown
The title says it all. The time elapsed when **Loading Chunks** and **Finalizing Octree** is not displayed.
However, the time elapsed when **Rendering** is.

During chunk loading:
image

During octree creation:
image

During rendering:
image

leMaik:
Closes #1754

### DIFF
--- a/chunky/src/java/se/llbit/util/ProgressListener.java
+++ b/chunky/src/java/se/llbit/util/ProgressListener.java
@@ -17,23 +17,25 @@
  */
 package se.llbit.util;
 
+import java.time.Duration;
+
 /**
  * Progress listener.
  *
  * @author Jesper Öqvist <jesper@llbit.se>
  */
 public interface ProgressListener {
-  ProgressListener NONE = (task, done, start, target) -> {};
+  ProgressListener NONE = (task, done, start, target, duration) -> {};
 
   /**
    * Update progress without ETA.
    */
-  void setProgress(String task, int done, int start, int target);
+  void setProgress(String task, int done, int start, int target, Duration duration);
 
   /**
    * Update progress with ETA.
    */
-  default void setProgress(String task, int done, int start, int target, String eta) {
-    setProgress(task, done, start, target);
+  default void setProgress(String task, int done, int start, int target, Duration duration, Duration eta) {
+    setProgress(task, done, start, target, duration);
   }
 }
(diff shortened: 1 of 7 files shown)


chunky/src/java/se/llbit/util/ProgressListener.java
```diff
@@ -17,23 +17,25 @@
...
    */
-  void setProgress(String task, int done, int start, int target);
+  void setProgress(String task, int done, int start, int target, Duration duration);
 
   /**
    * Update progress with ETA.
    */
-  default void setProgress(String task, int done, int start, int target, String eta) {
-    setProgress(task, done, start, target);
+  default void setProgress(String task, int done, int start, int target, Duration duration, Duration eta) {
```
ShirleyNekoDev, line 38: Same here - is eta "Estimated Time of Arrival" the remaining duration or the total estimated duration of the task?
```suggestion
  default void setProgress(String task, int done, int start, int target, Duration duration, Duration remainingTime) {
```

ShirleyNekoDev (changes requested):
Aside from naming, I really like the use of Duration here :)

chunky/src/java/se/llbit/util/ProgressListener.java
```diff
@@ -17,23 +17,25 @@
...
  */
 public interface ProgressListener {
-  ProgressListener NONE = (task, done, start, target) -> {};
+  ProgressListener NONE = (task, done, start, target, duration) -> {};
 
   /**
    * Update progress without ETA.
    */
-  void setProgress(String task, int done, int start, int target);
+  void setProgress(String task, int done, int start, int target, Duration duration);
```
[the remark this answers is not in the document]
leMaik: Good point! I'll rename it to `elapsedTime` which should make it clear without requiring any javadoc.

chunky/src/java/se/llbit/util/ProgressListener.java
leMaik: I even considered changing this to an `Instant` that has the actual _time_ due to the strict meaning of "ETA" but then decided that we never show the actual time and if we do, we can just calculate it from the remaining time. 

Renaming is a great idea, as above.

ShirleyNekoDev (approved):

PR #1772, opened 2024-09-24
Replace StringUtil.join with String.join.

leMaik:


NotStirred (approved):
:+1: why did this ever exist? Was `join` added in 1.8 or something?

leMaik:
Yes, it was indeed added in 1.8

PR #1795, opened 2024-12-07
Registerable WorldFormat api

NotStirred:
I mainly want comments on general structure, everything else is bikesheddable. I'll go over everything again and polish it up before review.

Lots of changes here, so far I've only looked at loading worlds from the gui, haven't looked at reloading from the previous 
session

`Chunk`, `World`, `Dimension` are now abstract classes, the previous implementation is now `JavaChunk` etc.
there is also a skeleton of registerable `WorldFormat`s. Each is asked if a directory is valid for it, if the user selects it, the format is asked to load the world.

The ultimate goal would probably be to have `Chunk` be an implementation detail of java world, but that's a much larger task as it would require a rewrite of parts of the map view and scene chunk loading

### DIFF
--- a/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
+++ b/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
@@ -85,9 +85,9 @@ public class WorldMapLoader implements ChunkTopographyListener, ChunkViewListene
 
     Optional<Dimension.Identifier> dimensionToLoad = Optional.of(world.currentDimension())
       .map(Dimension::getDimensionId)
-      .filter(dimension -> newWorld.availableDimensions().contains(dimension))
-      .or(newWorld::defaultDimension)
-      .or(() -> newWorld.availableDimensions().stream().findFirst());
+      .filter(dimension -> newWorld.getAvailableDimensions().contains(dimension))
+      .or(newWorld::getDefaultDimension)
+      .or(() -> newWorld.getAvailableDimensions().stream().findFirst());
 
     if (dimensionToLoad.isEmpty()) {
       Log.infof("No dimension loaded for world %s", newWorld.toString());
@@ -192,7 +192,7 @@ public class WorldMapLoader implements ChunkTopographyListener, ChunkViewListene
   /**
    * Set the current dimension.
    *
-   * @param value Must be a valid dimension see {@link World#availableDimensions()}
+   * @param value Must be a valid dimension see {@link World#getAvailableDimensions()}
    */
   public void setDimension(Dimension.Identifier value) {
     if (value != currentDimensionId) {
--- a/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
+++ b/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
@@ -22,12 +22,11 @@ import se.llbit.chunky.PersistentSettings;
 import se.llbit.chunky.renderer.ChunkViewListener;
 import se.llbit.chunky.ui.controller.ChunkyFxController;
 import se.llbit.chunky.world.*;
-import se.llbit.chunky.world.java.JavaWorld;
 import se.llbit.chunky.world.region.RegionChangeWatcher;
 import se.llbit.chunky.world.region.RegionParser;
 import se.llbit.chunky.world.region.RegionQueue;
 import se.llbit.chunky.world.listeners.ChunkTopographyListener;
-import se.llbit.chunky.world.worldformat.WorldFormat;
+import se.llbit.chunky.world.worldformat.WorldFormats;
 import se.llbit.log.Log;
 import se.llbit.util.annotation.Nullable;
 
@@ -73,7 +72,7 @@ public class WorldMapLoader implements ChunkTopographyListener, ChunkViewListene
     if (worldLocation == null) {
       return;
     }
-    this.loadWorld(WorldFormat.loadWorld(worldLocation).orElse(EmptyWorld.INSTANCE));
+    this.loadWorld(WorldFormats.createWorld(worldLocation).orElse(EmptyWorld.INSTANCE));
   }
   /**
    * This is called when a new world is loaded
--- a/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
+++ b/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
@@ -21,6 +21,7 @@ import se.llbit.chunky.PersistentSettings;
 import se.llbit.chunky.renderer.ChunkViewListener;
 import se.llbit.chunky.ui.controller.ChunkyFxController;
 import se.llbit.chunky.world.*;
+import se.llbit.chunky.world.java.JavaWorld;
 import se.llbit.chunky.world.region.RegionChangeWatcher;
 import se.llbit.chunky.world.region.RegionParser;
 import se.llbit.chunky.world.region.RegionQueue;
@@ -68,7 +69,7 @@ public class WorldMapLoader implements ChunkTopographyListener, ChunkViewListene
    * This is called when a new world is loaded
    */
   public void loadWorld(File worldDir) {
-    if (World.isWorldDir(worldDir)) {
+    if (JavaWorld.isWorldDir(worldDir)) {
       if (world != null) {
         world.currentDimension().removeChunkTopographyListener(this);
       }
--- a/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
+++ b/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
@@ -16,6 +16,7 @@
  */
 package se.llbit.chunky.map;
 
+import java.util.Optional;
 import java.util.function.BiConsumer;
 import se.llbit.chunky.PersistentSettings;
 import se.llbit.chunky.renderer.ChunkViewListener;
@@ -26,6 +27,9 @@ import se.llbit.chunky.world.region.RegionChangeWatcher;
 import se.llbit.chunky.world.region.RegionParser;
 import se.llbit.chunky.world.region.RegionQueue;
 import se.llbit.chunky.world.listeners.ChunkTopographyListener;
+import se.llbit.chunky.world.worldformat.WorldFormat;
+import se.llbit.log.Log;
+import se.llbit.util.annotation.Nullable;
 
 import java.io.File;
 import java.util.ArrayList;
@@ -65,28 +69,44 @@ public class WorldMapLoader implements ChunkTopographyListener, ChunkViewListene
     topographyUpdater.start();
   }
 
+  public void loadWorldFromDirectory(@Nullable File worldLocation) {
+    if (worldLocation == null) {
+      return;
+    }
+    this.loadWorld(WorldFormat.loadWorld(worldLocation).orElse(EmptyWorld.INSTANCE));
+  }
   /**
    * This is called when a new world is loaded
    */
-  public void loadWorld(File worldDir) {
-    if (JavaWorld.isWorldDir(worldDir)) {
-      if (world != null) {
-        world.currentDimension().removeChunkTopographyListener(this);
-      }
-      boolean isSameWorld = !(world instanceof EmptyWorld) && worldDir.equals(world.getWorldDirectory());
-      World newWorld = World.loadWorld(worldDir, currentDimensionId, World.LoggedWarnings.NORMAL);
-      newWorld.currentDimension().addChunkTopographyListener(this);
-      synchronized (this) {
-        world = newWorld;
-        updateRegionChangeWatcher(newWorld.currentDimension());
-
-        File newWorldDir = world.getWorldDirectory();
-        if (newWorldDir != null && !newWorldDir.equals(PersistentSettings.getLastWorld())) {
-          PersistentSettings.setLastWorld(newWorldDir);
-        }
-      }
-      worldLoadListeners.forEach(listener -> listener.accept(newWorld, isSameWorld));
+  public void loadWorld(World newWorld) {
+    if (this.world != null) {
+      this.world.currentDimension().removeChunkTopographyListener(this);
     }
+    boolean isSameWorld = !(this.world instanceof EmptyWorld) && newWorld.getWorldDirectory().equals(this.world.getWorldDirectory());
+
+    Optional<Dimension.Identifier> dimensionToLoad = Optional.of(world.currentDimension())
+      .map(Dimension::getDimensionId)
+      .filter(dimension -> newWorld.availableDimensions().contains(dimension))
+      .or(newWorld::defaultDimension)
+      .or(() -> newWorld.availableDimensions().stream().findFirst());
+
+    if (dimensionToLoad.isEmpty()) {
+      Log.infof("No dimension loaded for world %s", newWorld.toString());
+      return;
+    }
+
+    Dimension loadedDim = newWorld.loadDimension(dimensionToLoad.get());
+    loadedDim.addChunkTopographyListener(this);
+    synchronized (this) {
+      this.world = newWorld;
+      updateRegionChangeWatcher(loadedDim);
+
+      File newWorldDir = this.world.getWorldDirectory();
+      if (!newWorldDir.equals(PersistentSettings.getLastWorld())) {
+        PersistentSettings.setLastWorld(newWorldDir);
+      }
+    }
+    worldLoadListeners.forEach(listener -> listener.accept(newWorld, isSameWorld));
   }
 
   /**
@@ -152,14 +172,12 @@ public class WorldMapLoader implements ChunkTopographyListener, ChunkViewListene
   public void reloadWorld() {
     topographyUpdater.clearQueue();
     world.currentDimension().removeChunkTopographyListener(this);
-    World newWorld = World.loadWorld(world.getWorldDirectory(), currentDimensionId,
-        World.LoggedWarnings.NORMAL);
-    newWorld.currentDimension().addChunkTopographyListener(this);
+    world.loadDimension(currentDimensionId);
+    world.currentDimension().addChunkTopographyListener(this);
     synchronized (this) {
-      world = newWorld;
-      updateRegionChangeWatcher(newWorld.currentDimension());
+      updateRegionChangeWatcher(world.currentDimension());
     }
-    worldLoadListeners.forEach(listener -> listener.accept(newWorld, true));
+    worldLoadListeners.forEach(listener -> listener.accept(world, true));
     viewUpdated(mapView.getMapView()); // update visible chunks immediately
   }
 
@@ -174,6 +192,8 @@ public class WorldMapLoader implements ChunkTopographyListener, ChunkViewListene
 
   /**
    * Set the current dimension.
+   *
+   * @param value Must be a valid dimension see {@link World#availableDimensions()}
    */
   public void setDimension(Dimension.Identifier value) {
     if (value != currentDimensionId) {
--- a/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
+++ b/chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
@@ -22,6 +22,7 @@ import se.llbit.chunky.PersistentSettings;
 import se.llbit.chunky.renderer.ChunkViewListener;
 import se.llbit.chunky.ui.controller.ChunkyFxController;
 import se.llbit.chunky.world.*;
+import se.llbit.chunky.world.java.JavaWorldFormat;
 import se.llbit.chunky.world.region.RegionChangeWatcher;
 import se.llbit.chunky.world.region.RegionParser;
 import se.llbit.chunky.world.region.RegionQueue;
@@ -68,20 +69,27 @@ public class WorldMapLoader implements ChunkTopographyListener, ChunkViewListene
     topographyUpdater.start();
   }
 
-  public void loadWorldFromDirectory(@Nullable File worldLocation) {
-    if (worldLocation == null) {
+  public void loadWorldFromDirectory(@Nullable File worldLocation, @Nullable String worldFormatId) {
+    if (worldLocation != null) {
+      if (worldFormatId == null || worldFormatId.isEmpty()) {
+        worldFormatId = JavaWorldFormat.ID;
+      }
+      Optional<World.Info> info = WorldFormats.getWorldFormat(worldFormatId)
+        .flatMap(format -> format.getWorldInfo(worldLocation.toPath())) // attempt to get the given format
+        .or(() -> WorldFormats.getInfos(worldLocation.toPath()).stream().findFirst()); // get any format
+      info.ifPresent(this::loadWorld);
       return;
     }
-    this.loadWorld(WorldFormats.createWorld(worldLocation).orElse(EmptyWorld.INSTANCE));
+    setWorld(EmptyWorld.INSTANCE);
   }
+
   /**
-   * This is called when a new world is loaded
+   * Load the world referred to by the {@link World.Info}
+   *
+   * @return The loaded world. May be {@link EmptyWorld} if loading failed.
    */
-  public void loadWorld(World newWorld) {
-    if (this.world != null) {
-      this.world.currentDimension().removeChunkTopographyListener(this);
-    }
-    boolean isSameWorld = !(this.world instanceof EmptyWorld) && newWorld.getWorldDirectory().equals(this.world.getWorldDirectory());
+  public World loadWorld(World.Info info) {
+    World newWorld = WorldFormats.createWorld(info);
 
     Optional<Dimension.Identifier> dimensionToLoad = Optional.of(world.currentDimension())
       .map(Dimension::getDimensionId)
@@ -89,21 +97,45 @@ public class WorldMapLoader implements ChunkTopographyListener, ChunkViewListene
       .or(newWorld::getDefaultDimension)
       .or(() -> newWorld.getAvailableDimensions().stream().findFirst());
 
-    if (dimensionToLoad.isEmpty()) {
-      Log.infof("No dimension loaded for world %s", newWorld.toString());
-      return;
+    if (dimensionToLoad.isPresent()) {
+      newWorld.loadDimension(dimensionToLoad.get());
+    } else {
+      Log.infof("No dimension loaded for world %s", info.toString());
     }
 
-    Dimension loadedDim = newWorld.loadDimension(dimensionToLoad.get());
+    setWorld(newWorld);
+    return this.world;
+  }
+
+  /**
+   * Sets the map view world.
+   * <p>This is intended to be called with worlds with a dimension already loaded, as it will not trigger dimension
+   * loading.</p>
+   *
+   * @param newWorld The world to set
+   */
+  public void setWorld(World newWorld) {
+    if (this.world != null) {
+      this.world.currentDimension().removeChunkTopographyListener(this);
+    }
+
+    boolean isSameWorld = !(this.world instanceof EmptyWorld) && newWorld.getInfo().path().equals(this.world.getInfo().path());
+
+    Dimension loadedDim = newWorld.currentDimension();
+    if (loadedDim == EmptyWorld.INSTANCE.currentDimension()) {
+      Log.warn("Map view world was set but it has no dimension!");
+    }
+
     loadedDim.addChunkTopographyListener(this);
     synchronized (this) {
       this.world = newWorld;
       updateRegionChangeWatcher(loadedDim);
 
-      File newWorldDir = this.world.getWorldDirectory();
+      File newWorldDir = this.world.getInfo().path().toFile();
       if (!newWorldDir.equals(PersistentSettings.getLastWorld())) {
         PersistentSettings.setLastWorld(newWorldDir);
       }
+      PersistentSettings.setLastWorldFormat(newWorld.getInfo().worldFormat().getId());
     }
     worldLoadListeners.forEach(listener -> listener.accept(newWorld, isSameWorld));
   }
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -53,6 +53,7 @@ import se.llbit.chunky.world.biome.ArrayBiomePalette;
 import se.llbit.chunky.world.biome.Biome;
 import se.llbit.chunky.world.biome.BiomePalette;
 import se.llbit.chunky.world.biome.Biomes;
+import se.llbit.chunky.world.java.JavaWorld;
 import se.llbit.chunky.world.region.MCRegion;
 import se.llbit.json.*;
 import se.llbit.log.Log;
@@ -547,7 +548,7 @@ public class Scene implements JsonSerializable {
       loadedWorld = EmptyWorld.INSTANCE;
       if (!worldPath.isEmpty()) {
         File worldDirectory = new File(worldPath);
-        if (World.isWorldDir(worldDirectory)) {
+        if (JavaWorld.isWorldDir(worldDirectory)) {
           loadedWorld = World.loadWorld(worldDirectory, worldDimension, World.LoggedWarnings.NORMAL);
         } else {
           Log.info("Could not load world: " + worldPath);
--- a/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
@@ -53,8 +53,8 @@ import se.llbit.chunky.world.biome.ArrayBiomePalette;
 import se.llbit.chunky.world.biome.Biome;
 import se.llbit.chunky.world.biome.BiomePalette;
 import se.llbit.chunky.world.biome.Biomes;
+import se.llbit.chunky.world.region.Region;
 import se.llbit.chunky.world.worldformat.WorldFormats;
-import se.llbit.chunky.world.region.MCRegion;
 import se.llbit.json.*;
 import se.llbit.log.Log;
 import se.llbit.math.*;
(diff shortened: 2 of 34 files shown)


chunky/src/java/se/llbit/chunky/chunk/biome/BiomeData2d.java
```diff
@@ -34,7 +35,7 @@ public void clear() {
   }
 
   public static void loadBiomeDataByteArray(Tag chunkData, BiomeData2d biomeData, BiomePalette biomePalette) {
```
leMaik, line 37: Should this be moved to the world? Seems to be world format / chunk data dependant.

chunky/src/java/se/llbit/chunky/chunk/biome/BiomeData2d.java
```diff
@@ -46,7 +47,7 @@ public static void loadBiomeDataByteArray(Tag chunkData, BiomeData2d biomeData,
 
   public static void loadBiomeDataIntArray(Tag chunkData, BiomeData2d biomeData, BiomePalette biomePalette) {
```
leMaik, line 48: Same here, ie. let the biome data classes only contain the _data_ and add per world format factories to create them?

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -821,7 +821,7 @@ public synchronized void loadChunks(TaskTracker taskTracker, World world, Collec
       }
 
       for (RegionPosition region : regions) {
-        dimension.getRegion(region).parse(yMin, yMax);
+        ((JavaDimension) dimension).getRegion(region).parse(yMin, yMax);
```
leMaik, line 824: Since regions are handy for selecting many chunks, could we add virtual regions (= squares of many chunks) to Bedrock? As a UI feature, not as in actually loading regions, if the world format doesn't actually have regions.

chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
```diff
@@ -482,19 +483,15 @@ public File getSceneFile(String fileName) {
...
-                    yMax.set(320);
-                    mapView.setYMinMax(-64, 320);
-                  } else {
-                    yMin.setRange(0, 256);
-                    yMin.set(0);
-                    yMax.setRange(0, 256);
-                    yMax.set(256);
-                    mapView.setYMinMax(0, 256);
-                  }
+                  IntIntPair heightRange = mapLoader.getWorld().currentDimension().heightRange();
```
leMaik, line 486: Would `getMinY()` and `getMaxY()` be a simpler interface?

chunky/src/java/se/llbit/chunky/ui/controller/WorldChooserController.java
```diff
@@ -186,17 +188,25 @@ private void fillWorldList(final File worldSavesDir) {
...
-                    World.LoggedWarnings.SILENT));
+              for (WorldFormat worldFormat : WorldFormat.worldFormats) {
+                if (worldFormat.isValid(dir.toPath())) {
+                  try {
+                    World world = worldFormat.loadWorld(dir.toPath(), String.valueOf(World.OVERWORLD_DIMENSION));
+                    if (world != EmptyWorld.INSTANCE) {
+                      worlds.add(world);
+                    }
+                  } catch (IOException e) {
+                    throw new RuntimeException(e);
```
leMaik, line 207: Log this instead? Listing worlds shouldn't crash Chunky just because there's a bad world.

chunky/src/java/se/llbit/chunky/world/Dimension.java
```diff
@@ -202,43 +114,54 @@ public Heightmap getHeightmap() {
...
-    synchronized (this) {
-      regionMap.computeIfAbsent(pos.getLong(), p -> createRegion(pos));
+  /**
+   * @return A user presentable name of the dimension
+   */
+  public abstract String toString();
+
+  /** Add a chunk deletion listener. */
+  public void addChunkDeletionListener(ChunkDeletionListener listener) {
+    synchronized (chunkDeletionListeners) {
```
leMaik, line 124: Could we use some concurrent collection instead of synchronizing?

chunky/src/java/se/llbit/chunky/world/EmptyWorld.java
```diff
@@ -31,8 +38,23 @@ private EmptyWorld() {
     this.currentDimension = EmptyDimension.INSTANCE;
   }
 
+  @Override
+  public Set<Integer> listDimensions() {
```
leMaik, line 42: Why would dimensions be integers? Is that a sane assumption or just some Java detail?

chunky/src/java/se/llbit/chunky/world/JavaChunk.java
```diff
@@ -0,0 +1,377 @@
...
+
+  public JavaChunk(ChunkPosition pos, JavaDimension dimension) {
+    super(pos, dimension);
+  }
+
+  /**
+   * @param request fresh request set
+   * @return loaded data, or null if something went wrong
+   */
+  private Map<String, Tag> getChunkTags(Set<String> request) throws ChunkLoadingException {
```
leMaik, line 61: Off topic: I never really understood why we had this request/response interface for getting chunk tags.

chunky/src/java/se/llbit/chunky/world/JavaChunk.java
```diff
@@ -0,0 +1,377 @@
...
+        if (sectionY < minSectionY) {
+          minSectionY = sectionY;
+        }
+        if (sectionY > maxSectionY) {
+          maxSectionY = sectionY;
+        }
+      }
+    }
+
+    return new IntIntImmutablePair(minSectionY << 4, (maxSectionY << 4) + 15);
```
leMaik, line 367: Oh, I see… Getting the range requires the same logic internally for min and max. Fine, although the `IntIntImmutablePair` feels a bit too generic, but I'm not sure if adding a `Range` class is worth it (unless that would have more useful methods).

chunky/src/java/se/llbit/chunky/world/JavaDimension.java
```diff
@@ -0,0 +1,138 @@
...
+   * @return {@code true} if player data was reloaded.
+   */
+  public synchronized boolean reloadPlayerData() {
+    return ((JavaWorld) this.world).reloadPlayerData();
+  }
+
+  /**
+   * @return The chunk at the given position
+   */
+  public synchronized Chunk getChunk(ChunkPosition pos) {
```
leMaik, line 38: Does this method need to be synchronized? It only calls other methods.

chunky/src/java/se/llbit/chunky/world/JavaDimension.java
```diff
@@ -0,0 +1,138 @@
...
+      regionMap.computeIfAbsent(pos.getLong(), p -> createRegion(pos));
+    }
+  }
+
+  /**
+   * @param pos Region position
+   * @return The region at the given position
+   */
+  public synchronized Region getRegion(RegionPosition pos) {
+    return regionMap.computeIfAbsent(pos.getLong(), p -> {
```
leMaik, line 76: Instead of synchronizing, would it be better to use some synchronized map?

chunky/src/java/se/llbit/chunky/world/JavaDimension.java
```diff
@@ -0,0 +1,138 @@
...
+  public Date getLastModified() {
+    return new Date(this.dimensionDirectory.lastModified());
+  }
+
+  /**
+   * Get the data directory for the given dimension.
+   *
+   * @return File object pointing to the data directory
+   */
+  protected synchronized File getDimensionDirectory() {
```
leMaik, line 130: Does this need to be synchronized? Same for `getRegionDirectory`

chunky/src/java/se/llbit/chunky/world/JavaWorld.java
```diff
@@ -0,0 +1,381 @@
...
+public class JavaWorld extends World {
+
+  /** The currently supported NBT version of level.dat files. */
+  public static final int NBT_VERSION = 19133;
+
+  /** Minimum level.dat data version of tall worlds (21w06a). */
+  public static final int VERSION_21W06A = 2694;
+  public static final int VERSION_1_12_2 = 1343;
+
+  int versionId;
```
leMaik, line 53: is this intentionally package scoped?

chunky/src/java/se/llbit/chunky/world/region/RegionParser.java
```diff
@@ -60,7 +58,7 @@ public RegionParser(WorldMapLoader loader, RegionQueue queue, MapView mapView) {
         }
         ChunkView map = mapView.getMapView();
         if (map.isRegionVisible(position)) {
-          Dimension dimension = mapLoader.getWorld().currentDimension();
+          JavaDimension dimension = (JavaDimension) mapLoader.getWorld().currentDimension(); // FIXME: don't cast
```
leMaik, line 61: what's still only in `JavaDimension` that is needed here? That might show what the `Dimension` class is still missing

chunky/src/java/se/llbit/chunky/world/worldformat/Renderable.java
```diff
@@ -0,0 +1,10 @@
+package se.llbit.chunky.world.worldformat;
+
+import se.llbit.math.Octree;
+
+/** For simple world formats that only support being rendered directly */
+public interface Renderable extends Loadable {
+  void populateData(RenderableData data);
+
+  record RenderableData(Octree octree) {}
```
leMaik, line 9: That would require *two* octrees and a blockspec map, wouldn't it? As minimum data for rendering?

chunky/src/java/se/llbit/chunky/world/worldformat/WorldFormat.java
```diff
@@ -0,0 +1,15 @@
...
+
+import se.llbit.chunky.world.World;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Collection;
+import java.util.List;
+
+/** For worlds that have multiple dimensions, and fully support the map view */
+public interface WorldFormat extends Loadable {
```
leMaik, line 11: This could use the existing `Registerable` interface.

leMaik (commented):
I like the overall direction and the abstractions. :+1: 
Just some small suggestions, maybe we can simplify some code patterns while we're at it.

chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
NotStirred: The idea being abstract out a "selection region" that also happens to be the same size as a java region?

chunky/src/java/se/llbit/chunky/world/Dimension.java
NotStirred: I'm not sure this needs to be synchronised at all honestly, I'll go through and verify
these listenerse were just moved from a different place

chunky/src/java/se/llbit/chunky/world/JavaChunk.java
NotStirred: yeah me neither, great flexibility but it's unnecessary in our usecase

chunky/src/java/se/llbit/chunky/world/JavaChunk.java
NotStirred: Yeah I think the range class would be useful honestly, consider
```java
yMinSlider.setRangeFrom(heightRange);
mapView.setYRange(heightRange);
```
Plus a few other places

chunky/src/java/se/llbit/chunky/world/region/RegionParser.java
NotStirred: Dimension doesn't have regions. the map view is very region based.
Maybe we should simulate regions on every world to cope with this as you previously suggested

chunky/src/java/se/llbit/chunky/world/worldformat/Renderable.java
NotStirred: Yes. The intention for this api is schematics where everything should be loaded immediately.

I guess if we're simulating regions they can just be smuggled in as real worlds too. Thoughts?

NotStirred: Yes. The intention for this api is schematics where everything should be loaded immediately.

I guess if we're simulating regions they can just be smuggled in as real worlds too. Thoughts?

leMaik: Sounds good. Schematics support would be great! Maybe that's even be a good first test of this new API?

NotStirred:
Not sure we need to go with the virtual regions approach as discussed before https://github.com/chunky-dev/chunky/pull/1795#discussion_r2238065642

The changes required to support either virtual regions, or a per-world-format impl are both extremely small

Regions only escape dimensions in map view in one place, and scene in one place.

leMaik:
As long as we can somehow select areas in worlds, that's fine. Regions might be a thing that is special to java worlds anways. Even being limited to chunks is a very technical thing. Why would a user _want_ to be limited to loading a 16x16 blocks grid.

We can figure that out later, but the whole chunk/region selection could be a UI-only thing.

chunky/src/java/se/llbit/chunky/world/worldformat/WorldFormat.java
```diff
@@ -0,0 +1,26 @@
...
+import se.llbit.util.Registerable;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+/** For worlds that have multiple dimensions, and fully support the map view */
+public interface WorldFormat extends Registerable {
+  /**
+   * This method will be called on every possible world directory (typically this is every directory in `.minecraft/saves`).
+   *
```
leMaik: The javadoc here doesn't explain what this class does or what the isValid method does.

chunky/src/java/se/llbit/chunky/map/WorldMapLoader.java
```diff
@@ -151,14 +171,12 @@ public void regionUpdated(RegionPosition region) {
   public void reloadWorld() {
     topographyUpdater.clearQueue();
     world.currentDimension().removeChunkTopographyListener(this);
-    World newWorld = World.loadWorld(world.getWorldDirectory(), currentDimensionId,
-        World.LoggedWarnings.NORMAL);
-    newWorld.currentDimension().addChunkTopographyListener(this);
+    world.loadDimension(currentDimensionId);
+    world.currentDimension().addChunkTopographyListener(this);
```
leMaik, line 207: Removing the listeners, loading the same dimension again, and adding the listeners could be handled internally in e.g. `world.currentDimension().reload()` or `World.reloadCurrentDimension()`

`WorldMapLoader` and `World` were tightly coupled before, so this might be out-of-scope for this PR.

chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
```diff
@@ -459,19 +460,14 @@ public File getSceneFile(String fileName) {
...
```
leMaik: +1 for removing the hard-coded values from here

chunky/src/java/se/llbit/chunky/ui/controller/ChunkyFxController.java
```diff
@@ -656,14 +652,10 @@ public File getSceneFile(String fileName) {
...
-    mapLoader.loadWorld(PersistentSettings.getLastWorld());
-    if (mapLoader.getWorld().getVersionId() >= World.VERSION_21W06A) {
-      mapView.setYMin(-64);
-      mapView.setYMax(320);
-    } else {
-      mapView.setYMin(0);
-      mapView.setYMax(256);
-    }
+    mapLoader.loadWorldFromDirectory(PersistentSettings.getLastWorld());
+    IntIntPair heightRange = mapLoader.getWorld().currentDimension().heightRange();
```
leMaik: From an API standpoint, maybe this should be `HeightRange` with `getYMin()` and `getYMax()`, that way we could have javadoc and document whether the values are inclusive or exclusive (and maybe have some util methods).

chunky/src/java/se/llbit/chunky/ui/render/tabs/GeneralTab.java
```diff
@@ -592,12 +593,14 @@ private void updateCanvasCrop() {
   }
 
   private void updateYClipSlidersRanges(World world) {
-    if (world != null && world.getVersionId() >= World.VERSION_21W06A) {
-      yMin.setRange(-64, 320);
-      yMax.setRange(-64, 320);
-    } else {
-      yMin.setRange(0, 256);
-      yMax.setRange(0, 256);
+    if (world != null) {
```
leMaik, line 597: Should we disable the sliders if no world is loaded and/or set them to some default ranges and values?

chunky/src/java/se/llbit/chunky/ui/ChunkMap.java
```diff
@@ -227,8 +219,8 @@ public ChunkMap(WorldMapLoader loader, ChunkyFxController controller,
     if (view.chunkScale >= 16) {
       int minChunkX = region.x << 5;
       int minChunkZ = region.z << 5;
-      for (int chunkX = minChunkX; chunkX < minChunkX + MCRegion.CHUNKS_X; chunkX++) {
-        for (int chunkZ = minChunkZ; chunkZ < minChunkZ + MCRegion.CHUNKS_Z; chunkZ++) {
+      for (int chunkX = minChunkX; chunkX < minChunkX + Region.CHUNKS_X; chunkX++) {
```
leMaik, line 222: This still assumes that all worlds are made out of regions with 32x32 chunks, doesn't it?

chunky/src/java/se/llbit/chunky/world/java/region/JavaRegion.java
```diff
@@ -42,19 +46,8 @@
...
-  public static final int CHUNKS_X = 32;
-
-  /**
-   * Region Z chunk width
-   */
-  public static final int CHUNKS_Z = 32;
-
-  private static final int NUM_CHUNKS = CHUNKS_X * CHUNKS_Z;
+public class JavaRegion implements Region {
+  private static final int NUM_CHUNKS = Region.CHUNKS_X * Region.CHUNKS_Z;
```
leMaik, line 50: Is CHUNKS_X and CHUNKS_Z java-specific?

chunky/src/java/se/llbit/chunky/world/java/region/JavaRegionChangeWatcher.java
```diff
@@ -14,32 +14,34 @@
...
+  public JavaRegionChangeWatcher(WorldMapLoader loader, MapView mapView) {
     super(loader, mapView, "Region Refresher");
   }
 
   @Override public void run() {
     try {
       while (!isInterrupted()) {
         sleep(3000);
-        Dimension dimension = mapLoader.getWorld().currentDimension();
+        // MCRegionChangeWatcher is only created by JavaDimension, so this cast is always safe.
```
leMaik: "JavaRegionChangeWatcher is only…"

chunky/src/java/se/llbit/chunky/world/java/JavaChunk.java
```diff
@@ -0,0 +1,365 @@
...
+    request.add(Chunk.LEVEL_HEIGHTMAP);
+
+    Map<String, Tag> dataMap;
+    try {
+      dataMap = getChunkTags(request);
+    } catch (ChunkLoadingException e) { // we don't want to crash the map view if a chunk fails to load, so we warn the user
+      Log.warn(String.format("Failed to load chunk %s", position), e);
+      return false;
+    }
+    // TODO: improve error handling here.
```
leMaik, line 81: Was there some error handling before?

chunky/src/java/se/llbit/chunky/world/java/JavaChunk.java
```diff
@@ -0,0 +1,365 @@
...
+    if (dataMap == null) {
+      return false;
+    }
+    Tag data = tagFromMap(dataMap);
+
+    surfaceTimestamp = dataTimestamp;
+    version = chunkVersion(data);
+    IntIntImmutablePair chunkBounds = inclusiveChunkBounds(data);
```
leMaik: `HeightRange` might also help here to inherently document whether a range is inclusive or exclusive.

chunky/src/java/se/llbit/chunky/world/java/JavaChunk.java
```diff
@@ -0,0 +1,365 @@
...
+  private void loadSurface(@NotNull Tag data, ChunkData chunkData, int yMin, int yMax) {
+    if (data == null) {
+      surface = IconLayer.CORRUPT;
+      return;
+    }
+
+    Heightmap heightmap = dimension.getHeightmap();
+    Tag sections = getTagFromNames(data, LEVEL_SECTIONS, SECTIONS_POST_21W39A);
+    if (sections.isList()) {
+      if (version == ChunkVersion.PRE_FLATTENING || version == ChunkVersion.POST_FLATTENING) {
```
leMaik, line 107: Otherwise we might want to log a warning that the chunk version is not supported yet. Although that probably won't happen soon?

chunky/src/java/se/llbit/chunky/world/java/JavaChunk.java
```diff
@@ -0,0 +1,365 @@
...
+    } else {
+      int[] fallback = new int[X_MAX * Z_MAX];
+      for (int i = 0; i < fallback.length; ++i) {
+        fallback[i] = chunkData.maxY();
+      }
+      return fallback;
+    }
+  }
+
+  protected boolean shouldReloadChunk() {
```
leMaik: This could be renamed to `shouldReload` because this already is the Chunk class.

chunky/src/java/se/llbit/chunky/world/java/JavaChunk.java
```diff
@@ -0,0 +1,365 @@
...
+    timestamp = Math.min(timestamp, biomesTimestamp);
+    if (timestamp == 0) {
+      return true;
+    }
+    Region region = dimension.getRegion(position.getRegionPosition());
+    return region.chunkChangedSince(position, timestamp);
+  }
+
+  /** Detect Minecraft version that generated the chunk. */
+  private static ChunkVersion chunkVersion(@NotNull Tag data) {
```
leMaik: getChunkVersion

chunky/src/java/se/llbit/chunky/world/java/JavaDimension.java
```diff
@@ -0,0 +1,140 @@
...
+    });
+  }
+
+  @Override
+  public Region getRegionWithinRange(RegionPosition pos, int yMin, int yMax) {
+    return getRegion(pos);
+  }
+
+  @Override
+  public boolean regionExists(RegionPosition pos) {
```
leMaik: `hasRegion`

chunky/src/java/se/llbit/chunky/world/java/JavaDimension.java
```diff
@@ -0,0 +1,140 @@
...
+    File regionFile = new File(getRegionDirectory(), pos.getMcaName());
+    return regionFile.exists();
+  }
+
+  @Override
+  public boolean regionExistsWithinRange(RegionPosition pos, int minY, int maxY) {
```
leMaik: `hasRegionWithinRange` and then take a HeightRange/IntIntPair as range?

chunky/src/java/se/llbit/chunky/world/java/JavaDimension.java
```diff
@@ -0,0 +1,140 @@
...
+    return this.regionExists(pos);
+  }
+
+  @Override
+  public IntIntPair heightRange() {
```
leMaik: `getHeightRange`, also should these be constants? post 21w06a worlds might even be higher, is this just the default height range? If so, this method should be renamed to `getDefaultHeightRange` and we might need a `hasCustomHeightRangeSupport` or something.

chunky/src/java/se/llbit/chunky/world/java/JavaDimension.java
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * @return File object pointing to the region file directory
+   */
+  public synchronized File getRegionDirectory() {
+    return new File(getDimensionDirectory(), "region");
+  }
+
+  @Override
+  public String getName() {
+    return dimensionDirectory.getName() ;
```
leMaik: the dimension directory names changed recently; should we use something else instead of the directory name? there wasn't a directory for the overworld before

chunky/src/java/se/llbit/chunky/world/java/JavaWorld.java
```diff
@@ -0,0 +1,256 @@
...
+import java.util.stream.Collectors;
+import java.util.zip.GZIPInputStream;
+
+public class JavaWorld extends World {
+  /** The currently supported NBT version of level.dat files. */
+  public static final int NBT_VERSION = 19133;
+
+  /** Minimum level.dat data version of tall worlds (21w06a). */
+  public static final int VERSION_21W06A = 2694;
+  public static final int VERSION_1_12_2 = 1343;
```
leMaik, line 28: Is `VERSION_1_12_2` unused?

chunky/src/java/se/llbit/chunky/world/java/JavaWorld.java
```diff
@@ -0,0 +1,256 @@
...
+      return new JavaDimension(world, dimensionId, dimensionDirectory, playerEntities, spawnPos);
+    }
+
+    dimensionDirectory = switch (dimensionId.getNamespacedName()) { // TODO in Java 21+ we can use `switch (dimensionId)` here
+      case "minecraft:the_nether" -> new File(worldDirectory, "DIM-1");
+      case "minecraft:the_end" -> new File(worldDirectory, "DIM1");
+      default -> worldDirectory;
+    };
+    if (new File(dimensionDirectory, "region3d").exists()) {
+      return new CubicDimension(world, dimensionId, dimensionDirectory, playerEntities, spawnPos);
```
leMaik, line 143: Just remember to update this once CubicChunks supports 26.1 or later. 🙃

chunky/src/java/se/llbit/chunky/world/java/JavaWorld.java
```diff
@@ -0,0 +1,256 @@
...
+        }
+      }
+    }
+  }
+
+  /**
+   * Reload player data for the current dimension. This method is not in Dimension because players are per-world, not per-dimension
+   * @return {@code true} if player data was reloaded.
+   */
+  synchronized boolean reloadPlayerData() {
```
leMaik, line 190: Is this intentionally package-private because the JavaDimension calls this?

chunky/src/java/se/llbit/chunky/world/java/JavaWorld.java
```diff
@@ -0,0 +1,256 @@
...
+      this.playerEntities.clear();
+      this.playerEntities.addAll(getPlayerEntityData(worldDirectory, player));
+    } catch (IOException e) {
+      Log.infof("Could not read the level.dat file for world %s while trying to reload player data!", levelName);
+      return false;
+    }
+    return true;
+  }
+
+  public static boolean isWorldDir(File worldDir) {
```
leMaik: …ectory

chunky/src/java/se/llbit/chunky/world/region/Region.java
```diff
@@ -6,6 +6,15 @@
 import se.llbit.chunky.world.RegionPosition;
 
 public interface Region extends Iterable<Chunk> {
+  /**
+   * Region X chunk width
+   */
+  int CHUNKS_X = 32;
```
leMaik, line 12: Is this independent of the world format? If this is just a UI thing for selection, we should move it into the UI classes.

chunky/src/java/se/llbit/chunky/world/worldformat/WorldFormat.java
```diff
@@ -0,0 +1,26 @@
+package se.llbit.chunky.world.worldformat;
+
+import se.llbit.chunky.world.World;
+import se.llbit.util.Registerable;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+/** For worlds that have multiple dimensions, and fully support the map view */
```
leMaik: Since "has no dimensions" is the same as "has one dimension" and that would be supported, I guess we could lower the restrictions. The map view could draw schematics too, so that's nothing the world format should be bothered about.

Maybe change it to `A world format that Chunky can load."

chunky/src/java/se/llbit/chunky/world/worldformat/WorldFormats.java
```diff
@@ -0,0 +1,55 @@
...
+        } catch (IOException e) {
+          Log.error(String.format("An error occurred when trying to load a world using format `%s` from %s", format.getName(), dir.getAbsolutePath()), e);
+        }
+      }
+    });
+
+    if (providedWorlds.size() > 1) {
+      // Maybe allow the user to select which?
+      // This method is called from a variety of different popup/menu situations, is this ^ possible?
+      Log.warn(String.format("The directory %s has multiple valid world formats: %s", dir.getAbsolutePath(), String.join(", ", providedWorlds.keySet())));
```
leMaik: Load a random one, that'll be fun in #help 🤣

chunky/src/java/se/llbit/chunky/world/EmptyWorld.java
```diff
@@ -31,8 +35,22 @@ private EmptyWorld() {
     this.currentDimension = EmptyDimension.INSTANCE;
   }
 
+  @Override
+  public Set<Dimension.Identifier> availableDimensions() {
```
leMaik: `getAvailableDimensions` and `getDefaultDimension`

chunky/src/java/se/llbit/chunky/world/World.java
```diff
@@ -37,32 +26,20 @@
...
-  private final String levelName;
-  private int gameMode = 0;
-  private final long seed;
-
-  private int versionId;
+  protected final String levelName;
+  protected int gameMode = 0;
+  protected final long seed;
 
   /** Timestamp for level.dat when player data was last loaded. */
```
leMaik: *Timestamp when player data was last... I don't know. Loaded? Saved?

leMaik (changes requested):
Mostly minor comments on the API. Overall this refactoring is still great and it's nice to see how CubicChunks, Java edition and, in the future, bedrock support fit together.

It seems like the whole regions and chunks stuff is mostly needed for the UI (outside of java world loading, of course)?

chunky/src/java/se/llbit/chunky/world/java/JavaChunk.java
NotStirred: No, this todo has been here since before 2020 iirc

chunky/src/java/se/llbit/chunky/world/java/JavaDimension.java
NotStirred: Currently this represents the actual height of the dimension, we check the world's version and give the appropriate height range.

This is where the ymin/ymax on the map and scene UIs comes from.

chunky/src/java/se/llbit/chunky/world/java/JavaWorld.java
NotStirred: Yes.

chunky/src/java/se/llbit/chunky/world/worldformat/WorldFormats.java
NotStirred: Currently the order is based on arbitrary hashmap ordering.

Though two world formats saying the same world is compatible in both would be... interesting.

chunky/src/java/se/llbit/chunky/world/World.java
NotStirred: :+1: Also this should be in JavaWorld

chunky/src/java/se/llbit/chunky/world/region/Region.java
NotStirred: Yes, it's used in scene's chunk loading to batch by region, and in the map view for per-region selecting

As well as being used in `JavaRegion` for its internal state

chunky/src/java/se/llbit/chunky/ui/ChunkMap.java
NotStirred: Yes, currently I've got "virtual" regions in bedrock that just pass calls to the dimension.

Any worldformat could use the virtual regions, and they aren't format specific (schematics/etc.). Maybe it's the simplest approach.

It does mean every dimension needs to hold onto a bunch of virtual regions that don't do much, which is unfortunate

chunky/src/java/se/llbit/chunky/world/java/JavaDimension.java
NotStirred: Previously the names would be `<worldname>` `DIM-1` and `DIM1`
Now the names would be `minecraft:overworld`, `minecraft:the_nether`, `minecraft:the_end`

Probably should change this at some point, but maybe another PR?

chunky/src/java/se/llbit/chunky/world/java/JavaWorld.java
NotStirred: Yeah, I don't really like it but don't see another good option

NotStirred:
`WorldFormat#isValid` is now technically unnecessary as `WorldFormat#getWorldInfo` returns optional on failure.

Plus we're doing the same checks in both anyway. Thoughts?

leMaik:
Returning Optional in case of failure instead of throwing an exception doesn't sound right.

NotStirred:
It's failure in the sense that there was no world info there, maybe the method would be better named `getWorldInfoIfPresent`. We could throw if it's not present, and keep `isValid`.
Previously we would return `EmptyWorld.INSTANCE` on failure (see #1896).


Maybe `getWorldInfo` should also `throws IOException` such that it:
- returns `WorldInfo` on success
- returns empty optional if not valid for the world format
- throws on unexpected error, to be reported to the user

Also, currently errors are sometimes suppressed inside `JavaWorld`, I believe the intention is to avoiding creating a bunch of warnings when you open the world chooser. I think it makes sense for the caller (`WorldFormats`) to do that.

PR #1809, opened 2025-06-13
Fix spherical aperture depth of field and add custom aperture support

Peregrine05:
**Before:**
image

**After:**
image

Custom aperture support which previously did not work with projectors that used the `SphericalApertureProjector`:
image

### DIFF
--- a/chunky/src/java/se/llbit/chunky/renderer/projection/ApertureProjector.java
+++ b/chunky/src/java/se/llbit/chunky/renderer/projection/ApertureProjector.java
@@ -24,6 +24,7 @@ import se.llbit.chunky.renderer.ApertureShape;
 import se.llbit.chunky.resources.BitmapImage;
 import se.llbit.log.Log;
 import se.llbit.math.Ray;
+import se.llbit.math.Vector2;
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/chunky/renderer/projection/ApertureProjector.java
```diff
@@ -93,11 +93,11 @@ static private BitmapImage loadApertureMask(ApertureShape apertureShape) {
...
-  @Override public void apply(double x, double y, Random random, Vector3 o, Vector3 d) {
-    wrapped.apply(x, y, random, o, d);
-
-    d.scale(subjectDistance / d.z);
-
+  /**
+   * Finds a random point in the aperture mask.
+   * @return Array of two doubles which are the x- and y- coordinates of the point.
+   */
+  protected double[] getPointInAperture(Random random) {
```
leMaik, line 100: This could as well return `Vector2`

Peregrine05: Done.

PR #1826, opened 2025-08-13
Abstract PlayerEntity in preparation for other mobs

ThisIsBrady:
This PR refactors PlayerEntity into a HumanoidEntityBase in preparation for mobs that share the model and gear slots with it. 

* Moved Player model to HumanoidEntityBase class
* Updated armor texture paths
* Updated to also load armor items from the new `equipment` NBT tag (1.21.5)
* Moved inventory loading from PlayerEntityData to HumanoidEntityBase

Looking for review, especially regarding PlayerEntityData. I'd like to fully remove that class to have the PlayerEntity/HumanoidEntity classes own that data, but world loading needs to load some of it before the Entities are actually constructed.


leMaik:
Is this about entities that can have items? I like Bukkit's `InventoryHolder` interface name for that. :thinking:
The rest looks good. Regarding NBT and texture path changes: Just make sure that we are compatible with older Minecraft versions too.

PR #1577, opened 2023-04-15
Use more intuitive sun controls; refactor code

Peregrine05:
- Re-added an `Enable sunlight` control, which replaces `Sun Sampling Strategy`: `NON_LUMINOUS`.
- Added a control to enable and disable skymap interpolation.
- Refactored code that returns sky color.

image
image

PR #1830, opened 2025-08-27
Refactor the path tracer

### LINKED ISSUE #26: Settings for emitted light color
Add settings in the Render Controls for emitted light color.

### LINKED ISSUE #361: Adjustable cloud transparency & shadows
Clouds seem to make some scenes look really dark. I would like to adjust the amount of shade they provide, all the way down to "zero" in some cases.

### LINKED ISSUE #502: Low lying fog? Volumetric clouds?
I believe this forum is slightly dead and the last update was long ago but i would to see some kind of feature that allows for volumetric fog that i can set to end at a specific height. Imagine a sunrise in the mountains with Trees in the valeys and Just having some gorgeous low hanging clouds/fog between the Trees and Just having the top of the Trees sticking out of the low lying clouds/fog 
It would be so perfect and would make for some beautifull renders

### LINKED ISSUE #612: [discussion] Refactoring suggestion

### LINKED ISSUE #636: Water plane and normal water interaction
Apart from the jello water that should be fixed by #635, using water plane creates some artifacts when interacting with normal water.
When the water plane is set at a different height than the normal water, we get a "double submerging" effect.
Water plane above normal water :
higher-level-new-fix
(Visible on the left island)

Water plane below normal water:
lower-level-new-fix
(Visible on the left island)

When the water plane is at the same level than the normal water, there is still an artifact where the octree stops
Water plane at the same level:
same-level-new-fix
(See the top right corner of the image)

### LINKED ISSUE #665: Limitations of the current fog implementation

### LINKED ISSUE #893: Proper slime and honey intersection
The current implementation of slime/honey intersection looks good but is incorrect.

When intersecting honey and slime blocks, the ray intersects two surfaces in the same block ("at once") and then overlays the colors, which is technically wrong.

It should instead hit the slime surface, refract due to entering slime/honey and then hit the core in the next ray iteration. The ray would travel through slime/honey for some distance, which it doesn't do right now.

### LINKED ISSUE #980: Add support for chaining of post processors
It would be a nice addition to be able to chain post processing operations to get for example a bloom effect but with ACES tone mapping.

As I see things, there is 3 types of post processing operations:
 * Operations that take an image with linear colors and output an image with linear colors, for example a bloom effect, a operation to change saturation/exposure, etc... Those operation are HDR -> HDR.
 * Operations that take an image with linear color and output an image with non-linear colors, aka tone mapping operations, for example gamma correction ACES tone mapping etc... Those operations are HDR -> HDR but the output is implicitly non-linear.
 * Clipping operation that take a non-linear HDR image and output a SDR image, for example RGB clamping or the operations described here https://bottosson.github.io/posts/gamutclipping/ 

A normal post processing pipeline should contain 0 to n type 1 operations, 1 type 2 operation and 1 type 3 operation in this order. But for simplicity we could choose to not enforce that.

All 3 types of operations could be represented by the same interface whose main method takes an image where components are stored as float and outputs an image where components are stored as float as well. What would change is how the various implementations of this interface interpret the input data based on their operation type.
Type 1 and type 2 would interpret their input as linear HDR, type 3 would interpret its input as non-linear HDR. Type 1 would output linear HDR, type 2 would output non-linear HDR and type 3 would output non-linear SDR (values limited between 0 and 1). The final conversion to 8-bits will be done after all the processing operations and is not considered a processing operation itself as it has a different interface.

### LINKED ISSUE #1162: The `Face camera` and `Face target` buttons are non-functional
After selecting an entity in the entities list, using the `Face camera` and `Face target` buttons has no visible effect on the entity.

### LINKED ISSUE #1230: Lighting is not uniform underwater
The lighting underneath the surface of the water is not uniform. When viewed from above the surface of the water, The lighting becomes darker for the first block under the water surface, and periodic waves of brighter lighting are seen intersecting with the surface of the water at the top corners of the blocks. The next blocks underneath cut into brighter lighting and then continue with a more normal lighting.
Emitters test-128

When viewed from underneath the surface of the water, the lighting cutoffs are not visible, and a similar effect of periodic waves of brighter lighting is seen aligning with the edges of the blocks.
Emitters test-33

### LINKED ISSUE #1258: Change the apparent brightness of emitters based on their intensity
Currently, if emitters are set to be very intense, the lighted area around them will outshine the emitter itself, when it should be the other way around. The apparent brightness of the emitter should change with the emitter intensity to be more realistic.

### LINKED ISSUE #1409: Expose opaque material property in `Materials Tab`
People often request a way to disable/hide blocks such that they do not render.

`Specular` set to a value of `-1` is currently the only way to perform this action without editing the code however this method is unofficial and has numerous issues with the most egregious being it makes Chunky extremely unstable. This method also produces an x-ray bug type effect; That being directly adjacent blocks to the "hidden" block do not have a surface rendered.

ie -
image

---

Exposing the `Opaque` material property in the `Materials Tab` would give people access to a much more stable and predicable method of hiding blocks.

As noted, we would need to include a tooltip, or note, indicating that Chunks need to be reloaded in order for these changes to apply.

### LINKED ISSUE #1476: Hide the Cloud size and position controls when clouds are disabled
Similar to #1454, make `Cloud size`, and `Cloud X`-`Z` collapsible, and hide them when `Enable clouds` is not selected. I think this will just aid in decreasing overall visual clutter, particularly because the position options don't actually provide meaningful information unless you've got clouds enabled.
In the pull request where I first brought this up I mentioned hiding the sun controls based on `Draw sun` being disabled, but I've now realised that's not actually desirable or correct behavior, so I think they should stay as they are currently.

### LINKED ISSUE #1477: `minecraft:light` minimum ray depth to emit light is 5
`minecraft:light` is an odd and hacky case where we need a block that can emit light but not show in the render.

Preview
image

Unlike other light emitting blocks that emit light at a minimum of ray depth 3:

image

However `minecraft:light` requires a minimum of ray depth 5 to work.

Ray Depth 4
image

Ray Depth 5
image

Ray Depth 6
image

Ray Depth 25
image

---

### LINKED ISSUE #1482: Create an interface for additional GUIs for projector customization
This will be relevant for ODS interpupillary distance adjustability.

### LINKED ISSUE #1515: IOR is incorrect

### LINKED ISSUE #1591: Smoothness should affect transmitted light

### LINKED ISSUE #1617: Water world water and normal water are different colors.
image

### LINKED ISSUE #1642: Water plane looks different inside/outside the octree
The water plane looks different when it is outside of the octree compared to inside the octree but no chunks. Orignally reported in Discord: https://discord.com/channels/541221265512464394/1156386708145848451/1156386708145848451

image

@ThatRedox already investigated this issue a bit: 
> looks like rays that intersect with the water plane will continously enter and exit the water octree
> Also rd<=4 will not have the weird artifacts in the octree
>
> ok so it looks like its going something like:
> 1. Initial ray going down
> 1. Hits water plane -> air-water refraction
> 2. Hits air in water octree -> water-air reflection, now going up
> 3. Hits water plane -> air-water refraction
> 4. Hits air in water octree -> water-air refraction
> 5. Hits sky

### LINKED ISSUE #1669: Add option to override sun color to match sunlight color.
Currently FAST and other sun sampling modes have very different results since FAST forces the sun color to the sunlight color control while other methods sample the sun's texture directly. We should add an option to override the sun color in other sun sampling modes too.

### LINKED ISSUE #1671: Smarter Intersections and expose it on Materials tab
As seen in #1515, the way Chunky handles intersections in between materials is sometimes inconsistent. Additionally there are other quirks noticed between different blocks models, ie glass panes do not intersect on the transparent part because of how `QuadModel` handles alpha. However how we intersect something like glass should be different from something like flowers and could be adjustable. Additionally a user might be inclined to enable or disable back faces for some block models. As such, the way that intersection logic works should at least partly dependent on a blocks Material properties. The advantage of having exposed materials properties is that on some materials we can have simplified logic allowing for faster marching though the oct tree, while other materials would have more careful intersection parameters. Addionally, this can be made forward compatible with existing scenes.

Here is some intersection logic considerations:
Are we looking at the front or back face
1.1 Are we inside a block? (might be diff from 1 )
If looking at a FRONT face
do we want to intersect on change of material (ie glass air)
do we want to intersect on 0 alpha (glass)
do we want to intersect on geometry

Looking at a BACK face
do we want to intersect on change of material (ie glass air)
do we want to intersect on 0 alpha (glass)
do we want to intersect on geometry
do we want to intersect if inside a block

Finally: Heres what we want to do on an intersection:
pass though the ray, dont do anything (want to minimize these cases)
update ray direction, dont update cumulative color (ie to fix refractions without rendering back side) (dono if this would just cause artifacts 
treat as a full surface
Im not sure if there are more relevant cases but that's the stuff I came up with

For example with glass allowing intersections on blackfaces only on material boundaries would probably fix #1515 without being as hacky as #1601

Peregrine05:
Closes #26
Closes #361
Closes #502
Closes #612
Closes #636
Closes #665
Closes #893
Closes #980
Closes #1162
Closes #1230
Closes #1258
Closes #1409
Closes #1476
Closes #1477
Closes #1482
Closes #1515
Closes #1591
Possibly closes #1617
Possibly closes #1642
Closes #1669
Closes #1671

Closes #1494 in favor of this
Closes #1577 in favor of this
Closes #1601 in favor of this
Includes the changes from #1627, and adds a GUI for the reference color system.
Closes #1660 in favor of this
Closes #1668 in favor of this
Closes #1673 in favor of this
Includes the changes from #1809

The primary purpose of this PR is to refactor the path tracer to use a simple ray, and intersection records. The new ray stores a `Vector3` origin, a `Vector3` direction, the material it is traversing, and a flags integer. The IntersectionRecord stores distance, geometry normal, shading normal, uv coordinates, material, color, and a flags integer.

I fixed a few other problems, but unfortunately did not think to change branches while doing so.

Notably:

- I made use of the [ControlsFX](https://controlsfx.github.io/) library for the `ToggleSwitch` control, which is a better design choice than a `CheckBox` when the effects of the control are realized immediately, rather than later.

- The `Lighting` tab was removed. Its contents pertaining to the sky and to the sun were relocated to the new `Environment` tab. The contents pertaining to emitters were relocated to the new `Emitters` tab.

- The `Sky & Fog` tab was removed. Its contents pertaining to the sky were relocated to the `Environment` tab. Its contents pertaining to fog and clouds were relocated to the new `Fog & Clouds` tab.

- The `Water visibility`, `Water opacity`, and water color controls were removed, in favor of using material properties to control these properties. Notably, biome-tinted water is no longer possible, but it doesn't look realistic to begin with.

- More material properties were added to the `Materials` tab.

- Materials can be right-clicked in the render preview to be edited through a dialog box. The advantages of this approach are that individual blockstates can be edited, which is something that the `Materials` tab does not support. While changes to octree materials are saved to the blockPalette, allowing them to persist through scene saves and restarts, they are not saved to the scene.json, making external editing difficult. Also, individual materials in the BVH can be edited; however, since individual BVH materials are not saved to the scene.json, these changes will not persist through scene saves and restarts.

These material properties are:

- `Emittance color`: Changes the tint of the color of emitted light.

- `Emitter mapping offset`: See #1627 for details.

- `Emitter mapping type`: See #1627 for details.

- `Use reference colors`: Extracted this option to its own control, so that it can be used in conjunction with the other emitter mapping types.

- `Alpha`: Changes the opacity of the texture.

- `Subsurface scattering`: Changes the percentage of rays that have their normal inverted, so that they scatter to the other side of the texture.

- `Diffuse color`: Changes the tint of diffuse reflections.

- `Specular`: Changes the percentage of rays that are specularly reflected with no texture tinting.

- `Transmission smoothness`: Changes the smoothness for transmitted and refracted rays.

- `Opaque`: This option requires a chunk reload for its effects to be realized.

- `Hidden`: Prevents intersections with this material entirely.

- `Volume density`: When the current ray medium is the selected material, this sets the density of the volume scatter medium.

- `Absorption`: Rays travelling through the medium will be attenuated if the value is greater than 0.

- `Metalness`: Changes the percentage of specularly-reflected rays that are texture-tinted.

- `Transmission metalness`: Changes the percentage of specularly-transmitted of refracted rays that are texture tinted.

The reference color GUI should be self-explanatory.

Other things of note are:

- Branched path tracing is not functional at the moment. The infrastructure is still present in the renderer, but it is not implemented in the path tracer. As such, branch count is hard-coded to *1*, and the control is hidden.

- Fancier translucency is removed, because the new path tracer should handle transparent colors correctly* (At least it seems to do so.)

- The sun sampling options have been changed. `Sample through opacity` does what Chunky's current `FAST` does. `Sample only` stops when it hits certain interactions (like glass). `Diffuse` treats the sun as part of the skymap. `Mix` allows diffuse sun intersections after hitting those certain interactions that `Sample only` does not handle, but prevents diffuse sun intersections when sampling the sun directly is possible. Overall, the results should be the same as with `Diffuse`.

- `Draw sun` enables or disables sunlight entirely, through adding or removing the sun texture from the skymap. It does not stop "sun sampling", however, as Chunky will still sample the sky where the sun is. This can be used to sample an HDRi, for example.

- `Use flat texture` overrides the sun texture provided by the resource pack with a simple solid color square. It is enabled by default, and should be used when you want a specific sun color.

- `Sunlight color` tints the sun texture when `Use flat texture` is disabled, and controls the sun color directly when it is enabled.

- The `Nishita` sky model now has many more customization options available.

There may still be plenty of bugs, so this PR could use some testing.


Peregrine05:
To build and run the PR, follow the instructions below.

---

Download ControlsFX from here: https://repo1.maven.org/maven2/org/controlsfx/controlsfx/11.2.1/controlsfx-11.2.1.jar

Create a new folder somewhere, such as `controlsfx` in the `.chunky` folder. Add the ControlsFX Jar to the folder.

Build Chunky with `./gradlew buildReleaseJar`. (The update site doesn't seem to work for this PR.) Run the build (which will be in `build/installer/`).

 In the `Java options` field in the Launcher, move the cursor to the end of the value of the `--module-path` argument. Enter a  semicolon `;`, and then add the path to the folder containing the ControlsFX Jar. Then move the cursor to the end of the value of the `--add-modules` argument. Enter a comma `,`, and then enter: `org.controlsfx.controls`.

Your Java options should look something like:

`--module-path /home/user/.chunky/javafx/javafx-sdk-17.0.2/lib;/home/user/.chunky/controlsfx --add-modules javafx.controls,javafx.fxml,org.controlsfx.controls`

chunky/src/java/se/llbit/chunky/block/FixedTopBottomRotatableTexturedBlock.java
```diff
@@ -13,13 +13,13 @@ public FixedTopBottomRotatableTexturedBlock(String name, String facing, Texture
         this(name, facing, front, side, side, side, top, top);
     }
 
-    public FixedTopBottomRotatableTexturedBlock(String name, String facing, Texture front, Texture side, Texture top, Texture bottom) {
-        this(name, facing, front, side, side, side, top, bottom);
+    public FixedTopBottomRotatableTexturedBlock(String name, String facing, Texture front, Texture side, Texture bottom, Texture top) {
```
leMaik, line 16: Why switch these arguments?

chunky/src/java/se/llbit/chunky/block/MinecraftBlockProvider.java
```diff
@@ -1549,7 +1552,7 @@ public Block getBlockByTag(String namespacedName, Tag tag) {
       case "ice":
         return new MinecraftBlockTranslucent(name, Texture.ice);
       case "snow_block":
-        return new MinecraftBlock(name, Texture.snowBlock);
+        return new MinecraftBlockTranslucent(name, Texture.snowBlock);
```
leMaik, line 1555: is snow actually translucent?

chunky/src/java/se/llbit/chunky/block/Block.java
```diff
@@ -119,7 +105,7 @@ public boolean isBlockWithEntity() {
     return false;
   }
 
-  public Entity toEntity(Vector3 position) {
+  public Entity[] toEntity(Vector3 position) {
```
leMaik, line 108: Why was this change necessary? Are there blocks that require multiple entities? Cake with candles? :thinking:

chunky/src/java/se/llbit/chunky/entity/Lectern.java
```diff
@@ -14,7 +14,7 @@
 import se.llbit.math.primitive.Primitive;
 import se.llbit.util.JsonUtil;
 
-public class Lectern extends Entity implements Poseable {
+public class Lectern extends Entity {
```
leMaik, line 17: Oh, I see… separating lectern and book.

chunky/src/java/se/llbit/chunky/block/Block.java
leMaik: We need to break this PR down into smaller changes that can actually be merged, that would probably be simpler than trying to resolve the conflicts.

I'd merge #1376 so this change could be dropped from this PR.

chunky/src/java/se/llbit/chunky/renderer/postprocessing/AldridgeFilmicFilter.java
```diff
@@ -0,0 +1,88 @@
+package se.llbit.chunky.renderer.postprocessing;
```
leMaik, line 1: I like the tonemap changes; they're also a good candidate to be separated into a separate PR.

chunky/src/java/se/llbit/chunky/block/MinecraftBlockProvider.java
Peregrine05: In Minecraft, it isn't, but in real life, snow has subsurface scattering, so this change will allow the user to add subsurface scattering if they desire.

chunky/src/java/se/llbit/chunky/block/Block.java
Peregrine05: Okay; I have reverted the changes to entities in favor of #1376.

Peregrine05:
Conflicts have been resolved.

PR #1833, opened 2025-08-31
Remove block entity blocks from the octree and remove the unused invisible field.

leMaik:
This builds upon #1376. While working on that PR, I found that we don't remove block entities from the octree yet, even if they have no actual block model (ie. `intersect` always returns `false`), which is the case for all block entity blocks except decorated pots.

This PR leads to fewer blocks in the octree, thus possibly smaller octrees and more merged nodes. E.g. signs are air now and not sign blocks that need intersection.

### DIFF
--- a/chunky/src/java/se/llbit/chunky/block/Block.java
+++ b/chunky/src/java/se/llbit/chunky/block/Block.java
@@ -13,6 +13,7 @@ import se.llbit.math.Vector3;
 import se.llbit.nbt.CompoundTag;
 import se.llbit.nbt.Tag;
 
+import java.util.Collection;
 import java.util.Random;
 
 public abstract class Block extends Material {
@@ -119,7 +120,7 @@ public abstract class Block extends Material {
     return false;
   }
 
-  public Entity toEntity(Vector3 position) {
+  public Collection<Entity> toEntity(Vector3 position) {
     throw new Error("This block type can not be converted to an entity: "
       + getClass().getSimpleName());
   }
--- a/chunky/src/java/se/llbit/chunky/block/Block.java
+++ b/chunky/src/java/se/llbit/chunky/block/Block.java
@@ -151,7 +151,8 @@ public abstract class Block extends Material {
   }
 
   /**
-   * Whether to remove this block from the octree if it contains entities (i.e. {@link #hasEntities()} returns true).
+   * Whether to remove this block from the octree if it contains entities (i.e. {@link #hasEntities()} or {@link #isBlockEntity()}
+   * return true).
    * <p>
    * Most blocks are replaced by their entities (eg. signs create a sign entity that does the rendering and the block itself
    * does nothing, but some blocks use block model and entities, e.g. candle (where the candle flame is an entity but the candle is a block).
--- a/chunky/src/java/se/llbit/chunky/block/Block.java
+++ b/chunky/src/java/se/llbit/chunky/block/Block.java
@@ -103,7 +103,7 @@ public abstract class Block extends Material {
   }
 
   public Entity toBlockEntity(Vector3 position, CompoundTag entityTag) {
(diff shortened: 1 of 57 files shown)


chunky/src/java/se/llbit/chunky/block/Block.java
```diff
@@ -97,31 +91,67 @@ public String toString() {
...
-    throw new Error("This block type can not be converted to a block entity: "
+  /**
+   * Create a block entity from this block and the given block entity tag at the specified position.
+   *
+   * @param position  Position
+   * @param entityTag Block entity tag
+   * @return The block entity created from this block's data and the entity tag
+   * @throws UnsupportedOperationException If this block is not a block entity (i.e. {@link #isBlockEntity()} returns false
+   */
+  public Entity createBlockEntity(Vector3 position, CompoundTag entityTag) {
```
NotStirred, line 115: An additional thought, is it possible for a user to need both the entity tag and the ability to return multiple entities?

chunky/src/java/se/llbit/chunky/block/Block.java
```diff
@@ -97,31 +91,67 @@ public String toString() {
...
-    throw new Error("This block type can not be converted to an entity: "
-      + getClass().getSimpleName());
+  /**
+   * Whether to remove this block from the octree if it contains entities (i.e. {@link #hasEntities()} or {@link #isBlockEntity()}
+   * return true).
+   * <p>
+   * Most blocks are replaced by their entities (eg. signs create a sign entity that does the rendering and the block itself
+   * does nothing, but some blocks use block model and entities, e.g. candle (where the candle flame is an entity but the candle is a block).
+   */
+  public boolean isReplacedByEntities() {
```
NotStirred, line 153: My understanding was that `isBlockEntity` represented this behaviour already, you either `isBlockEntity` and get replaced with the result, or `hasEntity` and don't.

NotStirred (approved):
Looks great :+1: just some bikesheddable random thoughts

chunky/src/java/se/llbit/chunky/block/Block.java
NotStirred: Does it make any sense to just have this api `hasEntity` `createEntities` `isReplacedByEntities`, removing the second path?

Or have I just misunderstood the idea?

chunky/src/java/se/llbit/chunky/block/Block.java
leMaik: The closest we have is the decorated pot, which uses the block entity information to update the block tag (creating a new `Block` instance) and then uses an entity to create the upper part (which is higher than 1x1x1 and thus can't be in the octree).

chunky/src/java/se/llbit/chunky/block/Block.java
leMaik: I absolutely agree. The current interface let's the block decide what to implement bit the callee decides which method to call. That only works well because there is only one callee.

leMaik: @NotStirred The problem is that we don't have the tile entity data when we instantiate the block. We load a chunk's tile entities after all blocks have been loaded and then either create block entities or create a new tag (and from that, a new block) with the tile entity data.

`Collection<Entity> createEntities(Vector3 position, Optional<CompoundTag> entityTag)`

If we were to do this, we would need to load the tile entities first, put them into some `Map<Position, CompoundTag>` and then get the correct entity tag while iterating through the blocks. The interface wouldn't be good though. If the tile entity is there, it isn't optional – it is required to create the entities.

A better way to think about `createEntities` and `createBlockEntity` is probably to think about block instances as factories for the entities that correspond to their block type. And of course, the callee of a factory knows which method to call (and the factory has methods to tell the callee what it supports).

PR #1835, opened 2025-09-03
Make ApertureShape implement Registerable

Peregrine05:


leMaik:
To make this a proper plugin api, we should remove the enum and add a way for plugins to add custom aperture shapes. Given that we have a `custom` type for aperture shape, what's the use-case for this?

PR #1836, opened 2025-09-05
Fix cloud textures being overlaid when changing resource packs

### LINKED ISSUE #1831: Changing the clouds texture overlays the old texture and the new texture
When loading a new resource pack to change the clouds texture, Chunky essentially overlays the new texture over the old texture, resulting in a combination of the textures. Chunky must be restarted to reset the clouds texture to the current resource pack.

Peregrine05:
Fixes #1831


leMaik (approved):
lgtm, I'll rename `zero` to `clear` though

PR #1840, opened 2025-09-14
Parse canvasConfig width and height if present

Peregrine05:
Chunky has a `canvasConfig` now, so this PR updates the scene parser in the scene loader dialog to use the `canvasConfig` `width` and `height` values, if present, and to fall back to the scene JSON root `width` and `height` otherwise.


Peregrine05:
Thanks for catching that. When I saw the issue present, I thought it would also be present on master (like with the resin bricks issue) but I forgot to actually check.

leMaik:
tbh I was confused for a moment too and had to check the code on `master` first :sweat_smile:

PR #1856, opened 2025-12-24
Remove waterlogged special cases

### LINKED ISSUE #1692: Singletons (mainly `Air#INSTANCE`) and constants are incompatible with `BlockSpec#maybeWaterlogged`
`BlockSpec#maybeWaterlogged` modifies the returned object; if that object is a constant any previous or future reference to that object is modified. 
For example if a world contains `air` and `air + waterlogged=true` then all air blocks in the entire world will be waterlogged.

Any `BlockProvider` implementation that returns the same object from `getBlockByTag` for different arguments could run into this issue.

leMaik:
Fixes #1692 in a (hopefully) slightly more elegant way.

~We could even move `waterlogged` and `isWaterlogged()` into `Block` where it would belong, but unfortunately the octree returns `Material` for blocks and changing that in 2.5.0 would break the octree plugin.~ Done and it doesn't break anything. :tada:

### DIFF
--- a/chunky/src/java/se/llbit/chunky/block/Block.java
+++ b/chunky/src/java/se/llbit/chunky/block/Block.java
@@ -32,6 +32,8 @@ public abstract class Block extends Material {
    */
   public boolean invisible = false;
 
+  protected boolean waterlogged = false;
+
   public Block(String name, Texture texture) {
     super(name, texture);
   }
@@ -175,5 +177,13 @@ public abstract class Block extends Material {
   public Block applyWaterlogging() {
     this.waterlogged = true;
     return this;
+  }
+
+  public boolean isWaterlogged() {
+    return waterlogged;
+  }
+
+  public boolean isWaterFilled() {
+    return waterlogged || isWater();
   }
 }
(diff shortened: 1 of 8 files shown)


chunky/src/java/se/llbit/chunky/block/MinecraftBlock.java
```diff
@@ -7,7 +7,12 @@
  * A simple opaque block with a single texture.
  */
 public class MinecraftBlock extends Block {
-  public static final Material STONE = new MinecraftBlock("stone", Texture.stone);
+  public static final Material STONE = new MinecraftBlock("stone", Texture.stone) {
+    @Override
+    public Block applyWaterlogging() {
+      throw new UnsupportedOperationException();
```
NotStirred: I wonder if this should only throw in a dev environment, just to avoid crashing in prod

Though chunky in general probably needs this everywhere

NotStirred (approved):
nice 👍

chunky/src/java/se/llbit/chunky/block/MinecraftBlock.java
leMaik: I was wondering too, but this particular `STONE` instance is only used in the preview renderer. I'll replace this with a `return this` instead. Waterlogged stone would just be stone anways. :shrug:

PR #1682, opened 2024-01-16
Improve launcher command line interface.

ThatRedox:
* Rewrite the launcher CLI using Apache Commons CLI.
* Improve launcher interactive setup.
* Setup unit tests for the launcher.

This does slightly change the usage of the CLI. The new parser parses all command line arguments until `--` for the launcher. Any arguments after `--` will be passed to Chunky. Any unknown commands will return with `Unrecognized option: -<command>`.

## Running `java -jar ChunkyLauncher.jar --help`:
```
usage: java -jar ChunkyLauncher.jar
Chunky Launcher 1.14.1, Copyright (c) 2010-2024 Jesper �qvist and Chunky Contributors

Chunky comes with ABSOLUTELY NO WARRANTY. This is free software, and you are welcome to redistribute
it under certain conditions. See the GNU General Public License v3 for more details.

    --console                               Forces debug console to be opened
    --dangerouslyDisableLibraryValidation   Disable library validation. This can be dangerous!
 -h,--help                                  Show this help message
    --javaOptions <options>                 Add a Java option when launching Chunky
    --launcher                              Forces the launcher GUI to be shown
    --setup                                 Runs the interactive command-line launcher setup
    --update <release channel>              Update Chunky to the latest release
    --verbose                               Enables verbose logging
    --version                               Show the launcher version and exit

Command line options after -- are passed to Chunky.
For Chunky's command line help, run:
  java -jar ChunkyLauncher.jar -- --help
```

## Running `java -jar ChunkyLauncher.jar --setup`
```
Memory limit (MiB) [8192]: 
Java options [--module-path "C:\Users\user\.chunky\javafx-sdk-17.0.2\lib" --add-modules javafx.controls,javafx.fxml]: 
Update site [https://chunky-pr.lemaik.de/]: 
Reload launcher metadata [Y/n]: 
...Launcher is up to date!
33 release channels found.
```

PR #1859, opened 2025-12-31
Add a new launcher cli option to select the version to launch

leMaik:
…and move the release channel into a distinct option.

This builds upon #1682 (which was never released, so the options change doesn't break anything) and allows stuff like `java -jar ChunkyLauncher.jar --releaseChannel snapshot --chunkyVersion latest` to launch the latest snapshot, even if the `stable` channel is the currently selected channel.

Also you can now download and launch PR builds with a single command :sparkles: 

```
java -jar ChunkyLauncher.jar \
  --updateSite https://chunky-pr.lemaik.de \
  --releaseChannel pr-1859 \
  --update \
  --chunkyVersion latest
```

### DIFF
--- a/launcher/src/se/llbit/chunky/launcher/ChunkyLauncher.java
+++ b/launcher/src/se/llbit/chunky/launcher/ChunkyLauncher.java
@@ -118,12 +118,6 @@ public class ChunkyLauncher {
     );
 
     options.addOption(Option.builder()
-      .longOpt("setup")
-      .desc("Runs the interactive command-line launcher setup")
-      .build()
-    );
-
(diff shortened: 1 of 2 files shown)


leMaik:
The second example above doesn't work because `--update` exits. I see how just updating might be useful.

Maybe we should switch to some library that supports subcommands, like picocli or JCommander?

ThatRedox:
I think there should be a way to do that as is, but if you want to switch to a better CLI parsing library I'm not against it.

leMaik:
@ThatRedox I found a workaround :tada:

PR #1869, opened 2026-02-06
Add default materials for weighted pressure plates

JustinTimeCuber:
Gives light and heavy weighted pressure plates the same default properties as gold and iron blocks respectively.

Before (strange, inconsistently lit banana):
Banana-4000 denoised

After (much nicer, consistently shiny banana):
metallic-pressure-plates


leMaik:
Never noticed that oversight before. Thanks for improving this! 🍌

PR #1880, opened 2026-04-09
feat: warn about unsupported blocks after chunk loading

### LINKED ISSUE #1801: Warn about unsupported blocks
While loading the chunks, we could collect all unsupported blocks in a `Set` and then show a warning to update to the latest Chunky version, look at the _minecraft_ label on GitHub or create a new issue.

mvanhorn:
## Summary

After loading chunks, Chunky now logs a warning listing any block types that weren't recognized by any BlockProvider. The warning suggests updating Chunky or checking the minecraft label on GitHub.

## Changes

- `BlockPalette.java`: Added a `ConcurrentHashMap.KeySetView<String>` to collect unknown block names during `put()`. Added `getUnsupportedBlocks()` accessor.
- `Scene.java`: After `loadChunks()` returns, checks the palette for unsupported blocks and logs a warning if any are found.

## Testing

- Could not run Gradle build locally (no Java runtime), but the changes follow existing patterns in the codebase

Fixes #1801

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
+++ b/chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
@@ -64,6 +64,7 @@ public class BlockPalette {
 
   private final Map<BlockSpec, Integer> blockMap;
   private List<Block> palette;
+  private final Set<String> unsupportedBlocks = ConcurrentHashMap.newKeySet();
 
   private ReentrantLock lock = new ReentrantLock();
 
@@ -131,12 +132,23 @@ public class BlockPalette {
       id = palette.size();
       blockMap.put(spec, id);
       Block block = spec.toBlock();
+      if (block instanceof UnknownBlock) {
+        unsupportedBlocks.add(block.name);
+      }
       applyMaterial(block);
       palette.add(block);
       return id;
     } finally {
       lock.unlock();
     }
+  }
+
+  /**
+   * Returns the set of block names that were not recognized by any block provider.
+   * These blocks are rendered as {@link UnknownBlock}.
+   */
+  public Set<String> getUnsupportedBlocks() {
+    return Collections.unmodifiableSet(unsupportedBlocks);
   }
 
   public Block get(int id) {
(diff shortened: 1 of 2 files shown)


chunky/src/java/se/llbit/chunky/renderer/scene/Scene.java
```diff
@@ -575,6 +575,13 @@ public synchronized void loadScene(RenderContext context, String sceneName, Task
...
         } else {
           loadChunks(taskTracker, loadedWorld, chunksToLoadByRegion);
+          if (palette != null) {
+            Set<String> unsupported = palette.getUnsupportedBlocks();
+            if (!unsupported.isEmpty()) {
+              Log.warn("Unsupported blocks found: " + String.join(", ", unsupported)
+                  + ". Consider updating Chunky or check https://github.com/chunky-dev/chunky/labels/minecraft for known issues.");
+            }
+          }
         }
```
NotStirred: This is only hit when loading an existing scene that doesn't have an octree

Should probably do this inside `loadChunks` itself

chunky/src/java/se/llbit/chunky/chunk/BlockPalette.java
```diff
@@ -64,6 +64,7 @@ public class BlockPalette {
 
   private final Map<BlockSpec, Integer> blockMap;
   private List<Block> palette;
+  private final Set<String> unsupportedBlocks = ConcurrentHashMap.newKeySet();
```
NotStirred: `put` locks internally, and we're guaranteed that all `put`s have finished by the time chunkloading is finished (and `getUnsupportedBlocks` is called).

Can use `HashMap` to avoid extra work in chunk loading hot path (@leMaik is this worth the performance? In <=1.12 palette is around 50% of chunk loading, >1.12 it's about 5%)

Would probably need some comments explaining that exactly what's going on with the concurrent access (I'll do that after merge)

NotStirred (changes requested):

mvanhorn:
Addressed in cb75335:
- Moved the unsupported blocks check from the Scene caller into `loadChunks` itself, so it runs for all callers (not just the scene reload path)
- Changed `unsupportedBlocks` from `ConcurrentHashMap.newKeySet()` to `HashSet` since `put()` locks internally and all puts complete before `getUnsupportedBlocks()` is called

NotStirred (approved):

PR #1890, opened 2026-05-20
Fix CLI not always printing the render completed message.

leMaik:
I'd appreciate some feedback on this bug. :pray: 

The "Render job finished." message and render time didn't always show up in the CLI. This fixes it and the fix looks straight forward, but was the old comparison intentional? Am I breaking something?


leMaik:
I'm pretty sure this happened if the branch count resulted in more spp than requested being rendered. E.g. if branch count is 2, but target spp is 5, it will render 6 spp.

PR #1895, opened 2026-07-08
Create a single place for all chunky thread/executors to be registered

### LINKED ISSUE #1893: Most threads are marked as non-daemon
This basically means chunky will always require a `System.exit` to close, which can skip non-`finally` block resource closures.

This includes `Region Parser` and `3D Render Worker` threads, and a few others.

NotStirred:
I'm not tied to any of the specifics here, but having a mechanism to wait on chunky to join its threads is very important with Bedrock.
Essentially leveldb is thread safe on reads, so I don't want to lock. If region parsers/chunk loading are active when the DB is closed it races and results in almost always a `segfault`/`pthread lock` error. The `ChunkyThread#joinAll()` lets the shutdown hook wait for everything before closing the db.

Closes #1893

- Allows chunky to close threads on shutdown without a System.exit()
- Additionally provides an api to wait on all chunky threads to be joined.

### DIFF
--- a/chunky/src/java/se/llbit/util/concurrent/ChunkyThread.java
+++ b/chunky/src/java/se/llbit/util/concurrent/ChunkyThread.java
@@ -1,19 +1,16 @@
 package se.llbit.util.concurrent;
 
 import se.llbit.chunky.main.Chunky;
-import se.llbit.log.Log;
+import se.llbit.util.annotation.NotNull;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 import java.util.function.Function;
 
 /**
  * {@link Thread}/{@link ExecutorService} resource management.
- * <h4>The goal of this glass is to primarily:</h4>
+ * <h4>The goal of this class is to primarily:</h4>
  * <ul>
  *   <li>Ensure all chunky threads (even daemons) are interrupted and given a chance clean up before getting killed by
  *       the runtime on termination.</li>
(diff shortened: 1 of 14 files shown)


chunky/src/java/se/llbit/util/concurrent/ChunkyThread.java
```diff
@@ -0,0 +1,239 @@
...
+   * <p>This method is always safe to call.</p>
+   *
+   * <p><b><i>WARNING: calling this from any thread registered with {@link #addThread(Thread)} may <u>deadlock</u>.</i></b></p>
+   *
+   * @param timeout The maximum time to wait <b><u>AFTER</u></b> a shutdown is initiated
+   * @param unit the time unit of the timeout argument
+   * @return Whether all threads were joined before returning
+   */
+  @PluginApi
+  public static boolean joinAll(long timeout, @NotNull TimeUnit unit) {
```
leMaik: This shouldn't be `@PublicApi` imho, and maybe we could even not make it public?

NotStirred: For bedrock I need some mechanism that guarantees `RegionParser` and `Scene#loadChunks` has stopped before closing the DB, currently I'm using this api locally

Do you have any other ideas that could replace this?

leMaik: I don't have one yet, but coupling "joining all chunky threads" and "load chunks" directly together doesn't feel right. Maybe some event listeners or make RegionParser and loadChunks return `CompletableFuture` to make it explicit that they are asynchronous?

leMaik: Btw. I really appreciate the effort to not leak threads anymore. :clap:

chunky/src/java/se/llbit/util/concurrent/ChunkyThread.java
```diff
@@ -178,6 +159,43 @@ public static boolean interruptAndJoinAll(long timeout, @NotNull TimeUnit unit)
...
+    for (ExecutorService executorService : executorServices) {
+      long waitTime = endTimeNanos - System.nanoTime();
+      anyAlive |= !executorService.awaitTermination(waitTime, TimeUnit.NANOSECONDS);
+    }
+    for (Thread thread : threads) {
+      long waitTime = endTimeNanos - System.nanoTime();
+      if (waitTime > 0) {
+        thread.join(waitTime); // joining with 0 is infinite wait time, very intuitive.
+      }
+      // Thread.isAlive() establishes a happens-before with the thread. As such the following are non-issues:
```
NotStirred, line 190: This is stated in [JLS17 at: `17.4.4`](https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.4.4): 
> The final action in a thread T1 synchronizes-with any action in another thread T2 that detects that T1 has terminated.

PR #1901, opened 2026-07-16
Fix biome texture being empty if any chunk has no blocks that are affected by the biome

leMaik:
As discussed on Discord: https://discord.com/channels/541221265512464394/541221265512464397/1526905107126423663

Thanks to @NotStirred for finding the bug.


NotStirred (approved):
👍

PR #1905, opened 2026-07-26
Resolve cyclic dependency between static initializers of EmptyWorld and EmptyDimension.

### LINKED ISSUE #1894: `EmptyWorld.INSTANCE` has a `null` `currentDimension()` because its static initializer has a cyclic dependency with `EmptyDimension`
This bug is currently not that bad, but if a chunky or a plugin were to ever classload `EmptyDimension` before `EmptyWorld`, `currentDimension` could be null which crashes everything.

Not really sure how to fix this, I guess we could give EmptyDimension a null `world`, but that means possible NPEs if its world is ever accessed

leMaik:
Fixes #1894


NotStirred (approved):
👍 fab

